PR #14, opened 2019-02-19
Fix to program hanging indefinitely when job count > numofworkers

### LINKED ISSUE #12: Task queue blocks on too many wildcard operations
Let me start by saying that `s5cmd` is a big improvement over `aws` and `s3cmd`. And wildcard support is by far the best feature of `s5cmd`.

However, it seems that `s5cmd` (release v0.6.0) is getting stuck when many wildcard commands are submitted using a file or stdin. As soon as number of wildcard (batch) operations is equal to or larger than `numworkers` nothing is actually downloaded. It looks like all workers are busy handling wildcards, waiting for actual put/get operations to finish. But there are no workers available to perform put/get operations.

This would not be a problem if number of workers could be set arbitrarily high. And technically it can. But we found out that on say i3.2 AWS instance (8 vCPUs) download speed is significantly reduced with 256 workers, compared to 16 or 32 workers. One can get around this by submitting listing commands in one file, then making a long list of non-batch operations in another file.

tuvangezer:
Semi-fixes #12 by showing an error and terminating the program when there are more wildcard operations than the number of workers instead of hanging indefinitely. 
Actually allowing the workers to run multiple wildcard operations without blocking requires the core logic to be changed.

### DIFF
--- a/core/worker.go
+++ b/core/worker.go
@@ -300,6 +300,7 @@ func (p *WorkerPool) Run(filename string) {
 	s := NewCancelableScanner(p.ctx, r).Start()
 
 	var j *Job
+	jobCount := 0
 	isAnyBatch := false
 	for {
 		line, err := s.ReadOne(p.subJobQueue, p.jobQueue)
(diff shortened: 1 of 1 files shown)


ilkinulas:
Thanks a lot for your efforts. We prefer fixing the bug entirely.
If you have any other solution to fix #12 please do not hesitate to submit another pr or update this one.

PR #49, opened 2020-02-06
Refactor s3 url parser

sonmezonur:
* Move s3 parser logic into URL package

* Add tests for s3 parser

* Simplify s3list logic

### DIFF
--- a/core/s3.go
+++ b/core/s3.go
@@ -6,8 +6,6 @@ import (
 	"errors"
 	"net/http"
 	"os"
-	"path"
-	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -44,54 +42,15 @@ func s3head(svc *s3.S3, obj *url.S3Url) (*s3.HeadObjectOutput, error) {
 type s3listItem struct {
 	*s3.Object
 
-	parsedKey      string
-	isCommonPrefix bool
+	key         string
+	isDirectory bool
 }
 
 func s3list(ctx context.Context, svc *s3.S3, s3url *url.S3Url, emitChan chan<- interface{}) error {
 	inp := s3.ListObjectsV2Input{
-		Bucket: aws.String(s3url.Bucket),
-	}
-
-	wildkey := s3url.Key
-	var prefix, filter string
-	loc := strings.IndexAny(wildkey, url.S3WildCharacters)
-	wildOperation := loc > -1
-	if !wildOperation {
-		// no wildcard operation
-		inp.SetDelimiter("/")
-		prefix = s3url.Key
-
-	} else {
-		// wildcard operation
-		prefix = wildkey[:loc]
-		filter = wildkey[loc:]
-	}
-	inp.SetPrefix(prefix)
-
-	var (
-		r   *regexp.Regexp
-		err error
-	)
-
-	trimPrefix := path.Dir(prefix) + "/"
-	if trimPrefix == "./" {
-		trimPrefix = ""
-	}
-	if !wildOperation {
-		// prevent "ls s3://bucket/path/key" from matching s3://bucket/path/keyword
-		// it will still match s3://bucket/path/keydir/ because we don't match the regex on commonPrefixes
-		filter = prefix[len(trimPrefix):]
-	}
-
-	if filter != "" {
-		filterRegex := regexp.QuoteMeta(filter)
-		filterRegex = strings.Replace(filterRegex, "\\?", ".", -1)
-		filterRegex = strings.Replace(filterRegex, "\\*", ".*?", -1)
-		r, err = regexp.Compile("^" + filterRegex + "$")
-		if err != nil {
-			return err
-		}
+		Bucket:    aws.String(s3url.Bucket),
+		Prefix:    aws.String(s3url.Prefix),
+		Delimiter: aws.String(s3url.Delimiter),
 	}
 
 	var mu sync.Mutex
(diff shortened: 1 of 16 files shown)


url/s3url.go
```diff
@@ -24,38 +40,112 @@ func (s S3Url) Format() string {
...
 }
 
 // HasWild checks if a string contains any S3 wildcard chars
 func HasWild(s string) bool {
-	return strings.ContainsAny(s, S3WildCharacters)
+	return strings.ContainsAny(s, s3WildCharacters)
 }
 
 // ParseS3Url parses a string into an S3Url
 func ParseS3Url(object string) (*S3Url, error) {
```
igungor: now that we're here, we should call this func `New`.

url/s3url.go
```diff
@@ -24,38 +40,112 @@ func (s S3Url) Format() string {
...
+	} else {
+		s.Prefix = s.Key[:loc]
+		s.filter = s.Key[loc:]
+	}
+}
+
+// setFilterRegex converts wildcard strings to regex format
+// and precompiles it for later usage. It is default to
+// ".*" to match every key on S3
+func (s *S3Url) setFilterRegex() error {
```
igungor: we should merge this func with `setPrefixAndFilter`. No need for a separate function.

url/s3url.go
```diff
@@ -1,18 +1,34 @@
...
+	s3WildCharacters string = "?*"
+
+	// s3Prefix is the prefix used on s3 URLs
+	s3Prefix string = "s3://"
+
+	// s3Separator is the path separator for s3 URLs
+	s3Separator string = "/"
+
+	// defaultRegex is the regex to match every key
+	defaultRegex string = ".*"
```
igungor: matchAllRe

url/s3url.go
sonmezonur: done

url/s3url.go
sonmezonur: done

core/s3.go
```diff
@@ -128,37 +86,35 @@ func s3list(ctx context.Context, svc *s3.S3, s3url *url.S3Url, emitChan chan<- i
...
 				continue
 			}
+
 			if !emit(&s3listItem{
-				Object:         c,
-				parsedKey:      key,
-				isCommonPrefix: isCommonPrefix,
+				Object:      c,
+				key:         key,
+				isDirectory: key[len(key)-1] == '/',
```
igungor: `strings.HasSuffix(key, "/")` is more readable.

s3url/s3url.go
```diff
@@ -0,0 +1,195 @@
+// Package url defines URL types and has helper methods to parse a string into URLs.
```
igungor: Package `s3url`

s3url/s3url.go
```diff
@@ -0,0 +1,195 @@
...
+	"regexp"
+	"strings"
+)
+
+const (
+	// s3WildCharacters is valid wildcard characters for a S3Url
+	s3WildCharacters string = "?*"
+
+	// s3Prefix is the prefix used on s3 URLs
+	s3Prefix string = "s3://"
```
igungor: let's call this `s3Schema` since we already have a term called `prefix`.

s3url/s3url.go
```diff
@@ -0,0 +1,195 @@
...
+
+	// s3Separator is the path separator for s3 URLs
+	s3Separator string = "/"
+
+	// matchAllRe is the regex to match everything
+	matchAllRe string = ".*"
+)
+
+// S3Url is the metadata that is used on S3 operations (listing, querying)
```
igungor: period.

s3url/s3url.go
```diff
@@ -0,0 +1,195 @@
...
+
+// Format formats the S3Url to the format "<bucket>[/<key>]".
+func (s S3Url) Format() string {
+	if s.Key == "" {
+		return s.Bucket
+	}
+	return s.Bucket + s3Separator + s.Key
+}
+
+// setFilterRegex creates url metadata for both wildcard and non-wildcard operations.
```
igungor: setPrefixAndFilter

s3url/s3url.go
```diff
@@ -0,0 +1,195 @@
...
+	parts := strings.SplitN(rawUrl, s3Separator, 4)
+	if parts[2] == "" {
+		return nil, fmt.Errorf("s3 url should have a bucket")
+	}
+	if HasWild(parts[2]) {
+		return nil, fmt.Errorf("bucket name cannot contain wildcards")
+	}
+	key := ""
+	if len(parts) == 4 {
+		key = strings.TrimLeft(parts[3], s3Separator)
```
igungor: is this TrimLeft or TrimPrefix?

s3url/s3url_test.go
```diff
@@ -0,0 +1,350 @@
...
+		after  *S3Url
+	}{
+		{
+			name: "wild_operation",
+			before: &S3Url{
+				Key: "a/b_c/*/de/*/test",
+			},
+			after: &S3Url{
+				Key:         "a/b_c/*/de/*/test",
+				Prefix:      "a/b_c/",
```
ilkinulas: let's add an assertion for the value of `Delimeter`.

s3url/s3url.go
```diff
@@ -0,0 +1,195 @@
...
+		Key:         s.Key,
+		Prefix:      s.Prefix,
+		filter:      s.filter,
+		filterRegex: s.filterRegex,
+	}
+}
+
+// Match check if given key matches with regex and
+// returns parsed key.
+func (s S3Url) Match(key string) (string, bool) {
```
ilkinulas: Can we update the signature of `Match` function like the following : 

`func (s S3Url) Match(key string) string` 

It can return empty string if there is not any match.  This will simplify the usage and test cases.

s3url/s3url.go
sonmezonur: `TrimLeft` would be a better choice since it removes all leading "/" characters while TrimPrefix does not. 

Example: Trimming `s3://bucket////key/b/c` should result with `key/b/c`. But if we use TrimPrefix, only the leading `/` will be removed and output will be like "//key/b/c". 
See: https://play.golang.org/p/GRP3JeKbl-5

igungor: 👍

s3url/s3url.go
sonmezonur: done

PR #50, opened 2020-02-11
Implement storage package

sonmezonur:


e2e/rm_test.go
```diff
@@ -70,7 +70,7 @@ func TestRemoveMultipleS3Objects(t *testing.T) {
 	result.Assert(t, icmd.Success)
 
 	assertLines(t, result.Stderr(), map[int]compareFunc{
-		0: suffix(` +OK "rm s3://%v/*" (4)`, bucket),
```
igungor: Have you found the cause of this change?

core/worker.go
```diff
@@ -30,47 +27,57 @@ type WorkerPoolParams struct {
...
-	awsSession    *session.Session
-	cancelFunc    context.CancelFunc
-	st            *stats.Stats
-	idlingCounter int32
+	ctx            context.Context
+	params         *WorkerPoolParams
+	jobQueue       chan *Job
+	subJobQueue    chan *Job
+	wg             *sync.WaitGroup
+	storageFactory func() (storage.Storage, error)
```
igungor: i think `newClient` sounds better

igungor: or just `client()`

core/worker.go
```diff
@@ -132,54 +128,38 @@ func (p *WorkerPool) runWorker(st *stats.Stats, idlingCounter *int32, id int) {
...
-
 		case <-p.ctx.Done():
 			return
 		}
 	}
 }
 
+// runJob runs parent job and returns sub-job.
+func (p *WorkerPool) runJob(wp WorkerParams, job *Job) *Job {
+	if err := job.Run(&wp); err != nil {
```
igungor: let's handle the non-error case first to reduce the indentation.

core/worker.go
sonmezonur: Changed to `newClient`

core/worker.go
sonmezonur: Logic was moved into `job.go` and checks non-error case first.

storage/s3.go
```diff
@@ -0,0 +1,354 @@
...
+		return err
+	}
+
+	for _, d := range o.Deleted {
+		s.stats.put(aws.StringValue(d.Key), StatsResponse{Success: true})
+	}
+
+	for _, e := range o.Errors {
+		s.stats.put(aws.StringValue(e.Key), StatsResponse{
+			Success: true,
```
igungor: should be false.

sonmezonur: done

PR #55, opened 2020-02-14
Start workers after validating jobs

sonmezonur:
Currently, workers are running without validating jobs. 

Help command should be isolated from the worker pool. Also, if the jobs are not valid, we should not start workers. This PR makes the following improvements:
 
- Remove `HELP` command from AcceptableErrors.
- Don't send help commands to the job queue.
- Start workers if jobs are parsable and not help.
- Use deferred calls for closing channels.

### DIFF
--- a/core/worker.go
+++ b/core/worker.go
@@ -144,7 +144,6 @@ func (p *WorkerPool) runWorker(st *stats.Stats, idlingCounter *int32, id int) {
 
 func (p *WorkerPool) parseJob(line string) *Job {
 	job, err := ParseJob(line)
-
 	if err != nil {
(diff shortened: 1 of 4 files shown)


core/worker.go
```diff
@@ -185,22 +189,39 @@ func (p *WorkerPool) pumpJobQueues() {
...
 	}
 
-	close(p.jobQueue)
-	p.wg.Wait()
+	if j.opts.Has(opt.Help) {
+		j.displayHelp()
+		return
+	}
+
+	// start workers if the job is parsable and not help
```
igungor: we can remove this comment.

sonmezonur: done

PR #56, opened 2020-02-17
Return structured type after running each job

sonmezonur:
* Remove AcceptableError
* Return JobResponse after each command
* Remove numFails, numSuccess and numAcceptableFails from Job
* Implement generic Log() function
* Do not print stats on parent job


core/op_s3.go
```diff
@@ -202,43 +228,53 @@ func S3Download(job *Job, wp *WorkerParams) (stats.StatType, error) {
...
-func S3Upload(job *Job, wp *WorkerParams) (stats.StatType, error) {
+func S3Upload(job *Job, wp *WorkerParams) (stats.StatType, *JobResponse) {
 	const opType = stats.S3Op
 
 	if ex, err := job.args[0].Exists(wp); err != nil {
-		return opType, err
+		return opType, &JobResponse{status: statusErr, err: err}
 	} else if !ex {
-		return opType, os.ErrNotExist
+		return opType, &JobResponse{status: statusErr, err: os.ErrNotExist}
```
igungor: what about a function like

```
func jobResponse(err error) *JobResponse {
	if err == nil {
		return &JobResponse{status: statusSuccess}
	}
	return &JobResponse{status: statusErr, err: err}
}
```

to make it more readable?

e2e/bang_test.go
```diff
@@ -54,7 +54,6 @@ func TestSuccessfulBangBang(t *testing.T) {
 	result := icmd.RunCmd(cmd)
 
 	result.Assert(t, icmd.Expected{ExitCode: 0})
-
```
igungor: let's keep this line :) for me it's readable this way.

core/op_s3.go
sonmezonur: done

e2e/bang_test.go
sonmezonur: updated

core/jobargument.go
```diff
@@ -75,56 +66,61 @@ func (a *JobArgument) Append(s string, isS3path bool) *JobArgument {
 	return a
 }
 
-func (a *JobArgument) CheckConditionals(wp *WorkerParams, src *JobArgument, opts opt.OptionList) (ret error) {
+// CheckConditions checks if the job satisfies the conditions if the job has -n, -s and -u flags.
+// It returns error-embedded JobResponse with status "warning" if none of the requirements are met.
+// It returns nil if any warning or error is encountered during this check.
+func (a *JobArgument) CheckConditions(wp *WorkerParams, src *JobArgument, opts opt.OptionList) *JobResponse {
```
igungor: Since you're already here, could you separate this method from JobArgument and make it a function please? A function like 

`CheckConditions(src, dst *JobArgument, wp, opts) *JobResponse` makes it clear that `a` is the destination here.

core/jobargument.go
```diff
@@ -75,56 +66,61 @@ func (a *JobArgument) Append(s string, isS3path bool) *JobArgument {
...
 		tDest, err := a.ModTime(wp)
 		if err != nil {
-			return err
+			return &JobResponse{status: statusErr, err: err}
 		}
 
 		tSrc, err := src.ModTime(wp)
 		if err != nil {
-			return err
+			return &JobResponse{status: statusErr, err: err}
```
igungor: we can use `jobResponse` helper in this file as well.

core/jobargument.go
sonmezonur: Done

core/jobargument.go
sonmezonur: I used jobResponse() helper on all responses.

PR #57, opened 2020-02-17
objurl: unify s3 and local file url representation

igungor:


objurl/objurl.go
```diff
@@ -21,26 +22,116 @@ const (
...
+// remote storage.
+type ObjectURL struct {
+	Type      objurlType
+	Scheme    string
+	Bucket    string
+	Path      string
+	Delimiter string
+	Prefix    string
+
+	trimmedPath string
```
sonmezonur: I think `relativePath` would be better.

igungor: done.

objurl/objurl.go
```diff
@@ -201,9 +201,9 @@ func (o *ObjectURL) Clone() *ObjectURL {
 		Path:      o.Path,
 		Prefix:    o.Prefix,
 		// FIXME(ig): TBD
```
sonmezonur: can we remove this line now?

igungor: done.

PR #65, opened 2020-02-24
Implement standalone job producer

sonmezonur:
* Parse command-line arguments to command struct

* Improve scanner

* Separate wildcard logic from worker

* Remove cyclic job dependency

* Limit number of goroutines with semaphore

* Remove Marker from S3 storage

* Remove subJobQueue


complete/complete.go
```diff
@@ -249,7 +249,7 @@ func s3predictor(a cmp.Args) []string {
 		url.Prefix = s3key
 
 		for object := range client.List(ctx, url, true, s3MaxKeys) {
-			if object.IsMarker() {
+			if object.Err != nil {
```
igungor: let's add a `// TODO` marker to log the error once we have a better logging mechanism.

core/command.go
```diff
@@ -6,10 +6,94 @@ import (
...
 
+	"github.com/peak/s5cmd/objurl"
 	"github.com/peak/s5cmd/op"
 	"github.com/peak/s5cmd/opt"
+	"github.com/peak/s5cmd/storage"
 )
 
+// Command is the representation of the command.
+type Command struct {
+	sourceDesc string
```
igungor: `sourceDesc` sounds like it's related with the `src` description, but it's actually raw command string, right? Could we rename this to more expressive one? `raw`, `original` etc. would work. If you have a better one, it's ok for me.

core/job.go
```diff
@@ -20,14 +17,14 @@ const dateFormat = "2006/01/02 15:04:05"
...
-	args       []*JobArgument
-	opts       opt.OptionList
-	subJobData *subjobStatsType
-	isSubJob   bool
-	response   *JobResponse
+	opts      opt.OptionList
+	operation op.Operation
+	src       []*objurl.ObjectURL
+	dst       *objurl.ObjectURL
+	cls       string
```
igungor: is this storageClass? If so, we can use storage.StorageClass type.

core/job.go
```diff
@@ -98,152 +78,39 @@ func (j *Job) Log() {
...
-func (j *Job) run(wp *WorkerParams) *JobResponse {
+// Run runs the Job, gets job response and logs the job status.
+func (j *Job) Run(wp *WorkerParams) {
 	cmdFunc, ok := globalCmdRegistry[j.operation]
 	if !ok {
-		return &JobResponse{
-			status: statusErr,
-			err:    fmt.Errorf("unhandled operation %v", j.operation),
-		}
+		log.Fatalf("unhandled operation %v", j.operation)
```
igungor: hmm, i didn't expect a direct exit here. is there a way to propagate the error?

core/job.go
```diff
@@ -98,152 +78,39 @@ func (j *Job) Log() {
...
+		return
 	}
 
-	kind, response := cmdFunc(j, wp)
+	// runner will get cmdFunc
```
igungor: we can remove this comment.

core/job_check.go
```diff
@@ -0,0 +1,66 @@
...
+			res = &JobResponse{status: statusWarning, err: ErrObjectIsNewer}
+		} else {
+			res = nil
+		}
+	}
+
+	return res
+}
+
+// getObjects creates new storage client and sends stat request.
```
igungor: getObject returns the Object structure describing the given object URL.

core/job_generator.go
```diff
@@ -0,0 +1,108 @@
+package core
```
igungor: Filename `core/job_generator.go` confused me a bit because there is both generator and producer. How about job_batch.go or sth?

core/job_producer.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/peak/s5cmd/op"
+	"github.com/peak/s5cmd/opt"
+	"github.com/peak/s5cmd/storage"
+)
+
+type producerFunc func(*Command, ...*objurl.ObjectURL) *Job
+
+type producerOp struct {
+	fn       producerFunc
+	fullScan bool
```
igungor: comment would be nice for `fullScan`.

core/job_producer.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (p *Producer) fullScan(ctx context.Context, command *Command, fn producerFunc) {
+	// TODO(os): handle errors
+	client, _ := p.newClient(command.src)
+	isRecursive := command.opts.Has(opt.Recursive)
+
+	var urls []*objurl.ObjectURL
+	for object := range client.List(ctx, command.src, isRecursive, storage.ListAllItems) {
+		if object.Err != nil || object.Mode.IsDir() {
```
igungor: let's add a `// TODO` marker here to log the error later.

core/job_producer.go
```diff
@@ -0,0 +1,86 @@
...
+	p.runJob(job)
+}
+
+func (p *Producer) lookup(ctx context.Context, command *Command, fn producerFunc) {
+	// TODO(os): handle errors
+	client, _ := p.newClient(command.src)
+	isRecursive := command.opts.Has(opt.Recursive)
+
+	for object := range client.List(ctx, command.src, isRecursive, storage.ListAllItems) {
+		if object.Err != nil || object.Mode.IsDir() {
```
igungor: let's add a `// TODO` marker here to log the error later.

core/op_s3.go
```diff
@@ -3,364 +3,193 @@ package core
...
-
-	response := CheckConditions(src, dst, wp, job.opts)
-	if response != nil {
-		return opType, response
-	}
-
-	srcFn := src.url.Base()
-	destFn := dst.url.Absolute()
+	srcFn := src.Base()
+	destFn := dst.Absolute()
```
igungor: while we are here, should we change `Fn` convention to more readable variable names? dstFilename or similar.

core/parse.go
```diff
@@ -163,61 +159,53 @@ func parseArgumentByType(s string, t opt.ParamType, fnObj *JobArgument) (*JobArg
...
 		if parts[0] == c.Keyword { // The first token is the name of our command, "cp", "mv" etc.
 			found = i // Save the id of the last matching command, we will use this in our error message if needed
 
-			// Enrich our skeleton Job with default values for this specific
-			// command
-			ourJob.command = c.Keyword
-			ourJob.operation = c.Operation
-			ourJob.args = []*JobArgument{}
-			ourJob.opts = c.Opts
+			// Enrich our skeleton Job with default values for this specific command
```
igungor: we can remove this comment.

core/worker.go
```diff
@@ -73,149 +67,88 @@ func NewWorkerPool(ctx context.Context, params *WorkerPoolParams, st *stats.Stat
...
+		job.Run(wp)
+	}()
 }
 
-// RunCmd will run a single command (and subsequent sub-commands) in the worker pool, wait for it to finish, clean up and return.
-func (p *WorkerPool) RunCmd(commandLine string) {
-	defer p.closeAndWait()
-
-	j := p.parseJob(commandLine)
+// RunCmd will run a single command (and subsequent sub-commands) in the worker manager,
```
igungor: there are no `sub-commands` now, right?

core/worker.go
```diff
@@ -230,28 +163,26 @@ func (p *WorkerPool) Run(filename string) {
...
+		command := w.parseCommand(cmd)
+		if command != nil {
+			producer.Run(w.ctx, command)
 		}
 	}
-	if isAnyBatch {
-		p.pumpJobQueues()
+
+	if err := scanner.Err(); err != nil {
+		if err == context.Canceled {
```
igungor: let's use `errors.Is`

core/parse.go
```diff
@@ -163,61 +159,53 @@ func parseArgumentByType(s string, t opt.ParamType, fnObj *JobArgument) (*JobArg
...
-	if jobdesc == "" || jobdesc[0] == '#' {
+// parseSingleCommand attempts to parse a single command description to a standalone Command struct.
+// It will loop through each accepted command-signature, trying to find the first one that fits.
+func parseSingleCommand(cmd string) (*Command, error) {
+	if cmd == "" || cmd[0] == '#' {
 		return nil, nil
 	}
 
-	if strings.Contains(jobdesc, "&&") {
+	if strings.Contains(cmd, "&&") {
```
igungor: we don't need them anymore. leftovers.

core/parse.go
```diff
@@ -163,61 +159,53 @@ func parseArgumentByType(s string, t opt.ParamType, fnObj *JobArgument) (*JobArg
...
 		return nil, errors.New("nested commands are not supported")
 	}
-	if strings.Contains(jobdesc, "||") {
+	if strings.Contains(cmd, "||") {
```
igungor: we don't need them anymore. leftovers.

core/command.go
igungor: Done.

core/job.go
igungor: done.

core/job.go
igungor: done.

core/job_producer.go
igungor: gone.

PR #84, opened 2020-03-02
core: set ContentType for uploaded files

### LINKED ISSUE #33: Upload to S3 always adds metadata "content-type: binary/octet-stream" instead of default per filetype.
The question is, is there a way to control the metadata / contentype of the upload? It seem to default to upload as `content-type: binary/octet-stream`

The goal is to either be able to set the metadata content-type, or make s3 use the default content-type per filetype. When using the painfully slow aws cli, `aws s3 cp` or `aws s3 sync` it will automatically add a correct content type to metadata. This is critical for us that uses s3 to host websites. 

I would be happy to look into a PR, but could use a hint or two to where to modify. 

This could origin from the difference between using `aws s3api` and `aws s3`. I belive this tool is i using s3api?

igungor:
Fixes #33


igungor:
@sonmezonur please review

igungor:
I'm going to wait for https://github.com/peak/s5cmd/pull/87

PR #85, opened 2020-03-03
Support json formatted output

### LINKED ISSUE #22: Add option for json output
It would be really useful if there was an option print the output in json format, so that it can be parsed easily.

sonmezonur:
Resolves #22


sonmezonur:
@igungor Can you review this PR? What I have done so far:

- Support single-line JSON formatted outputs
- Log job status after it is done rather than the preparation step. This removes logs like `Uploading...`, `Removing...` etc.
- Combine info and success log levels (since it is only logged after the job is successfully finished).
- Remove leading prefix `+` from info logs.
- I put message structures into new package `message` (not sure about this, we may reallocate it to a better place).

Example output:
- json
```
{
  "operation": "download",
  "success": true,
  "source": "s3://tmp/10mb",
  "destination": "10mb",
  "object": {
    "type": "file",
    "size": 104857600
  }
}
```
- text (default)
```
                    download s3://bigtmp/10mb
```

Also, we can cut off leading spaces for info logs on text format. What dou you think?

igungor:
> * Support single-line JSON formatted outputs
> * Log job status after it is done rather than the preparation step. This removes logs like `Uploading...`, `Removing...` etc.

👍 

> * Combine info and success log levels (since it is only logged after the job is successfully finished).
> * Remove leading prefix `+` from info logs.

👍 

> * I put message structures into new package `message` (not sure about this, we may reallocate it to a better place).

I think it's better to keep those types where they are used. ListBuckets knows better what to print, so let it manage its message container.

> Example output:
> 
> * json
> 
> ```
> {
>   "operation": "download",
>   "success": true,
>   "source": "s3://tmp/10mb",
>   "destination": "10mb",
>   "object": {
>     "type": "file",
>     "size": 104857600
>   }
> }
> ```

Nice. Though I thought we made json logs single line?
 
> * text (default)
> 
> ```
>                     download s3://tmp/10mb
> ```
> 
> Also, we can cut off leading spaces for info logs on text format. What dou you think?

that'd be great actually.

sonmezonur:
> Nice. Though I thought we made json logs single line?

It is a single line. I just made it for readability.

> I think it's better to keep those types where they are used. ListBuckets knows better what to print, so let it manage its message container.

I will separate each command and its message to go file starting with `cmd_` prefix.

sonmezonur:
@igungor Done

storage/storage.go
```diff
@@ -47,31 +48,71 @@ func NewClient(url *objurl.ObjectURL) (Storage, error) {
...
-	URL          *objurl.ObjectURL
-	Etag         string
-	ModTime      time.Time
-	Mode         os.FileMode
-	Size         int64
-	StorageClass StorageClass
-	Err          error
+	URL          *objurl.ObjectURL `json:"key,omitempty"`
+	Etag         string            `json:"etag,omitempty"`
+	ModTime      *time.Time        `json:"last_modified,omitempty"`
```
igungor: Why is this change necessary?

sonmezonur: In JSON output, non-pointer time.Time prints default time even if it is not given. That is why I updated it to pointer and put `omitempty` into its struct tag.

core/cmd_size.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// SizeMessage is the structure for logging disk usage.
+type SizeMessage struct {
+	Source       string `json:"source"`
+	StorageClass string `json:"storage_class,omitempty"`
+	Count        int64  `json:"count"`
+	Size         int64  `json:"size"`
+
+	ShowHumanized bool `json:"-"`
```
igungor: we can make this field private since it is internal

core/message.go
```diff
@@ -0,0 +1,116 @@
...
+
+// String is the string representation of ErrorMessage.
+func (e ErrorMessage) String() string {
+	err := cleanupSpaces(e.Err)
+	return fmt.Sprintf(e.Format, e.Job, err)
+}
+
+// JSON is the JSON representation of ErrorMessage.
+func (e ErrorMessage) JSON() string {
+	e.Err = cleanupSpaces(e.Err)
```
igungor: i'm not sure if setting e.Err would be persistent across function calls. `e` is not a pointer receiver. Could you test?

core/message.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// newError creates new ErrorMessage.
+func newError(job *Job, err error, format string) ErrorMessage {
+	errStr := ""
+	if err != nil {
+		errStr = err.Error()
+	}
+
+	msg := ErrorMessage{
```
igungor: return ErrorMessage{...}

core/message.go
```diff
@@ -0,0 +1,116 @@
...
+
+// JSON is the JSON representation of ErrorMessage.
+func (e ErrorMessage) JSON() string {
+	e.Err = cleanupSpaces(e.Err)
+	b, _ := json.Marshal(e)
+	return string(b)
+}
+
+// newError creates new ErrorMessage.
+func newError(job *Job, err error, format string) ErrorMessage {
```
igungor: newErrorMessage suits well

core/message.go
```diff
@@ -0,0 +1,116 @@
...
+
+// warningMessage is the helper function to create warning messages.
+func warningMessage(job *Job, err error) {
+	format := "%q (%v)"
+	msg := newError(job, err, format)
+	log.Logger.Warning(msg)
+}
+
+// errorMessage is the helper function to create error messages.
+func errorMessage(job *Job, err error) {
```
igungor: printError?

core/message.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// JSON is the JSON representation of DebugMessage.
+func (d DebugMessage) JSON() string {
+	b, _ := json.Marshal(d)
+	return string(b)
+}
+
+// debug is the helper function to create warning messages.
+func debug(format string, args ...interface{}) {
```
igungor: printDebug?

core/worker.go
```diff
@@ -92,8 +97,14 @@ func (w *WorkerManager) parseCommand(cmd string) *Command {
...
 	close(w.semaphore)
-	// Workermanager is responsible for logging, hence we run the close routine
+}
+
+// closeStdout prints stats message if it is enabled
+// and closes stdout channel.
+func (w *WorkerManager) closeStdout(force bool) {
+	// WorkerManager is responsible for logging, hence we run the close routine
 	// here.
+	stats.Print(force)
```
igungor: why can't we call this function in WorkerManager's close() method?

flags/flags.go
```diff
@@ -30,8 +30,9 @@ var (
 	PrintStats          = flag.Bool("stats", false, "Always print stats")
 	ShowVersion         = flag.Bool("version", false, "Prints current version")
 	EnableGops          = flag.Bool("gops", false, "Initialize gops agent")
-	Verbose             = flag.Bool("vv", false, "Verbose output")
 	NoVerifySSL         = flag.Bool("no-verify-ssl", false, "Don't verify SSL certificates")
+	JSON                = flag.Bool("json", false, "JSON formatted output")
+	LogLevel            = flag.String("log-level", "info", "Log level (possible values: 'debug', 'info', 'warning', error'")
```
igungor: what do you think about making it `--log=value`?

objurl/objurl.go
```diff
@@ -243,7 +244,14 @@ func (o *ObjectURL) Match(key string) bool {
...
 func (o *ObjectURL) String() string {
-	return o.Absolute()
+	if o.IsRemote() {
+		return o.Absolute()
+	}
+	return o.Base()
+}
+
+func (o *ObjectURL) MarshalJSON() ([]byte, error) {
+	return json.Marshal(o.String())
```
igungor: `return []byte(o.String()), nil`

storage/s3.go
```diff
@@ -359,8 +371,11 @@ func (s *S3) doDelete(ctx context.Context, chunk chunk, resultch chan *Object) {
 	}
 }
 
-// MultiDelete is a removal operation which removes multiple S3 objects from a
-// bucket using single HTTP request. It allows deleting objects up to 1000.
+// MultiDelete is a asynchronous removal operation for multiple objects.
+// It reads given url channel, creates multiple chunks and run these
+// chunks by parallel. Each chunk may have at most 1000 objects since DeleteObjects
```
igungor: in parallel.

storage/storage.go
```diff
@@ -47,38 +46,82 @@ func NewClient(url *objurl.ObjectURL) (Storage, error) {
...
 }
 
 // String returns the string representation of Bucket.
 func (b Bucket) String() string {
 	return fmt.Sprintf("%s  s3://%s", b.CreationDate.Format(dateFormat), b.Name)
 }
 
+// String returns the JSON representation of Bucket.
+func (b Bucket) JSON() string {
+	bytes, _ := json.Marshal(b)
```
igungor: this pattern is used a lot of times. we can write function like `jsonMarshal(v interface{}) string`

e2e/util_test.go
```diff
@@ -4,6 +4,7 @@ package e2e
 
 import (
 	"bytes"
+	jsonencode "encoding/json"
```
igungor: jsonpkg

storage/storage.go
sonmezonur: Where should it be located? Both `storage`, `stats` and `core` uses this. I could not find any good place to do this.

flags/flags.go
sonmezonur: It is better. Updated it.

core/message.go
sonmezonur: It is not needed since it is only needed in this function scope. But I found a better way. We can use `cleanupSpaces` method on `log` package.

core/cmd_size.go
sonmezonur: Done. I also updated other internal fields. It was intended for `message` package, that's why all of the fields were public.

e2e/util_test.go
sonmezonur: done

objurl/objurl.go
sonmezonur: It does not work same. It breaks to e2e tests :)

core/worker.go
sonmezonur: Run() mode(file) prints stats message as a default while RunCmd() does not. That is why a flag like `force` is required to preserve standard behavior. I did not want to pass a boolean parameter to close() method since it was confusing and only required for printing stats. `close()` sounds like a WorkerManager functionality to me. What do you think? Should we change the signature of close() to close(bool)?

core/message.go
sonmezonur: Sorry, I moved it into `newErrorMessage()`. We can clean up multi-line strings on the constructor.

objurl/objurl.go
igungor: Hmm, because of the quotes I guess.

log/log.go
```diff
@@ -2,121 +2,145 @@ package log
...
-	if *flags.Verbose {
-		l.printf(levelDebug, format, args...)
+	if *flags.JSON {
+		outputCh <- output{
+			message: message.JSON(),
+			std:     std,
+		}
+	} else {
+		outputCh <- output{
+			message: fmt.Sprintf("%v %v", level, message.String()),
```
igungor: while we're here, could we remove the leading whitespace on Info() calls? I mean this:

` download s3://bucket/obj`

igungor: ignore this, you already mentioned this on another comment.

storage/storage.go
igungor: I'm not sure. All packages may hold their own jsonMarshal() function.

log/log.go
sonmezonur: done. also, tests are updated to use equals() rt suffix() for the strict check.

storage/storage.go
sonmezonur: Done. Let's use `jsonMarshal` helper function on `core` and `storage` packages. I did not update `stats` since it uses this pattern only once.

README.md
```diff
@@ -76,39 +76,36 @@ Options:
...
-  -vv
-    	Verbose output
-  -y	Don't prompt user for typing 'yes'
 
 Commands:
-    !, cp, du, get, ls, mv, rm
+    cp, du, get, ls, mb, mv, rm
 
-To get help on a specific command, run "s5cmd <command> -h"
+To get help on a specific command, run "./s5cmd <command> -h"
```
igungor: this won't work it's locally build the tool.

README.md
```diff
@@ -236,17 +230,45 @@ To use a different profile, set the `AWS_PROFILE` env var. For more options, see
...
+    [operation] [source]
+
+    # example: download s3://bucket/key
+    ```
+    * failure
+
+    ```
+    [ERROR|WARNING] [message]
+
+    # example: ERROR "! touche": executable file not found in $PATH
```
igungor: `!` command is gone :)

core/cmd_size.go
```diff
@@ -0,0 +1,110 @@
...
+		return jobResponse(err)
+	}
+
+	for k, v := range storageTotal {
+		m := SizeMessage{
+			Source:       src.String(),
+			StorageClass: k,
+			Count:        v.count,
+			Size:         v.size,
+			shoHumanized: job.opts.Has(opt.HumanReadable),
```
igungor: typo. `showHumanized`

PR #87, opened 2020-03-03
core/job*: log object.Err values returned from storage.List

igungor:


e2e/du_test.go
```diff
@@ -155,7 +155,8 @@ func TestDiskUsageMissingObject(t *testing.T) {
 	result.Assert(t, icmd.Success)
 
 	assertLines(t, result.Stdout(), map[int]compareFunc{
-		0: suffix(` + 0 bytes in 0 objects: s3://%v/non-existent-file [Total]`, bucket),
-		1: equals(""),
+		0: equals("-ERR list: no object found"),
+		1: suffix(` + 0 bytes in 0 objects: s3://%v/non-existent-file [Total]`, bucket),
```
igungor: looks bad. we need to handle this case better. both `ERR` and `0 bytes in 0 objects` are not user-friendly outputs.

sonmezonur: agreed. can you put TODO label for this test so that we can fix it in following PRs.

core/job_exec.go
```diff
@@ -173,8 +173,8 @@ func BatchDelete(ctx context.Context, job *Job) *JobResponse {
 			// src is a glob
 			src := sources[0]
 			for obj := range client.List(ctx, src, true, storage.ListAllItems) {
-				if obj.Err != nil {
-					// TODO(ig): add proper logging
+				if err := obj.Err; err != nil {
+					log.Logger.Error("list: %v", err)
```
sonmezonur: Should we check context cancellation here? I am not sure if we should log them.

core/job_exec.go
```diff
@@ -232,8 +232,8 @@ func List(ctx context.Context, job *Job) *JobResponse {
 	}
 
 	for object := range client.List(ctx, src, true, storage.ListAllItems) {
-		if object.Err != nil {
-			// TODO(ig): expose or log the error
+		if err := object.Err; err != nil {
+			log.Logger.Error("list: %v", err)
```
sonmezonur: Same here.

core/job_exec.go
```diff
@@ -290,8 +290,11 @@ func Size(ctx context.Context, job *Job) *JobResponse {
 	totals := map[string]sizeAndCount{}
 
 	for object := range client.List(ctx, src, true, storage.ListAllItems) {
-		if object.Mode.IsDir() || object.Err != nil {
-			// TODO(ig): expose or log the error
+		if object.Mode.IsDir() {
+			continue
+		}
+		if err := object.Err; err != nil {
+			log.Logger.Error("list: %v", err)
```
sonmezonur: +++

core/job_exec.go
sonmezonur: Or, we can easily handle this on `storage.S3`  and not send cancellation errors to channel.

igungor: since `storage.List` is async, we piggyback the errors onto the Objects transferred with the channel as you know. Currently if a user gets a 403 (no access to the bucket), we ignore it.

sonmezonur: Sorry, I could not explain it well. Let me clarify deeply:

Currently, each operation takes the context as a first parameter and we are bypassing each error to object channel on async List operations without checking its type ([ref](https://github.com/peak/s5cmd/blob/master/storage/s3.go#L209)). Since we are going to log them, I think we should ignore RequestCancelled errors. It can be easily done by using `IsCancelationError()` method on this [line] (https://github.com/peak/s5cmd/blob/master/storage/s3.go#L209) and not sending these errors to channel.

Following is the example usage:
```
^C2020/03/03 23:18:33 # Got signal, cleaning up...
-ERR list: RequestCanceled: request context canceled caused by: context canceled
```

igungor: Hmm, right. I think all errors should propagate to the one who received it. Caller decides what to do with it. If the caller decides to log the error, we could suppress cancelation errors in the logger level. That'd require changing the logger.Error signature to receive `error` instead of `format, args...`.

sonmezonur: That is also another option and it makes sense for me to handle these errors on the logger level. We already need to break API for supporting structured logs (JSON etc.) so we can do it now :)

igungor: Just sent a commit that wouldn't change the signature but i'm not 100% sure about it. Also we should change logger method names to take `f` suffix, as in `formatted`. logger.Error sounds like it only receives a string.

sonmezonur: I am also not sure. How about:
```
       if err != nil {
                  if !isCancellationError(err) {		
                          log.Logger.Error("storage: %v", err)
                  }
		return
	}
```

igungor: Is this control on storage layer? It'd be better if we don't log here. If it's on the caller site, we need to add this check to multiple functions, right?

sonmezonur: It is on caller site. There will be duplicate code but we don't need to construct error on log package. Just a suggestion, we can keep current structure.

igungor: Done.

core/job_exec.go
igungor: Done.

core/job_exec.go
igungor: Done.

PR #92, opened 2020-03-13
Add --force-path-style flag to use path-style addressing.

joshuarobinson:
Many/most non-AWS object stores only support path-style addressing, but
recent changes switched to virtual-host addressing which broke access to
those object stores. This adds a flag to switch back to path style
addressing.

For more info, a virtual-host address looks like "http://bucketname.endpoint",
resulting in a DNS lookup. Path-style addressing looks like
"http://endpoint/bucketname" instead. The benefit of path-style
addressing is simplicity; you do not need to interact with DNS servers.

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -21,6 +21,7 @@ const (
 var (
 	CommandFile         = flag.String("f", "", "Commands-file or - for stdin")
 	EndpointURL         = flag.String("endpoint-url", "", "Override default URL with the given one")
+	ForcePathStyle      = flag.Bool("force-path-style", false, "Use path-style addressing")
 	WorkerCount         = flag.Int("numworkers", defaultWorkerCount, "Number of worker goroutines. Negative numbers mean multiples of the CPU core count")
 	DownloadConcurrency = flag.Int("dw", defaultDownloadConcurrency, "Download concurrency for each file")
 	DownloadPartSize    = flag.Int64("ds", 50, "Multipart chunk size in MB for downloads")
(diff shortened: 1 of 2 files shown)


joshuarobinson:
First time submitting changes, so please let me know if I've missed anything (tests, etc?)

Recent changes broke access to our on-prem object store (see [blog here](https://medium.com/@joshua_robinson/s5cmd-for-high-performance-object-storage-7071352cc09d))

igungor:
Hi @joshuarobinson, thanks for the pull request! Also about the blog post, thank you very much for detailed review and your time 💯 . Much appreciated.

I'd like to know how do you access to your on-prem object store? Is it like a minio installation? I'm asking this because may be we could drop the ``--force-path-style` flag altogether by inferring the path-style from the endpoint URL. If it's not the official S3/Google etc. end point, we could use path-style instead of virtual-host-style.

joshuarobinson:
Hi, thanks for the response.

Yes, our on-prem object is accessed identically to Minio --  and almost any
other on-prem object appliance.

I do like your proposed solution better than mine, but I will need to do
some research today to determine how the endpoint detection could work
(other than simply detecting an IP address vs hostname).

Btw, my favorite part of s5cmd is the "-f -" feature; it creates a really
flexible engine for advanced users.

-Joshua

igungor:
Hey @joshuarobinson 

I drafted a branch to showcase the approach I mentioned above. What do you think?

joshuarobinson:
Thanks,
I have checked your branch locally and confirmed it works.

And I like that the behavior explainable and intuitive -- if the user
manually sets "endpoint-url" (to something other than AWS transfer-accel or
GCP), the addressing scheme falls back to path-style.

I would approve of this change :)

igungor:
Glad it works for you. I'll go ahead and merge that branch. Thanks again!

joshuarobinson:
many thanks!

Issue #23, opened 2019-09-25
Correct exit status for ls command


ilyas28:
`s5cmd ls s3://path/to/non-existent/object` exits with code 0. This means this command cannot be used to check if an object exist is a bucket or not.

PR #94, opened 2020-03-16
Use urfave/cli package

igungor:
### TODO

- [x] Update README.md
  - [x] Advise installing from prebuilt binaries instead of `go get`. `master` branch is in a constant change right now and we can't guarantee backwards compatibility till v1.0.0 release.
  - [x] Change `Usage` section according to the latest `cli` changes.

### Breaking changes
- `s5cmd get` command is removed. We have 2 identical commands without a clear reasoning. `s5cmd cp <src> .` has the same result as `s5cmd get <src>`.
- `batch-rm` command is removed. Users can run `s5cmd rm` with variadic arguments. Wildcards are accepted (WIP)
- `s5cmd -f` became `s5cmd run`
- All errors are written to stderr
- `s5cmd ls s3://non-existent/object` returns `1` instead of `0`. Related: #23 
- Exit code for errors changed from `127` to `1`
- `s5cmd cp s3://bucket/object local/directory/` now creates the missing `local/directory/` instead of returning an error.


command/cmd_copy.go
```diff
@@ -0,0 +1,632 @@
...
+	dst *objurl.ObjectURL,
+	parents bool,
+) (*objurl.ObjectURL, error) {
+	objname := src.Base()
+	if parents {
+		objname = src.Relative()
+	}
+
+	if originalSrc.HasGlob() {
+		os.MkdirAll(dst.Absolute(), os.ModePerm)
```
ilkinulas: `os.MkdirAll` returns an error if create dir fails. We should handle it.

aykutfarsak:
Could we add the argument names also? It helps reading code, like 

> What is the meaning of the third param of List() method?

https://github.com/peak/s5cmd/blob/97f7a4940d7c405c6ab1b84aa78a915e59e8dff4/storage/storage.go#L25-L36

seruman:
> Could we add the argument names also? It helps reading code, like
> 
> > What is the meaning of the third param of List() method?
> 
> https://github.com/peak/s5cmd/blob/97f7a4940d7c405c6ab1b84aa78a915e59e8dff4/storage/storage.go#L25-L36

Since implementations omit unused ones, this might be helpful

command/cmd_copy.go
igungor: done.

igungor:
@aykutfarsak @seruman done. please add your comments to that particular line.

command/cmd_get.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/peak/s5cmd/storage"
+)
+
+var GetCommand = &cli.Command{
+	Name:     "get",
+	HelpName: "get",
+	Usage:    "TODO",
```
seruman: Missin usage

command/cmd_copy.go
```diff
@@ -0,0 +1,638 @@
...
+
+// expandSource returns the full list of objects from the given src argument.
+// If src is an expandable URL, such as directory, prefix or a glob, all
+// objects are returned by walking the source.
+func expandSource(
+	ctx context.Context,
+	src *objurl.ObjectURL,
+	isRecursive bool,
+) (<-chan *storage.Object, error) {
+	// TODO(ig): this function could be in the storage layer.
```
seruman: +1

igungor: not going to happen in this PR. looks easy but i think it's not.

command/cmd_get.go
igungor: `get` command is gone. Please see the description.

PR #100, opened 2020-03-17
goreleaser: Support multiple release artifacts

### LINKED ISSUE #21: Include binaries for supported platforms and architectures
Include x86_64 go binary as a Release artifact as a convenience to users so that they do not need to make the binary themselves.

sonmezonur:
Resolves #21

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -2,8 +2,7 @@ env:
   - GO111MODULE=on
 before:
   hooks:
-    - go mod tidy
-    - go get -v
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -1,16 +1,53 @@
+env:
+  - GO111MODULE=on
+before:
+  hooks:
+    - go mod tidy
```
igungor: we could just run `make` to cover all checks.

PR #104, opened 2020-03-17
Use struct on copy command

sonmezonur:


command/cmd_copy.go
```diff
@@ -153,103 +147,20 @@ func Copy(
...
-					return &errorpkg.Error{
-						Op:  op,
-						Src: src,
-						Dst: dsturl,
-						Err: err,
-					}
-				}
-				return nil
-			}
+			task = c.copy(ctx, srcurl, src, dsturl)
```
igungor: Copy -> c.copy -> c.doCopy feels like too many redirections to me. If we are going in this direction, we should rename `c.copy` to `c.prepareCopyTask` or similar. Same for `upload` and `download`.

sonmezonur: done

command/cmd_copy.go
```diff
@@ -156,11 +156,11 @@ func (c Copy) Run(ctx context.Context) error {
 
 		switch {
 		case srcurl.Type == dsturl.Type: // local->local or remote->remote
-			task = c.copy(ctx, srcurl, src, dsturl)
+			task = c.prepareCopyTask(ctx, srcurl, src, dsturl)
```
igungor: `srcurl` and `src` confused me. I know that srcurl is the original source URL but i'll forget it in a week :)

Let's make it `prepareCopyTask(ctx, origSrc, srcurl, dsturl)`

sonmezonur: Agreed. Also, we should be more strict on URL naming. All objurl types should have `url` suffix. (e.g srcurl instead of src). Because we already have raw src string on command structs, it might be confusing. I renamed all objurl variables.

PR #106, opened 2020-03-18
Support variable arguments for 'rm' command

### LINKED ISSUE #2: Support for multiple local sources (expanded wildcards)
If a wildcard are expanded to more than one local file, the command fails with "Invalid parameters" error message. As a workaround, local wildcards should be passed in single quotes to prevent wildcard expansion.

Commands with local-file sources should support more than one source to fix this. This would also enable multiple wildcards being specified as source.

Also keep the built-in wildcard expansion so that piped wildcards in a commands file continue to work.

sonmezonur:
This PR adds support for variadic delete operations. It works for both wildcard and non-wildcard strings and does not break backward-compatibility. 

For example;

```
s5cmd rm test.txt testdir/a.txt 'folder/*'
```


command/variadic.go
```diff
@@ -0,0 +1,138 @@
...
+type Arg struct {
+	origSrc *objurl.ObjectURL
+	obj     *storage.Object
+}
+
+// expandSources returns the full list of objects from the given src arguments.
+// If src is an expandable URL, such as directory, prefix or a glob, all
+// objects are returned by walking the source. It expands multiple resources asynchronously
+// and returns read-only arg channel.
+func expandSources(
```
ilkinulas: This function well deserves a unit test.

sonmezonur: I will add both e2e and unit test

sonmezonur: I added multiple e2e tests for the use case.

objurl/objurl.go
```diff
@@ -122,7 +122,10 @@ func (o *ObjectURL) Absolute() string {
 
 // Relative returns a URI reference based on the calculated prefix.
 func (o *ObjectURL) Relative() string {
-	return o.relativePath
+	if o.relativePath != "" {
```
sonmezonur: This change is required since Relative() for non-wildcard urls return empty. For example, `s3://bucket/key` has empty relative path and while using it with --parents flag, it constructs wrong destination URL. 
  
```
s5cmd cp --parents testfile "testdir/*" s3://bucket

# we need both the content inside /testdir and testfile.
```
Above example was not working before since testfile creates a target `s3://bucket` (which is wrong actually).

command/variadic.go
```diff
@@ -0,0 +1,138 @@
...
+	ctx context.Context,
+	isRecursive bool,
+	dsturl *objurl.ObjectURL,
+	srcurls ...*objurl.ObjectURL,
+) (<-chan *Arg, error) {
+	if len(srcurls) == 0 {
+		return nil, fmt.Errorf("at least one source url is required")
+	}
+	// all sources share same client
+	client, err := storage.NewClient(srcurls[0])
```
ilkinulas: Injecting storage class to this function may help writing unit tests easily.

sonmezonur: Since we expect different results for each given source URL, only mocking the result of storage.List() will not work (since all sources share same client). We either need storage slice instead of source URL slice (then we can mock each one) or find a way to return results based on given URL. 
```

**Comment by igungor** [comment]
PTAL

File: objurl/objurl.go (hunk 2 above)
Comment (igungor) [reply objurl/objurl.go]: 👍

File: README.md [hunk 4]
```diff
@@ -16,6 +16,7 @@ storage services and local filesystems.
 - Create buckets
 - Summarize objects sizes, grouping by storage class
 - Wildcard support for all operations
+- Varargs support for copy, move and delete operations
```
igungor: "Delete operation supports multiple arguments"

sonmezonur: done

command/variadic.go
sonmezonur: @ilkinulas Could you please review expandSources function again. Since it only supports delete operations, I think it is much clear.

igungor:
Conflict olmus.

command/variadic.go
sonmezonur: I added unit tests.

command/cmd_delete.go
```diff
@@ -97,3 +100,89 @@ func Delete(
...
+	objChan := make(chan *storage.Object)
+	go func() {
+		defer close(objChan)
+
+		var wg sync.WaitGroup
+		var objFound bool
+
+		for _, origSrc := range srcurls {
+			// call storage.List for only walking operations.
+			if origSrc.HasGlob() {
```
igungor: what about directories?

sonmezonur: It uses expandSource() method now. It makes stat call and walks directories if the source URL is local. I also added e2e tests for the directory usage.

PR #110, opened 2020-03-19
command: move copy flags from global to copy command

igungor:


command/cmd_copy.go
```diff
@@ -69,6 +81,18 @@ var CopyCommand = &cli.Command{
 			return fmt.Errorf("target %q can not contain glob characters", dst)
 		}
 
+		if c.Int64("upload-chunk-size") < 5 {
+			return fmt.Errorf("upload chunk size should be greater than 5 MiB")
+		}
+
+		if c.Int64("download-chunk-size") < 5 {
```
sonmezonur: before check needs to be updated.

igungor: done.

igungor:
PTAL

PR #123, opened 2020-03-24
storage: add GCS list support

### LINKED ISSUE #81: Support Google Cloud Storage
GCS [supports S3 protocol](https://cloud.google.com/storage/docs/interoperability). It'd be nice if we could support it.

One blocker is that GCS doesn't support ListObjectsV2 API. We need to fallback to ListObjects API if we're working on GCS.

igungor:
Fixes #81

### DIFF
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -28,6 +28,8 @@ import (
 
 var _ Storage = (*S3)(nil)
 
+var sentinelURL = urlpkg.URL{}
+
 const (
 	// deleteObjectsMax is the max allowed objects to be deleted on single HTTP
 	// request.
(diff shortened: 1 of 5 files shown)


storage/s3.go
```diff
@@ -212,6 +243,84 @@ func (s *S3) List(ctx context.Context, url *url.URL, _ bool) <-chan *Object {
 	return objCh
 }
 
+// listObjects is used for cloud services that does not support S3
+// ListObjectsV2 API. I'm looking at you GCS.
+func (s *S3) listObjects(ctx context.Context, url *url.URL) <-chan *Object {
```
sonmezonur: listObjects() and listObjectsV2() are almost identical. I think we should reuse the code by writing a simple helper function. I am not sure but we can do something like:
```
list(ctx context.Context, url *url.URL, isRecursive bool, fn func())
// it will take closure (either ListObjectsWithContext or ListObjectsV2PagesWithContext) and called according to storage type.

igungor: agreed, though we still need a control for v1/v2 conditions. i don't have an elegant solution.

sonmezonur: I also could not find. Let's merge this one until finding an elegant solution.

PR #124, opened 2020-03-24
Keep ObjectURL origin

### LINKED ISSUE #116: Keep ObjectURL origin
While expanding a given ObjectURL in storage.List method, keep a reference to the the original expandable form.

For example;

originalSrc: `s3://bucket/prefix/obj*`
expandedSrc: `s3://bucket/prefix/object1.gz`

`expandedSrc.Origin()` should return `originalSrc`.

sonmezonur:
Resolves #116 

We can keep the reference URL when we call url.Clone() method.


command/cmd_copy.go
```diff
@@ -532,7 +529,7 @@ func prepareCopyDestination(
 
 	// Absolute <src> path is given. Use given <dst> and local copy operation
 	// will create missing directories if <dst> has one.
-	if !origSrc.HasGlob() {
+	if !srcurl.HasGlob() {
```
igungor: i think we should explicitly call srcurl.Origin().HasGlob(). `srcurl` itself is an absolute url, not a glob.

storage/url/url.go
```diff
@@ -277,6 +284,9 @@ func (u *URL) MarshalJSON() ([]byte, error) {
 
 // HasGlob checks if a string contains any wildcard chars.
 func (u *URL) HasGlob() bool {
+	if u.origin != nil {
```
igungor: we can revert this change. callers must call url.Origin().HasGlob() if the intention is to check source origin has glob.

sonmezonur: The intention here is to recursively call origin URL. Because fs behaves differently from the s3 client. We both clone the original URL after expanding glob([ref](https://github.com/peak/s5cmd/blob/65fa203c5b053060563f99b9cf2588da25633030/storage/fs.go#L89)) and walking directory ([ref](https://github.com/peak/s5cmd/blob/65fa203c5b053060563f99b9cf2588da25633030/storage/fs.go#L113)). That is why source URL has an origin and also origin can keep its origin :D.   When I check just the origin without this logic, some of the tests are failing.

sonmezonur: done. How about a explicit SetOrigin() method similar to SetRelative() ? Can you take a look?

PR #125, opened 2020-03-25
command/copy: drop local->local copy support

### LINKED ISSUE #118: Drop local->local copy support
`s5cmd` supports local to local copy/move operations. I'm not sure we have a strong use-case to support it.

Also we have the `-R` recursive parameter which is used only for local->local copy. Seeing this parameter in the copy help output confuses me whether I need to use it to upload a directory to S3 or not.

I think we should drop `-R` and local->local copy support entirely if there's no opposition.

igungor:
We had '-R' recursive flag for local->local copy operations. Now that
there is no support, flag has been dropped.

Fixes #118


storage/fs.go
```diff
@@ -161,16 +156,11 @@ func (f *Filesystem) readDir(ctx context.Context, src *url.URL, ch chan *Object)
...
 }
 
-func (f *Filesystem) walkDir(ctx context.Context, src *url.URL, isRecursive bool) <-chan *Object {
+func (f *Filesystem) walkDir(ctx context.Context, src *url.URL) <-chan *Object {
 	ch := make(chan *Object)
 	go func() {
 		defer close(ch)
 
-		if !isRecursive {
-			f.readDir(ctx, src, ch)
```
sonmezonur: Remove fs.readDir() method. It is not used anymore.

igungor: done.

PR #128, opened 2020-03-25
Print destination url on info logs

### LINKED ISSUE #120: Print destination for copy/move operations
Example:

`copy: dir/file s3://bucket/prefix/object`

Also maybe we should print `time` and `log level` as well, such as;

`2020-03-24 10:10:10 INFO "copy: dir/file s3://bucket/prefix/object"`

sonmezonur:
Resolves #120


sonmezonur:
@igungor ping

Log format:
```
{operation} {source} {destionation}
// e.g cp s3://bucket/key absolutekey
```

e2e/rm_test.go
```diff
@@ -472,8 +472,8 @@ func TestVariadicMultipleLocalFilesWithDirectory(t *testing.T) {
 		0: equals(""),
 		1: equals("rm file1.txt"),
 		2: equals("rm file2.txt"),
-		3: equals("rm readme.md"),
-	}, sortInput(true))
+		3: equals("rm testdir/readme.md"),
+	}, strictLineCheck(true), sortInput(true))
```
igungor: we can remove strictLineCheck(true). it's enabled by default.

e2e/rm_test.go
```diff
@@ -430,10 +430,10 @@ func TestRemoveLocalDirectory(t *testing.T) {
...
 	assertLines(t, result.Stdout(), map[int]compareFunc{
 		0: equals(""),
-		1: equals("rm file1.txt"),
-		2: equals("rm file2.txt"),
-		3: equals("rm readme.md"),
-	}, sortInput(true))
+		1: equals("rm testdir/file1.txt"),
+		2: equals("rm testdir/file2.txt"),
+		3: equals("rm testdir/readme.md"),
+	}, strictLineCheck(true), sortInput(true))
```
igungor: we can remove strictLineCheck(true). it's enabled by default.

sonmezonur: done

e2e/rm_test.go
sonmezonur: done

PR #129, opened 2020-03-26
Align ls -e output

### LINKED ISSUE #101: `s5cmd ls -e` output is broken
Listing prefix/object mix with the `etag` parameter results in an unaligned output.

```
Δ s5cmd ls -e 's3://bucket/tmp/a/'
                                      DIR b/
2020/03/16 13:02:07   d41d8cd98f00b204e9800998ecf8427e            0 file2.txt
```

sonmezonur:
Fixes #101

### DIFF
--- a/command/cmd_list.go
+++ b/command/cmd_list.go
@@ -163,11 +163,11 @@ const (
 
 // String returns the string representation of ListMessage.
 func (l ListMessage) String() string {
-	var listFormat = "%19s %1s %-1s %8s %s"
+	var listFormat = "%19s %1s %-1s %12s %s"
 	var etag string
 	if l.showEtag {
 		etag = l.Object.Etag
-		listFormat = "%19s %1s %-38s %8s %s"
+		listFormat = "%19s %1s %-38s %12s %s"
 	}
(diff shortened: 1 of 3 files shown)


command/cmd_list.go
```diff
@@ -158,12 +158,18 @@ func (l ListMessage) humanize() string {
...
 const (
-	listFormat = "%19s %1s %-6s %12s %s"
 	dateFormat = "2006/01/02 15:04:05"
 )
 
 // String returns the string representation of ListMessage.
 func (l ListMessage) String() string {
+	var listFormat = "%19s %1s %-1s %12s %s"
+	var etag string
+	if l.showEtag {
```
igungor: could you add an e2e test case?

sonmezonur: sure

sonmezonur: I added alignment() function to check the last indexes of strings. But since gofakess3 does not support tags (returns ?). It won't check if ETag flag brokes the alignment of output.  We can revert this change if you wish, or keep it for safety.

PR #131, opened 2020-03-26
Handle too many open files errors

### LINKED ISSUE #52: fcntl: too many open files
I'm on MacOS 10.15.3 and I'm trying to upload to S3 a folder that contains 2616 folders with 1 to 10 files each.

With `s5cmd -stats -r 0 -vv cp -n --parents <src> <dest>` I immediately see this error:
```
VERBOSE: wildOperation lister is done with error: fcntl: too many open files
```
but the uploads seem to proceed, though the uploaded files at the end are generally less than 100.
Stats output:
```
2020/02/12 15:05:24 # All workers idle, finishing up...
2020/02/12 15:05:24 # Stats: S3             119   52 ops/sec
2020/02/12 15:05:24 # Stats: Failed         130   57 ops/sec
2020/02/12 15:05:24 # Stats: Total          249  109 ops/sec 2.282740841s
```

If I run the same command with the `-numworkers  16` option, the copy ends without errors and all files are correctly uploaded to S3

```
~$ ulimit -H -n
unlimited

~$ ulimit -S -n
256

~$ launchctl limit maxfiles
maxfiles    10240          10240
```

aykutfarsak:
It tries to increase the soft limit of open files to avoid hitting the limit by the OS.

It also catches `too many open files` errors and types a warning that includes a suggestion about `-numworkers` argument. Then exits instantly.

Fix #52


command/cmd_copy.go
```diff
@@ -147,6 +147,11 @@ func (c Copy) Run(ctx context.Context) error {
 	go func() {
 		defer close(errDoneCh)
 		for err := range waiter.Err() {
+			if strings.Contains(err.Error(), "too many open files") {
+				fmt.Println("WARNING You are hitting the max open file limit allowed by your OS. You can increase open file limit or try to decrease the number of workers with the parameter '-numworkers'.")
```
sonmezonur: Should we log these messages with log package?

```
msg := log.ErrorMessage{
	Err:       fmt.Errorf("error content here"),
}
log.Error(msg)
```
Also, custom message type and log level (WARNING in this case) can be added into log package if it is required.

aykutfarsak: @sonmezonur on my tests that use log package, the tool exits without printing the logs. `logger.out()` is responsible for printing logs and it is running in a goroutine. 

Is there any way to drain the logs before exit?

sonmezonur: [log.Close()](https://github.com/peak/s5cmd/blob/master/command/app.go#L97) is responsible for closing the logger and draining messages.

aykutfarsak: 👍 done

log/log.go
```diff
@@ -32,6 +32,11 @@ func Info(msg Message) {
 	global.printf(LevelInfo, msg, os.Stdout)
 }
 
+// Warning prints message in warning mode.
+func Warning(msg Message) {
+	global.printf(LevelWarning, msg, os.Stdout)
```
sonmezonur: I think we should output these logs into Stderr.

aykutfarsak: done

parallel/parallel.go
```diff
@@ -90,3 +93,24 @@ func (w *Waiter) Wait() {
...
+	err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rLimit)
+	if err != nil {
+		return
+	}
+
+	if rLimit.Cur >= minOpenFilesLimit {
+		return
+	}
+
+	if rLimit.Max < minOpenFilesLimit {
```
igungor: do we need this check?

aykutfarsak: If the hard limit is lower then 1000, it returns `invalid argument` on Setrlimit() call.

igungor: thanks

PR #132, opened 2020-03-26
storage/s3: add custom retryer for custom error codes

igungor:
We occasionally receive InternalError codes from S3 API. It was on our retry list before.

Adding a custom retryer that implements request.Retryer interface.

### DIFF
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -619,14 +619,13 @@ func newSession(opts S3Options) (*session.Session, error) {
 }
 
 type requestRetryer struct {
-	*client.DefaultRetryer
+	client.DefaultRetryer
+	maxRetries int
 }
 
 func newRequestRetryer(maxRetries int) *requestRetryer {
 	return &requestRetryer{
-		DefaultRetryer: &client.DefaultRetryer{
-			NumMaxRetries: maxRetries,
-		},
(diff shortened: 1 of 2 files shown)


storage/s3.go
```diff
@@ -616,6 +618,30 @@ func newSession(opts S3Options) (*session.Session, error) {
...
+}
+
+// ShouldRetry overrides the SDK's built in DefaultRetryer adding customization
+// to retry S3 InternalError code.
+func (c *CustomRetryer) ShouldRetry(req *request.Request) bool {
+	if errHasCode(req.Error, "InternalError") {
+		return true
+	}
+	return c.DefaultRetryer.ShouldRetry(req)
+
```
sonmezonur: we can remove this line

igungor: done.

storage/s3.go
```diff
@@ -616,6 +618,29 @@ func newSession(opts S3Options) (*session.Session, error) {
 	return sess, nil
 }
 
+// CustomRetryer wraps the SDK's built in DefaultRetryer adding additional
+// error codes. Such as, retry for S3 InternalError code.
+type CustomRetryer struct {
```
sonmezonur: should we make this unexported?

PR #133, opened 2020-03-26
Change default copy behaviour to preserve directory structure

### LINKED ISSUE #107: Change default copy behaviour to --parents
Not 100% sure about this, adding this here to track the issue.

Currently s5cmd flattens source directory hiearchy while downloading (cp/mv). For example:
```
s3://bucket/obj1
s3://bucket/obj2
s3://bucket/01/obj2
```

`s5cmd cp 's3://bucket/*' .`

The default behaviour is to find each object recursively, and download them while ignoring all prefixes after the wildcard. Result:

```
$ ls 

obj1
obj2
```

There'll be only 1 `obj2` but we're not very confident which one.

We should make --parents the default behaviour, and `--flatten` an option.

sonmezonur:
Resolves #107


e2e/cp_test.go
```diff
@@ -1176,6 +1455,46 @@ func TestCopySingleS3ObjectIntoAnotherBucket(t *testing.T) {
...
+		filename = "testfile1.txt"
+		content  = "this is a file content"
+	)
+
+	putFile(t, s3client, srcbucket, filename, content)
+
+	src := fmt.Sprintf("s3://%v/%v", srcbucket, filename)
+	dst := fmt.Sprintf("s3://%v/", dstbucket)
+
+	cmd := s5cmd("cp", "-flatten", src, dst)
```
igungor: --flatten.

sonmezonur:
> let's use `%v` in all tests instead of `%s`.

done

e2e/cp_test.go
sonmezonur: done

PR #139, opened 2020-03-30
command: add cat command

### LINKED ISSUE #20: Add option to cp to stdout
aws s3 cp has a magic "-" option which, on linux and macos at least, works like writing to /dev/stdout. s5cmd does not look like it supports providing an in-order stream of bytes (writing to stdout or a fifo fails because they don't allow seeking). 

Example use case: We would like to be able to pipe extremely large files (terabytes) from s3 into another program such as sha512sum without writing it to disk or storing the entire file in RAM.

seruman:
Fixes #20


command/cat.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/peak/s5cmd/storage"
+	"github.com/peak/s5cmd/storage/url"
+)
+
+var catHelpTemplate = `Name:
+	{{.HelpName}} - {{.Usage}}
+
+Usage:
+	{{.HelpName}} [options] source destination
```
igungor: there is no destination

command/cat.go
```diff
@@ -0,0 +1,90 @@
...
+		return Cat(c.Context, src, c.Int64("part-size")*megabytes)
+	},
+}
+
+func Cat(ctx context.Context, src *url.URL, partSize int64) error {
+	client, err := storage.NewClient(src)
+	if err != nil {
+		return err
+	}
+	_, err = client.Get(ctx, src, sequentialWriterAt{w: os.Stdout}, 1, partSize)
```
igungor: instead of sending actualsize/partSize number of requests sequentially, we can just send a `GetObject` request to download the object in 1 pass. S3 implemention of `Get` should fallback to `GetObject` if given concurrency is 1.

After that, there won't be any need for `part-size` flag.

e2e/cat_test.go
```diff
@@ -0,0 +1,212 @@
...
+			cmd: []string{
+				"cat",
+				src,
+			},
+			expected: expected,
+		},
+		{
+			name: "cat remote object with json flag",
+			cmd: []string{
+				"-json",
```
igungor: --json

command/cat.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	return nil
+}
+
+type sequentialWriterAt struct {
+	w io.Writer
+}
+
+func (fw sequentialWriterAt) WriteAt(p []byte, offset int64) (n int, err error) {
```
igungor: what does `f` stand for? 😃

e2e/cat_test.go
```diff
@@ -0,0 +1,212 @@
...
+			result := icmd.RunCmd(cmd)
+
+			result.Assert(t, icmd.Expected{ExitCode: 1})
+
+			assertLines(t, result.Stderr(), tc.expected)
+		})
+	}
+}
+
+// this method will create a string with 64666688 in size (~61.670 MB)
```
igungor: getSequentialFileContent function will

command/cat.go
igungor: also please don't do named returns.

command/cat.go
```diff
@@ -0,0 +1,90 @@
...
+	{{.HelpName}} - {{.Usage}}
+
+Usage:
+	{{.HelpName}} [options] source destination
+
+Options:
+	{{range .VisibleFlags}}{{.}}
+	{{end}}
+Examples:
+	1. Print an S3 object's content to stdout
```
igungor: we don't have a check if given argument has an s3 scheme.

command/cat.go
```diff
@@ -0,0 +1,90 @@
...
+	},
+}
+
+func Cat(ctx context.Context, src *url.URL, partSize int64) error {
+	client, err := storage.NewClient(src)
+	if err != nil {
+		return err
+	}
+	_, err = client.Get(ctx, src, sequentialWriterAt{w: os.Stdout}, 1, partSize)
+	if err != nil {
```
sonmezonur: Could you return raw err directly? Since we didn't wrap original errors before.

sonmezonur:
BTW README needs to be updated after all changes are made :) (etc. features and usage png)

e2e/cat_test.go
seruman: done

command/cat.go
seruman: missed that 🙃, done

e2e/cat_test.go
seruman: done

command/cat.go
seruman: replaced it with 'remote'

command/cat.go
seruman: done 👍

command/cat.go
seruman: updated

command/cat.go
seruman: would not `GetObject` load file contents into memory all at once?

e2e/cat_test.go
```diff
@@ -0,0 +1,212 @@
...
+		},
+		{
+			name: "cat remote object with glob",
+			cmd: []string{
+				"--json",
+				"cat",
+				src + "/*",
+			},
+			expected: map[int]compareFunc{
+				0: contains(`{"operation":"cat","command":"cat s3://bucket/file.txt/*","error":"remote source \"s3://bucket/file.txt/*\" can not contain glob characters"}`),
```
sonmezonur: We can use json() for testing json-formatted output. It supports multi-line strings.  Also `assertLines()` method can get  jsonCheck(bool) optional function for strict check. It validates if all lines at stout or stderr is json-decodable.

command/cat.go
igungor: Why?

command/cat.go
igungor: I mean, if we only allow `cat` to print remote object content, we need a check at `Before` stage.

igungor:
> BTW README needs to be updated after all changes are made :) (etc. features and usage png)

Also please update CHANGELOG.md file according to the changes.

command/cat.go
sonmezonur: I think `return err` would be better. We don't need if check.

e2e/cat_test.go
seruman: Thanks, I missed that, but `json()` also replaces spaces in quotes.
Added `jsonCheck(true)` checks.

command/cat.go
seruman: updated

command/cat.go
seruman: Updated, also added bucket check

command/cat.go
seruman: @igungor  could you check https://github.com/peak/s5cmd/pull/139/commits/5431ab52f69d11f0c3c4e1233146c814ed82dfd8 ?
Did you mean something like that?

sonmezonur: Do we really need this change or am I missing something? 

```
Specifying a Downloader.Concurrency of 1 will cause the Downloader to download the parts from S3 sequentially.
```

s3manager already checks the given concurrency and makes sequential download if it is 1. I think also giving some dummy part-size (let's say -1) will work, we don't need part-size flag if so.

igungor: I checked the code and if I'm not mistaken, SDK sends actualSize/partSize number of requests sequentially. So there will be many number of Range requests apart from consuming the object content.

sonmezonur: Ok. Thanks for the clarification. It is more clear now

seruman: I guess it's this part
https://github.com/aws/aws-sdk-go/blob/34abf397f4b59e6cae12bb41ea75dcc1d0dc25a9/service/s3/s3manager/download.go#L311-L333

command/cat.go
```diff
@@ -71,17 +64,17 @@ var CatCommand = &cli.Command{
...
 
-func Cat(ctx context.Context, src *url.URL, partSize int64) error {
+func Cat(ctx context.Context, src *url.URL) error {
 	client, err := storage.NewClient(src)
 	if err != nil {
 		return err
 	}
 
-	_, err = client.Get(ctx, src, sequentialWriterAt{w: os.Stdout}, 1, partSize)
+	_, err = client.Get(ctx, src, sequentialWriterAt{w: os.Stdout}, 1, -1)
```
igungor: a comment would be nice. what is 1, what is -1, and why.

storage/s3.go
```diff
@@ -341,6 +341,7 @@ func (s *S3) Copy(ctx context.Context, from, to *url.URL, metadata map[string]st
 
 // Get is a multipart download operation which downloads S3 objects into any
 // destination that implements io.WriterAt interface.
+// 'partSize' is ignored if given concurrency is 1
```
igungor: period.

storage/s3.go
```diff
@@ -726,3 +700,19 @@ func errHasCode(err error, code string) bool {
 func IsCancelationError(err error) bool {
 	return errHasCode(err, request.CanceledErrorCode)
 }
+
+// writerAtAdapter is an 'io.Writer' adapter for 'io.WriterAt'
```
igungor: period.

storage/s3.go
seruman: done

storage/s3.go
seruman: done

command/cat.go
seruman: added

seruman:
@ibrahim #140 it failed again
https://travis-ci.org/github/peak/s5cmd/builds/669156750?utm_source=github_status&utm_medium=notification

command/cat.go
```diff
@@ -0,0 +1,90 @@
...
+
+Options:
+	{{range .VisibleFlags}}{{.}}
+	{{end}}
+Examples:
+	1. Print a remote object's content to stdout
+		 > s5cmd {{.HelpName}} s3://bucket/prefix/object
+`
+
+var catCommandFlags []cli.Flag
```
igungor: leftover?

storage/s3.go
```diff
@@ -341,22 +341,34 @@ func (s *S3) Copy(ctx context.Context, from, to *url.URL, metadata map[string]st
 
 // Get is a multipart download operation which downloads S3 objects into any
 // destination that implements io.WriterAt interface.
+// Makes a single 'GetObject' call if 'concurrency' is 1 and ignores 'partSize'
```
igungor: period

seruman: 🤦‍♂

igungor:
> @ibrahim #140 it failed again
> https://travis-ci.org/github/peak/s5cmd/builds/669156750?utm_source=github_status&utm_medium=notification

yeah i know. cronic issue and we don't have a solution yet.

storage/s3.go
seruman: should be good now 🤞

command/cat.go
seruman: thanks, forgot it

seruman:
> waiting for README and CHANGELOG changes.

updated both files and `doc/usage.png`

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ This is a major release with many breaking changes.
 - Added `--json` flag for JSON logging. ([#22](https://github.com/peak/s5cmd/issues/22))
 - Added [S3 transfer acceleration](https://docs.aws.amazon.com/AmazonS3/latest/dev/transfer-acceleration.html) support. ([#40](https://github.com/peak/s5cmd/issues/40))
 - Added [Google Cloud Storage](https://github.com/peak/s5cmd#google-cloud-storage-support) support. ([#81](https://github.com/peak/s5cmd/issues/81))
+- Added `cat` command to print remote object contents to stdout ([20](https://github.com/peak/s5cmd/issues/20))
```
igungor: let's keep the style consistent. `#20`

sonmezonur:
> > waiting for README and CHANGELOG changes.
> 
> updated both files and `doc/usage.png`

It would be nice if you also introduce it in the features section.

PR #145, opened 2020-04-01
refactored s3 storage caching

ilkinulas:


command/app.go
```diff
@@ -77,9 +77,9 @@ var app = &cli.App{
 			NoVerifySSL: noVerifySSL,
 		}
 
-		storage.SetS3Options(s3opts)
+		err := storage.SetS3Options(s3opts)
```
sonmezonur, line 80: If we cannot create a session at the beginning, whole program will fail. In previous version, each job was trying to reuse the first successful session and create a new session if there is an error on previous attempts. I think this approach is so much cleaner since we don't need to handle errors in everywhere. But I am not sure we can safely exit the program here. What do you think?

ilkinulas: I think; using a  `fail fast`  approach is not a bad choice, since creating an s3 session is at the core of this tool.

storage/s3.go
```diff
@@ -74,10 +74,12 @@ func newS3Factory(opts S3Options) func() (*S3, error) {
 
 // newCachedS3 function returns a cached S3 storage with a re-used session if
```
sonmezonur, line 75: we should update this comment.

storage/s3.go
```diff
@@ -74,10 +74,12 @@ func newS3Factory(opts S3Options) func() (*S3, error) {
...
 // available. Re-used AWS sessions dramatically improve performance.
-var newCachedS3 func() (*S3, error)
+var cachedS3 *S3
 
-func SetS3Options(opts S3Options) {
-	newCachedS3 = newS3Factory(opts)
+func SetS3Options(opts S3Options) error {
```
sonmezonur, line 79: I think we should rename this function. `Init()` would be better.

storage/s3.go
```diff
@@ -74,10 +74,12 @@ func newS3Factory(opts S3Options) func() (*S3, error) {
...
+	s3, err := NewS3Storage(opts)
+	cachedS3 = s3
+	return err
```
sonmezonur, line 83: we can remove the empty line

command/app.go
```diff
@@ -77,9 +77,9 @@ var app = &cli.App{
 			NoVerifySSL: noVerifySSL,
 		}
 
-		storage.SetS3Options(s3opts)
+		err := storage.Init(s3opts)
```
sonmezonur, line 80: ``` return storage.Init(s3opts) ```

ilkinulas: I prefer leaving it as `return err`

`return storage.Init(s3opts)` makes me think that the function is returning an initialized storage :)

sonmezonur: hmm. you are right :)

ilkinulas: But `return storage.Init(s3opts)` guarantees that the last step in function is initializing the storage and avoids scenarios like:
```go
err := storage.Init(s3opts)

// this function should not be called if err!=nil
another_function()

return err
```

PR #162, opened 2020-05-02
parallel: move global manager into its own file

igungor:


parallel/parallel.go
```diff
@@ -3,31 +3,16 @@ package parallel
...
-	"syscall"
 )
 
-const minNumWorkers = 2
-const minOpenFilesLimit = 1000
-
-var global *Manager
+const (
+	minNumWorkers     = 2
+	minOpenFilesLimit = 1000
```
sonmezonur, line 10: we can also move `minOpenFilesLimit` into global.go since it is used by `adjustOpenFilesLimit()` func

igungor: done.

PR #163, opened 2020-05-02
storage/s3: allow any storage-class value

### LINKED ISSUE #160: Add latest AWS S3 storage class types
Can you please add the latest storage class types for S3:
'STANDARD','REDUCED_REDUNDANCY','STANDARD_IA','ONEZONE_IA','INTELLIGENT_TIERING','GLACIER','DEEP_ARCHIVE'
https://github.com/peak/s5cmd/blob/e1bf6cf824684f2d7f7c3f6296c265ac825a1c02/command/cp.go#L96

--storage-class (string) The type of storage to use for the object. Valid choices are: STANDARD | REDUCED_REDUNDANCY | STANDARD_IA | ONEZONE_IA | INTELLIGENT_TIERING | GLACIER | DEEP_ARCHIVE. Defaults to 'STANDARD'
https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html

Maybe it is also an idea to add the latest google cloud storage classes (standard, nearline, coldline):
https://cloud.google.com/storage/docs/storage-classes

igungor:
- e2e/cp: add --storage-class=GLACIER test
- also show 2 character storage class short code in 'ls' output

If you pass an invalid storage class, S3 will return:

```
s5cmd cp --storage-class=invalidstorageclass file s3://bucket/
ERROR "cp file s3://bucket/file": InvalidStorageClass: The storage class you specified is not valid status code: 400...
```

Fixes #160

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -131,7 +131,7 @@ var copyCommand = &cli.Command{
 			ifSourceNewer:  c.Bool("if-source-newer"),
 			flatten:        c.Bool("flatten"),
 			followSymlinks: !c.Bool("no-follow-symlinks"),
-			storageClass:   storage.LookupClass(c.String("storage-class")),
+			storageClass:   storage.StorageClass(c.String("storage-class")),
 			concurrency:    c.Int("concurrency"),
 			partSize:       c.Int64("part-size") * megabytes,
 		}.Run(c.Context)
@@ -214,7 +214,7 @@ func (c Copy) Run(ctx context.Context) error {
 			continue
 		}
 
-		if object.StorageClass == storage.StorageGlacier {
+		if object.StorageClass.IsGlacier() {
 			err := fmt.Errorf("object '%v' is on Glacier storage", object)
 			printError(c.fullCommand, c.op, err)
 			continue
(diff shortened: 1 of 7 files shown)


sonmezonur:
What do you think about removing shortcodes for storage class types? I know it would be a breaking change but I prefer to see INTELLIGENT_TIERING instead of IT in ls output :)

igungor:
> What do you think about removing shortcodes for storage class types? I know it would be a breaking change but I prefer to see INTELLIGENT_TIERING instead of IT in ls output :)

Hmm, may be we should drop it altogether and make it an option and show the full names. Such as `s5cmd ls -s s3://bucket/*`

sonmezonur:
> > What do you think about removing shortcodes for storage class types? I know it would be a breaking change but I prefer to see INTELLIGENT_TIERING instead of IT in ls output :)
> 
> Hmm, may be we should drop it altogether and make it an option and show the full names. Such as `s5cmd ls -s s3://bucket/*`

Sounds good to me.

igungor:
> Sounds good to me.

I'm going to merge this. Adding the show-storage-class-flag deserves its own PR.

igungor:
Could you review again please?

PR #169, opened 2020-05-28
cp: Set storage class if provided

sonmezonur:
If we don't provide `--storage-class` flag, cp gives InvalidStorageClass error on the master branch. I think it should be useful to give a default value for this flag. Also, updated usage to include new storage class types.

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -22,6 +22,7 @@ import (
 const (
 	defaultCopyConcurrency = 5
 	defaultPartSize        = 50 // MiB
+	defaultStorageClass    = "STANDARD"
 	megabytes              = 1024 * 1024
 )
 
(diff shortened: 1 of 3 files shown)


command/cp.go
```diff
@@ -92,8 +93,10 @@ var copyCommandFlags = []cli.Flag{
 		Usage: "do not follow symbolic links",
 	},
 	&cli.StringFlag{
-		Name:  "storage-class",
-		Usage: "set storage class for target ('STANDARD','REDUCED_REDUNDANCY','GLACIER','STANDARD_IA')",
+		Name:        "storage-class",
+		Value:       defaultStorageClass,
```
igungor, line 97: I think we need to set the storage class if provided, otherwise we're changing the class of an object.

https://github.com/peak/s5cmd/blob/19656e16d6bdfcda1d4a2627ad6ea5f282396778/storage/s3.go#L369-L372

sonmezonur: Okay, I analyzed the problem incorrectly. We don't need to set a default value. Currently, we are setting empty value for storage class (which is an invalid value) if no flag given in [S3 copy](https://github.com/peak/s5cmd/blob/master/storage/s3.go#L309). This causes errors like

```
 InvalidStorageClass: The storage class you specified is not valid status code: 400,
```
I updated PR accordingly. Could you pls review the PR again?

PR #174, opened 2020-06-08
Replace obsolete global option value in app.go

### LINKED ISSUE #173: What's the expected output of --log verbose (seems like it has no effect)?
Hello,

What's the expected output of  `--log verbose`? The reason I'm asking is because I run the same command with and without `--log verbose`, but there's no difference in the output. E.g. both commands return the same output:

`s5cmd --log verbose ...`
`s5cmd ...`

Am I passing the value `verbose` in a wrong format?

emres:
Replace obsolete global option value in app.go

The value `verbose` for global option `--log` has been deprecated in favor of `debug`. This fixes #173 .

### DIFF
--- a/command/app.go
+++ b/command/app.go
@@ -49,7 +49,7 @@ var app = &cli.App{
 		&cli.StringFlag{
 			Name:  "log",
 			Value: "info",
-			Usage: "log level: (verbose, info, error)",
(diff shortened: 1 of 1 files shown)


igungor:
Thanks!

PR #180, opened 2020-06-25
storage/s3: add more retry tests

igungor:


storage/s3_test.go
```diff
@@ -266,44 +267,134 @@ func TestS3ListContextCancelled(t *testing.T) {
...
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			mockApi.Handlers.Send.Clear() // mock sending
+			mockApi.Handlers.Unmarshal.Clear()
+			mockApi.Handlers.UnmarshalMeta.Clear()
+			mockApi.Handlers.ValidateResponse.Clear()
+			mockApi.Handlers.Unmarshal.PushBack(func(r *request.Request) {
+				r.Error = tc.err
+				r.HTTPResponse = &http.Response{StatusCode: 500}
```
sonmezonur, line 382: Not all errors listed above return 500 according to doc (https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html).

igungor: It's set to make aws defaultretryer happy. Otherwise, there are many status codes that we need to track. I'm going to remove that code now.

Issue #178, opened 2020-06-22
Windows filepath backslashes are not converted when uploaded to S3 recursively


mrekim0:
I created a windows binary as of fce1e2a, and noticed that the filepath retains the backslash when recursively copying to S3, which means the S3 folder keys separators are not parsed as folders.

This occurs in Windows 10 and in Cygwin, even with unix paths

**Example:**

**Folder contents**

```
d:\workspace
   |_dev
      |_foldera
         |_file.txt
```

`s5cmd cp d:\workspace\dev s3://mybucketname/`

**Expected result:**

`cp d:\workspace\dev\foldera\file.txt s3://mybucketname/dev/foldera/file.txt`

**Actual Result:**

`cp d:\workspace\dev\foldera\file.txt s3://mybucketname/dev\foldera\file.txt`

igungor:
https://github.com/peak/s5cmd/pull/181 should fix this issue.

PR #181, opened 2020-07-02
command/cp: fixed windows filepath issue on upload

fbarotov:
solves issue #178

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,14 @@ jobs:
     - run: go build -mod=vendor .
 
   test:
-    runs-on: [ubuntu-latest, windows]
+    strategy:
+      matrix:
+        os:
+          - macos
+          - ubuntu
+          - windows
+
+    runs-on: ${{ matrix.os }}-latest
 
     steps:
     - uses: actions/checkout@v2
@@ -34,7 +41,14 @@ jobs:
     - run: make test
 
   qa:
-    runs-on: [ubuntu-latest, windows]
+    strategy:
+      matrix:
+        os:
+          - macos
+          - ubuntu
+          - windows
+
+    runs-on: ${{ matrix.os }}-latest
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-on: [pull_request]
+on: [push, pull_request]
 name: CI
 jobs:
   build:
@@ -23,7 +23,7 @@ jobs:
     - run: go build -mod=vendor .
 
   test:
-    runs-on: ubuntu-latest
+    runs-on: [ubuntu-latest, windows]
 
     steps:
     - uses: actions/checkout@v2
@@ -34,7 +34,7 @@ jobs:
     - run: make test
 
   qa:
-    runs-on: ubuntu-latest
+    runs-on: [ubuntu-latest, windows]
 
     steps:
     - uses: actions/checkout@v2
--- a/command/cp.go
+++ b/command/cp.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"runtime"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
@@ -521,14 +520,6 @@ func prepareRemoteDestination(
 	objname := srcurl.Base()
 	if isBatch && !flatten {
 		objname = srcurl.Relative()
-	}
-	// if local->remote operation is performed on windows, \ should be
-	// replaced with /
-	// otherwise destination will be of the form,
-	// e.g., s3://myBucket\my\local\file\path
-	os := runtime.GOOS
-	if os == "windows" && strings.Contains(objname, "\\"){
-		objname = strings.ReplaceAll(objname, "\\", "/")
 	}
 
 	if dsturl.IsPrefix() || dsturl.IsBucket() {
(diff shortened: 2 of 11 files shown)


e2e/util_test.go
```diff
@@ -72,8 +73,17 @@ func setup(t *testing.T, options ...option) (*s3.S3, func(...string) icmd.Cmd, f
 	for _, option := range options {
 		option(opts)
 	}
+	// testdir := fs.NewDir() tries to create a new directory which
+	// has a prefix = [test function name][operation name]
+	// e.g., prefix' = "TestCopySingleS3ObjectToLocal/cp_s3://bucket/object_file"
+	// but on windows, directories cannot contain a colon
+	// so we replace them with hyphen
+	prefix := t.Name()
+	if runtime.GOOS == "windows"{
```
sonmezonur, line 82: Seems like this code is not formatted. Please run `go fmt ./...` after each update.

command/cp.go
```diff
@@ -395,6 +396,8 @@ func (c Copy) doUpload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) er
 	size := obj.Size
 
 	if c.deleteSource {
+		// close the file before deleting
```
sonmezonur, line 399: We can remove this comment.

strutil/strutil.go
```diff
@@ -40,3 +42,11 @@ func JSON(v interface{}) string {
 	bytes, _ := json.Marshal(v)
 	return string(bytes)
 }
+
+// ForwardSlashify replaces all '\' with '/' on Windows OS
```
sonmezonur, line 46: Function comments should end with a period.

strutil/strutil.go
```diff
@@ -40,3 +42,11 @@ func JSON(v interface{}) string {
...
+func ForwardSlashify(str string) string{
```
seruman, line 47: If I'm not missing anything we might be able to use [`filepath.ToSlash()` ](https://golang.org/pkg/path/filepath/#ToSlash)

> ToSlash returns the result of replacing each separator character in path with a slash ('/') character. Multiple separators are replaced by multiple slashes.

fbarotov: thanks! yes it seems filepath.ToSlash() does the job.

storage/url/url.go
```diff
@@ -47,6 +48,7 @@ type URL struct {
 
 // New creates a new URL from given path string.
 func New(s string) (*URL, error) {
+	s = strings.ReplaceAll(s, "\\", "/")
```
sonmezonur, line 51: Why we need this change?

fbarotov: in windows os, to allow commands like `cp e2e\app_test.go s3://bucket-faru-1` i.e., backslashed entry

sonmezonur: This function is used for creating both remote and local URLs. It is not ideal but both S3 and Unix systems may have `\\` in their files/directories. 

For Unix, `t\\est` is a valid pathname. I created an example directory:
```
tree t\\est
t\\est
|-- anotherfilewithbackslashes\\
`-- testfile

0 directories, 2 files
```


For S3, it is also valid:
```
s5cmd ls s3://bigtmp/
                                 DIR k\a/
                                 DIR ka/
2020/07/06 12:22:04                0 newtest
2020/07/06 12:21:40                0 test
```

I suspect if we do backlash replacement on the raw source,  we may affect original user input. For example, for the following command:

```
s5cmd cp test s3://bucket/prefixwith\\backslash/filewith\\backslash
# it will upload into bucket/prefixwith/backslash/filewith/backslash
```

I know it is an edge case (generally we don't use mixed file paths) and won't happen too much,  but still, we need to consider this case :) 


We should do this once we are sure that the OS is WINDOWS and the given URL is not REMOTE(S3 URL).  Maybe [here](https://github.com/peak/s5cmd/blob/master/storage/url/url.go#L53-L62) we may check OS and set URL accordingly. What do you think?

Note: We can add some simple tests for this scenario.

fbarotov: Thanks a lot. Thats nice observation and suggestion. will check that too.

e2e/cp_test.go
```diff
@@ -803,6 +807,55 @@ func TestCopyDirToS3(t *testing.T) {
...
+	defer workdir.Remove()
+	srcpath := workdir.Path()
+	dstpath := fmt.Sprintf("s3://%v/", bucket)
+
+	cmd := s5cmd("cp", workdir.Path()+"/", dstpath)
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+
+	fmt.Printf("src is: %s\n", srcpath)
```
sonmezonur, line 841: We can remove logs.

storage/url/url.go
```diff
@@ -157,13 +161,15 @@ func (u *URL) Dir() string {
...
-	if u.IsRemote() {
-		joinfn = path.Join
+	joinfn := path.Join
+
+	if runtime.GOOS == "windows" {
+		s = filepath.ToSlash(s)
 	}
 
 	clone := u.Clone()
 	clone.Path = joinfn(clone.Path, s)
```
sonmezonur, line 171: If we don't need `filepath.Join`anymore, we may directly use `path.Join(clone.Path, s)` here.

.github/workflows/ci.yml
```diff
@@ -34,7 +41,14 @@ jobs:
     - run: make test
 
   qa:
-    runs-on: ubuntu-latest
+    strategy:
```
igungor, line 44: @fbarotov we should remove this matrix since `qa` doesn't contain any OS specific checks.

fbarotov: got it, thanks a lot.

PR #185, opened 2020-07-10
command/{cp, mv}, storage/{s3, s3_test}: kms encryption support

### LINKED ISSUE #18: Support for kms encryption?
In the vanilla aws cli, one would copy objects that are encrypted with a kms key as follows:
`aws s3 cp s3://foo s3://bar --sse aws:kms`

Is there something equivalent in s5cmd? Or does the tool currently not support encryption?

PS
This is an amazing tool, especially the auto-completion functionality!

fbarotov:
Resolves #18


storage/s3_test.go
```diff
@@ -398,3 +405,203 @@ func TestS3Retry(t *testing.T) {
...
+	}
+
+	if rv := reflect.ValueOf(v[0]); rv.Kind() == reflect.Ptr {
+		return rv.Elem().Interface()
+	}
+
+	return v[0]
+}
+
+func setExpectedErrs(t *testing.T, expected string, got interface{}) {
```
seruman, line 426: This functions asserts error rather, could you name it accordingly?

storage/s3_test.go
```diff
@@ -398,3 +405,203 @@ func TestS3Retry(t *testing.T) {
...
+	if got == nil {
+		if expected != "" {
+			t.Errorf("Expected %q, but received %q", "", got)
+		}
+	} else if expected != got {
```
seruman, line 431: I think `else` could be removed, we could just check if they are equal or not.

storage/s3_test.go
```diff
@@ -398,3 +405,203 @@ func TestS3Retry(t *testing.T) {
...
+		})
+	}
+}
+func TestS3PutEncryptionRequest(t *testing.T) {
+	testcases := []struct {
+		name     string
+		sse      string
+		sseKeyId string
+		// expected
+		esse      string
```
seruman, line 528: > We can prepend expected to expected value vars. It would be easier to distinguish them from inputs further below.

storage/s3.go
```diff
@@ -374,14 +398,35 @@ func (s *S3) Put(
...
 		u.PartSize = partSize
 		u.Concurrency = concurrency
 	})
 
 	return err
 }
 
+// validateEncryptParams validates encryption parameters and
+// returns corresponding values.
+func validateEncryptParams(metadata map[string]string) (encryptMethod, encryptKey string, err error) {
```
seruman, line 416: Instead of doing validation against `aws:kms`, can't we just pass whatever passed as flags to SDK? `aws:kms` check is too AWS specific since we support S3 compatible services as much as we can.

storage/s3_test.go
fbarotov: in this part, since we are comparing interface{} with string, we would like to consider empty string and nil interface as being equal

command/cp.go
```diff
@@ -107,6 +107,16 @@ var copyCommandFlags = []cli.Flag{
...
 	},
+	&cli.StringFlag{
+		Name:    "sse-encrypt",
+		Aliases: []string{"sse"},
+		Usage:   "server side encryption",
+	},
+	&cli.StringFlag{
+		Name:    "key-id",
+		Aliases: []string{"kid"},
+		Usage:   "encryption key id",
```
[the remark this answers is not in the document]
fbarotov: since in the future server side encryptions other than kms might be supported, i thought using key-id might be reusable in those cases. what do you think?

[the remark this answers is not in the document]
fbarotov: since in the future other encryption methods might be supported, key-id parameter can be reused in those cases. what do you think?

storage/s3_test.go
seruman: What I meant was something like;

 ```go
func setExpectedErrs(t *testing.T, expected string, got interface{}) {
	if got == nil {
		if expected != "" {
			t.Errorf("Expected %q, but received %q", "", got)
			return
		}
	}

	if expected != got {
		t.Errorf("Expected %q, but received %q", expected, got)
		return
	}
}
```

Comment (fbarotov) [reply storage/s3_test.go:432]: oh ok, thanks!

File: command/cp.go [hunk 6]
```diff
@@ -64,6 +64,10 @@ Examples:
 
 	10. Mirror an S3 prefix to target S3 prefix
 		 > s5cmd {{.HelpName}} -n -s -u s3://bucket/source-prefix/* s3://bucket/target-prefix/
+	11. Perform KMS Server Side Encryption of the object(s) at the destination
```
aykutfarsak, line 67: Could we add one newline between usage examples as above ones?

storage/s3_test.go
```diff
@@ -1,13 +1,20 @@
...
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	urlpkg "net/url"
 	"os"
+	"reflect"
 	"testing"
 
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
```
aykutfarsak, line 14: Could you group these 2 dependencies with other aws-sdk imports?

storage/s3_test.go
```diff
@@ -398,3 +405,198 @@ func TestS3Retry(t *testing.T) {
...
+				"EncryptionKeyId":  tc.sseKeyId,
+			})
+
+			if err != nil && err.Error() == defaultReqErr {
+				return
+			}
+
+			if (err == nil || tc.expectedErr == nil) && tc.expectedErr != err {
+				t.Errorf("Expected %q, but received %q", tc.expectedErr, err)
+			} else if err.Error() != tc.expectedErr.Error() {
```
aykutfarsak, line 516: IMO, no need to `else if`, separate an `if` block is more readable (same on line 597)

storage/s3_test.go
```diff
@@ -398,3 +405,198 @@ func TestS3Retry(t *testing.T) {
...
+				"EncryptionKeyId":  tc.sseKeyId,
+			}, 1, 5242880)
+
+			if err != nil && err.Error() == defaultReqErr {
+				return
+			}
+
+			if (err == nil || tc.expectedErr == nil) && tc.expectedErr != err {
+				t.Errorf("Expected %q, but received %q", tc.expectedErr, err)
+			} else if err.Error() != tc.expectedErr.Error() {
```
aykutfarsak, line 597:
```suggestion
			}
			
			if err.Error() != tc.expectedErr.Error() {
```

storage/s3.go
```diff
@@ -693,3 +713,36 @@ func (a *writeAtAdapter) Write(p []byte) (int, error) {
...
+	m["ACL"] = acl
+}
+func (m Metadata) ACL() string {
+	return m["ACL"]
+}
+func (m Metadata) SetStorageClass(class string) {
+	m["StorageClass"] = class
+}
+func (m Metadata) StorageClass() string {
+	return m["StorageClass"]
```
seruman, line 729: I think it would be safer to check if the keys exists
```go
value, ok := m["key"]
if !ok {
	return ""
}

return value
```
There are already checks for empty string in
https://github.com/peak/s5cmd/blob/608ee4dd730041b8ba76581e65408b884ccdf3bb/storage/s3.go#L310-L311
https://github.com/peak/s5cmd/blob/608ee4dd730041b8ba76581e65408b884ccdf3bb/storage/s3.go#L315-L316

storage/s3_test.go
```diff
@@ -398,3 +403,200 @@ func TestS3Retry(t *testing.T) {
...
+
+			expectedSSE:      "aws:kms",
+			expectedSSEKeyId: "sdkjn12SDdci#@#EFRFERTqW/ke",
+		},
+		{
+			name:     "provide key without encryption flag, shall be ignored",
+			sseKeyId: "1234567890",
+		},
+	}
+	const defaultReqErr = "request was canceled by the user"
```
seruman, line 470: might be defined as type `error` instead of doing `fmt.Errorf` for each case below

storage/s3_test.go
```diff
@@ -398,3 +406,178 @@ func TestS3Retry(t *testing.T) {
...
+					StatusCode: http.StatusOK,
+					Body:       ioutil.NopCloser(strings.NewReader("")),
+				}
+
+				params := r.Params
+				sse := val(params, "ServerSideEncryption")
+				key := val(params, "SSEKMSKeyId")
+
+				if !(sse == nil && tc.expectedSSE == "") {
+					assert.Assert(t, cmp.Equal(sse, tc.expectedSSE))
```
ilkinulas, line 485:
```suggestion
					assert.Equal(t, sse, tc.expectedSSE)
```
This produces more informative error messages when tests fail.

Same for other `assert.Assert(t, cmp.Equal(.....))` usages for strings.

command/cp.go
```diff
@@ -64,6 +64,12 @@ Examples:
 
 	10. Mirror an S3 prefix to target S3 prefix
 		 > s5cmd {{.HelpName}} -n -s -u s3://bucket/source-prefix/* s3://bucket/target-prefix/
+
+	11. Perform KMS Server Side Encryption of the object(s) at the destination
+		> s5cmd -sse aws:kms {{.HelpName}} s3://bucket/object s3://target-bucket/prefix/object
```
igungor, line 69: -sse flag is part of the subcommand, not the globals. this command is gonna fail.

command/cp.go
```diff
@@ -64,6 +64,12 @@ Examples:
...
+
+	12. Perform KMS-SSE of the object(s) at the destination using customer managed Customer Master Key (CMK) key id
+		> s5cmd -sse aws:kms -sse-kms-key-id $0meR@nDomK(-Y s3://bucket/object s3://target-bucket/prefix/object
```
igungor, line 72: -sse flag is part of the subcommand, not the globals. this command is gonna fail.

igungor: let's use a boring key id :)

<your-ksm-key-id> or similar.

command/cp.go
```diff
@@ -379,9 +397,11 @@ func (c Copy) doUpload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) er
 
 	dstClient := storage.NewClient(dsturl)
 
-	metadata := map[string]string{
-		"StorageClass": string(c.storageClass),
-		"ContentType":  guessContentType(f),
+	metadata := storage.Metadata{
+		"StorageClass":     string(c.storageClass),
```
igungor, line 401: let's start chain methods to avoid handwriting these keys now. E.g.:

```go
storage.NewMetadata().
    SetStorageClass(...).
    SetContentType(...).
    SetEncryptionMethod(...).
    SetEncryptionKeyID(...)

File: storage/s3_test.go [hunk 16]
```diff
@@ -398,3 +406,180 @@ func TestS3Retry(t *testing.T) {
...
+	}
+
+	if rv := reflect.ValueOf(v[0]); rv.Kind() == reflect.Ptr {
+		return rv.Elem().Interface()
+	}
+
+	return v[0]
+}
+
+func TestS3CopyEncryptionRequest(t *testing.T) {
```
igungor, line 427: this test is identical to the `acl` test. please take a look how we could merge them.

fbarotov: since acl and kms related changes are in two different pull requests, should i merge these two PRs or should we merge after both get into master branch?

igungor: another pr is ok

CHANGELOG.md
```diff
@@ -7,11 +7,14 @@
 - Dropped storage class short codes display from default behaviour of `ls` operation. Instead, use `-s` flag with `ls`
 to see full names of the storage classes when listing objects.
 
+
 #### Features
+- Added Server-side Encryption (SSE) support for mv/cp operations. It uses customer master keys (CMKs) managed by AWS Key Management Service.
```
igungor, line 12: missing issue reference

igungor:
@fbarotov tests on macos failed. could you take a look?

README.md
```diff
@@ -117,7 +119,15 @@ $ tree
 #### Upload a file to S3
 
     s5cmd cp object.gz s3://bucket/
-
+    
+- by setting server side encryption (*aws kms*) of the file:
+    
+    `s5cmd cp -sse aws:kms -sse-kms-key-id <your-kms-key-id> object.gz s3://bucket/`
```
sonmezonur, line 125: This format does not seem identical to other examples. Can you please use code block like other examples?

```
s5cmd cp -sse aws:kms -sse-kms-key-id <your-kms-key-id> object.gz s3://bucket/
```

igungor: once this is done, i'm gonna merge this pr.

PR #186, opened 2020-07-11
command/cp: add ACL support

### LINKED ISSUE #26: Add option for ACL on cp command
Hey!

Awesome work! This is super fast.

Can you add an option for --acl bucket-owner-full-control when using cp please?

Cheers!
Josh

fbarotov:
Resolves #26


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## not released yet
 
+#### Added
+- Support for ACL flag. ([#26](https://github.com/peak/s5cmd/issues/26))
```
igungor, line 6: Could you use the same section structure as the previous release please? I mean, Bugfixes, Features, Improvements etc.

igungor: `Add [access control lists (ACLs)](https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html) support to enable managing access to buckets and objects.`

command/cp.go
```diff
@@ -107,6 +107,10 @@ var copyCommandFlags = []cli.Flag{
 		Value:   defaultPartSize,
 		Usage:   "size of each part transferred between host and remote server, in MiB",
 	},
+	&cli.StringFlag{
+		Name:  "acl",
+		Usage: "Access Control List",
```
igungor, line 112: please elaborate this sentence, just like the other flag usages.

command/cp.go
```diff
@@ -382,6 +388,7 @@ func (c Copy) doUpload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) er
 	metadata := map[string]string{
 		"StorageClass": string(c.storageClass),
 		"ContentType":  guessContentType(f),
+		"acl":          c.acl,
```
igungor, line 391: `"ACL"`

storage/s3.go
```diff
@@ -294,6 +294,30 @@ func (s *S3) listObjects(ctx context.Context, url *url.URL) <-chan *Object {
...
+func setAcl(acl **string, metadata map[string]string) error {
+	if acl == nil {
+		return fmt.Errorf("acl reference cannot point to nil")
+	}
+	aclVal := metadata["acl"]
+	if aclVal == "" {
+		return nil
+	}
+
+	supportedActions := map[string]bool{
```
igungor, line 308: s5cmd supports other S3-compatible storage services. These "supported" actions may not be available, or there could be more options in other services. Let's keep the code simple and delegate the validation to the service.

storage/s3.go
```diff
@@ -294,6 +294,30 @@ func (s *S3) listObjects(ctx context.Context, url *url.URL) <-chan *Object {
 	return objCh
 }
 
+// setAcl sets value of acl flag pointer according to the metadata
+// provided.
+func setAcl(acl **string, metadata map[string]string) error {
```
igungor, line 299: 2-level indirection doesn't look very readable to me.

While we're here, let's create a new type

`type Metadata map[string]string`

and define methods on that type.

`func (m Metadata) SetACL(acl string)`
`func (m Metadata) SetStorageClass(class string)`
`func (m Metadata) ACL() string`
`func (m Metadata) StorageClass() string`
etc.

would be more safe and readable.

fbarotov: great, thanks! will do that.

CHANGELOG.md
fbarotov: yeah, I should follow the pattern in changelog.md to keep consistency intact. thanks for pointing it out.

storage/s3.go
```diff
@@ -294,33 +294,9 @@ func (s *S3) listObjects(ctx context.Context, url *url.URL) <-chan *Object {
...
-		return fmt.Errorf("provided acl flag value is not supported")
-	}
-	*acl = aws.String(aclVal)
-	return nil
-}
-
 // Copy is a single-object copy operation which copies objects to S3
 // destination from another S3 source.
-func (s *S3) Copy(ctx context.Context, from, to *url.URL, metadata map[string]string) error {
+func (s *S3) Copy(ctx context.Context, from, to *url.URL, _metadata map[string]string) error {
```
igungor, line 299: no underscores :)

storage/s3.go
```diff
@@ -382,11 +358,12 @@ func (s *S3) Put(
 	ctx context.Context,
 	reader io.Reader,
 	to *url.URL,
-	metadata map[string]string,
+	_metadata map[string]string,
```
igungor, line 361: same

storage/s3_test.go
```diff
@@ -479,25 +475,25 @@ func TestS3AclFlagOnCopy(t *testing.T) {
 
 				aclVal := val(r.Params, "ACL")
 
-				setExpectedErrs(t, tc.eAcl, aclVal)
+				assertEqual(t, tc.expectedAcl, aclVal)
```
igungor, line 478: please use `go-cmp` if you have struct/interface assertions.

fbarotov: great, thanks :)

storage/s3.go
sonmezonur: what do you think of changing the signature of Copy and Put?

 ```
Copy(ctx context.Context, src, dst *url.URL, metadata Metadata) error
Put(ctx context.Context, src io.Reader, dst *url.URL, metadata Metadata, concurrency int, partSize int64) error
```

I don't see any reason to use `map[string]string` rather than `Metadata`

File: storage/s3.go [hunk 9]
```diff
@@ -693,3 +703,24 @@ func (a *writeAtAdapter) Write(p []byte) (int, error) {
 	a.offset += int64(n)
 	return n, nil
 }
+
+type Metadata map[string]string
```
sonmezonur, line 707: I think we should move this type and its functions under `storage/object.go`.

storage/s3.go
```diff
@@ -693,3 +703,24 @@ func (a *writeAtAdapter) Write(p []byte) (int, error) {
...
+
+func (m Metadata) SetACL(acl string) {
+	m["ACL"] = acl
+}
```
sonmezonur, line 711: Putting empty line between each function would be more readable.

storage/s3.go
fbarotov: I thought about doing that, but we will have to do modifications at Storage interface and other places which use map to pass metadata, and also map might be easier to reason about. Will make your suggested modification, seems logical. thanks

storage/s3.go
sonmezonur: sorry. I mean `storage/storage.go` :)

storage/s3_test.go
```diff
@@ -398,3 +406,162 @@ func TestS3Retry(t *testing.T) {
...
+	return v[0]
+}
+
+func TestS3AclFlagOnCopy(t *testing.T) {
+	testcases := []struct {
+		name string
+		acl  string
+
+		expectedAcl string
+		expectedErr error
```
igungor, line 433: we don't have a test case with an expected error. please remove any unused code from test tests.

storage/s3_test.go
```diff
@@ -1,15 +1,23 @@
...
 	"fmt"
+	"io"
 	"net/http"
 	urlpkg "net/url"
 	"os"
+	"reflect"
 	"testing"
 
+	"gotest.tools/v3/assert"
+
```
igungor, line 15: please group the imports.

command/cp.go
```diff
@@ -382,6 +388,7 @@ func (c Copy) doUpload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) er
 	metadata := map[string]string{
```
sonmezonur, line 388:
```go
metadata := storage.Metadata {
      "StorageClass": string(c.storageClass),
      "ContentType":  guessContentType(f),
      "ACL":          c.acl,
}
```

Same applies to [line 429](https://github.com/peak/s5cmd/blob/7da33dcd2367e3639bb263ec26c5b617c09febf2/command/cp.go#L429)

storage/storage.go
```diff
@@ -175,3 +175,41 @@ type notImplemented struct {
...
 }
+
+type Metadata map[string]string
+
+func (m Metadata) SetACL(acl string) {
+	m["ACL"] = acl
+}
+
+func (m Metadata) ACL() string {
+	acl, ok := m["ACL"]
```
sonmezonur, line 186: these controls are not needed. just use
```
func (m Metadata) ACL() string {
	return m["ACL"]
}
```

fbarotov: those checks are there following [this](https://github.com/peak/s5cmd/pull/185#discussion_r455538532) feedback. I thought the reasoning might be for readability. Shall I remove those checks?

storage/s3_test.go
```diff
@@ -398,3 +404,150 @@ func TestS3Retry(t *testing.T) {
...
+		{
+			name: "acl flag without value, flag should be ignored",
+		},
+		{
+			name:        "acl flag with a value",
+			acl:         "bucket-owner-full-control",
+			expectedAcl: "bucket-owner-full-control",
+		},
+	}
+	defaultReqErr := fmt.Errorf("request was canceled by the user")
```
ilkinulas, line 444: Let's remove `defaultReqErr` and update mocked response as :

```go
r.HTTPResponse = &http.Response{
	StatusCode: http.StatusOK,
	Body:       ioutil.NopCloser(strings.NewReader("")),
}
```

Returning  an error for each `mockS3.Copy` call is a bit confusing.

Same for `put` tests.

storage/s3_test.go
```diff
@@ -398,3 +404,150 @@ func TestS3Retry(t *testing.T) {
...
+		name string
+		acl  string
+
+		expectedAcl string
+	}{
+		{
+			name: "no acl flag",
+		},
+		{
+			name: "acl flag without value, flag should be ignored",
```
ilkinulas, line 436:
```suggestion
			name: "acl flag without value, flag should be ignored",
			acl:  "",
```

First two test cases seem identical.
Same for `put` tests.

storage/storage.go
```diff
@@ -175,3 +175,44 @@ type notImplemented struct {
...
+
+func (m Metadata) StorageClass() string {
+	stclass, ok := m["StorageClass"]
+	if !ok {
+		return ""
+	}
+	return stclass
+}
+
+func (m Metadata) SetContentType(contentType string) Metadata {
```
sonmezonur, line 207: what do you think of removing these set methods since it is not currently used? once we need them in the future, we may add them.

fbarotov: sounds logical, thanks.

command/cp.go
```diff
@@ -421,6 +428,7 @@ func (c Copy) doCopy(ctx context.Context, srcurl *url.URL, dsturl *url.URL) erro
 
 	metadata := map[string]string{
```
igungor, line 429: we should use storage.Metadata here

storage/storage.go
seruman: My bad, I forgot that map returns zero value for non-existent keys, you can remove them.

storage/s3_test.go
```diff
@@ -404,6 +409,145 @@ func TestS3Retry(t *testing.T) {
...
+			metadata := NewMetadata().SetACL(tc.acl)
+
+			err = mockS3.Copy(context.Background(), u, u, metadata)
+
+			if err != nil {
+				t.Errorf("Expected %v, but received %q", nil, err)
+			}
+		})
+	}
+}
```
sonmezonur, line 489: empty line

storage/storage.go
```diff
@@ -178,14 +178,52 @@ func (e notImplemented) Error() string {
...
+func (m Metadata) SetSSE(sse string) Metadata {
+	m["EncryptionMethod"] = sse
+	return m
+}
+
+func (m Metadata) SSEKeyId() string {
+	return m["EncryptionKeyId"]
+}
+
+func (m Metadata) SetSSEKeyId(kid string) Metadata {
```
igungor, line 226: prefer `ID` over `Id`. See https://github.com/golang/go/wiki/CodeReviewComments#initialisms

storage/storage.go
```diff
@@ -175,3 +175,55 @@ type notImplemented struct {
...
+	return m["EncryptionMethod"]
+}
+
+func (m Metadata) SetSSE(sse string) Metadata {
+	m["EncryptionMethod"] = sse
+	return m
+}
+
+func (m Metadata) SSEKeyID() string {
+	return m["EncryptionKeyId"]
```
igungor, line 223: ID

storage/storage.go
```diff
@@ -175,3 +175,55 @@ type notImplemented struct {
...
+}
+
+func (m Metadata) SetSSEKeyID(kid string) Metadata {
+	m["EncryptionKeyId"] = kid
```
igungor, line 227: ID

PR #187, opened 2020-07-11
command/ls, ese/ls_test: added an option to show storage class

### LINKED ISSUE #165: Add an option to show storage class
`s5cmd ls` show a custom abbreviation of an objects storage class. It's better to show full name of a class, instead of an abbreviation.

I'm thinking of `-s` to show **s**torage class. For example;

```
$ s5cmd ls -s s3://bucket/

2020/05/06 09:35:18               STANDARD   13 a.xml
2020/05/06 09:33:57             ONEZONE_IA   23 a.html
2020/05/06 09:34:31            STANDARD_IA   14 a.json
2020/05/06 09:35:44    INTELLIGENT_TIERING   15 a.md
```

fbarotov:
Resolves #165

### DIFF
--- a/e2e/ls_test.go
+++ b/e2e/ls_test.go
@@ -263,8 +263,8 @@ func TestListS3ObjectsAndFolders(t *testing.T) {
 		3: suffix("DIR d/"),
 		4: suffix("DIR e/"),
 		5: suffix("DIR f/"),
-		6: suffix("? 298 report.gz"),
-		7: suffix("? 302 testfile1.txt"),
+		6: suffix("298 report.gz"),
+		7: suffix("302 testfile1.txt"),
 	}, alignment(true))
 }
 
@@ -291,7 +291,7 @@ func TestListS3ObjectsAndFoldersWithPrefix(t *testing.T) {
 
 	assertLines(t, result.Stdout(), map[int]compareFunc{
 		0: suffix("DIR t/"),
-		1: suffix("? 302 testfile1.txt"),
+		1: suffix("302 testfile1.txt"),
 	}, alignment(true))
 }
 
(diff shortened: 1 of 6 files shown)


e2e/ls_test.go
```diff
@@ -139,6 +141,30 @@ func TestListSingleWildcardS3Object(t *testing.T) {
...
+	putFile(t, s3client, bucket, "testfile1.txt", "this is a file content")
+
+	cmd := s5cmd("ls -s", "s3://"+bucket+"/testfile1.txt")
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+
+	// on mock s3 client, storage class short code is represented by `?`
+	// storage class full display is an empty string.
+	stclass := !strings.Contains(result.Stdout(), "?")
```
sonmezonur, line 163: I think we should not assert the output of the storage class since it is not correctly set by gofakes3. If we add this test and assert `?` or `empty strings` for stdout, we gonna test the behavior of gofakes3 rather than s5cmd. This test can be added in the future if they support storage classes. We can just test exitcode of command and skip other assertions:

```
	cmd := s5cmd("ls", "-s", "s3://"+bucket+"/testfile1.txt")
	result := icmd.RunCmd(cmd)

```

igungor:
I think, we should not show storage class by default. `--storage-class | -s` will show the full name of the storage class. What do you think?

fbarotov:
> I think, we should not show storage class by default. `--storage-class | -s` will show the full name of the storage class. What do you think?

@igungor so you mean we shall remove storage class short codes display from default behaviour?

igungor:
> @igungor so you mean we shall remove storage class short codes display from default behaviour?

Yes. I'm asking the question to all the stakeholders :)

sonmezonur:
> I think, we should not show storage class by default. `--storage-class | -s` will show the full name of the storage class. What do you think?

It would be breaking change for ls command but I agreed. Currently, we need to keep track of each storage class added to S3 in order to add its shortcode. By removing this option, we don't need to worry about these updates. Also, I don't think shortcodes are readable and I believe it should not be part of the default output.

PR #188, opened 2020-07-11
storage/s3: fix wildcard issue (#168)

### LINKED ISSUE #168: Wildcard mv causes infinite repetition if target destination also qualifies wildcart
Suppose you have objects like this:
```
...
s3://sandbox/single-4k.blob.4
s3://sandbox/single-4k.blob.5
s3://sandbox/single-4k.blob.6
...
s3://sandbox/single-16k.blob.1
s3://sandbox/single-16k.blob.2
...
```
Trying to move them to a new prefix that also matches the glob causes infinite repetition. For example:
```
s5cmd mv 's3://sandbox/single*.blob*' s3://sandbox/single/
```
Produces this after some time:
```
...
mv s3://sandbox/single/single/single/single-4k.blob.1 s3://sandbox/single/single/single/single/single-4k.blob.1
...
```
Naive solution is to have the application make a full pass of the bucket for keys matching the wildcard, keep state of the initial list of qualified keys, and pop them off the queue as the individual `mv`s complete. I am guessing that right now it is doing a lazy evaluation of qualified keys one at a time.

For others who are experiencing this, temporary workaround right now is to pipe a `s5cmd ls` wildcard output into `xargs` and individually `s5cmd mv`.

fbarotov:
Resolves #168

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
     - run: make test
 
   qa:
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,14 +41,7 @@ jobs:
     - run: make test
 
   qa:
-    strategy:
-      matrix:
-        os:
-          - macos
-          - ubuntu
-          - windows
-
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: windows-latest
 
     steps:
     - uses: actions/checkout@v2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 
 #### Bugfixes
 
-- Fixed infinite repetition issue on mv with wildcard same for source and destination ([#168](https://github.com/peak/s5cmd/issues/168))
+- Fixed infinite repetition issue on mv/cp operations which would occur
+ if the destination matched the source wildcard. ([#168](https://github.com/peak/s5cmd/issues/168))
 - Fixed windows filepath issue, where backslashes should be treated as the path delimiter. ([#178](https://github.com/peak/s5cmd/issues/178))
 - All tests pass on windows, by converting and treating file paths to UNIX filepath format.
 - Fixed a transfer issue where the object path contains particular regex metacharacters. ([#111](https://github.com/peak/s5cmd/pull/111)) [@brendan-matroid](https://github.com/brendan-matroid)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 #### Bugfixes
 
+- Fixed infinite repetition issue on mv with wildcard same for source and destination ([#168](https://github.com/peak/s5cmd/issues/168))
 - Fixed windows filepath issue, where backslashes should be treated as the path delimiter. ([#178](https://github.com/peak/s5cmd/issues/178))
 - All tests pass on windows, by converting and treating file paths to UNIX filepath format.
 - Fixed a transfer issue where the object path contains particular regex metacharacters. ([#111](https://github.com/peak/s5cmd/pull/111)) [@brendan-matroid](https://github.com/brendan-matroid)
--- a/e2e/mv_test.go
+++ b/e2e/mv_test.go
@@ -311,28 +311,3 @@ func TestMoveMultipleS3ObjectsToS3(t *testing.T) {
 		assert.Assert(t, ensureS3Object(s3client, bucket, "dst/"+filename, content))
 	}
 }
-
-// mv s3://bucket/single*.blob* s3://bucket/single/
-func TestMoveSourceAndDestinationMatchWildcardWithTenThousandAndOneS3ObjectsToS3(t *testing.T) {
-	t.Parallel()
-
-	const numFiles = 10001
-	bucket := s3BucketFromTestName(t)
-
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 #### Bugfixes
 
+- Fixed infinite repetition issue on mv with wildcard same for source and destination ([#168](https://github.com/peak/s5cmd/issues/168))
```
sonmezonur, line 8: We also have the same issue for **cp**. If we copy/move S3 objects into sub prefixes, this issue may occur.

.github/workflows/ci.yml
```diff
@@ -41,14 +41,7 @@ jobs:
...
   qa:
-    strategy:
-      matrix:
-        os:
-          - macos
-          - ubuntu
-          - windows
-
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: windows-latest
```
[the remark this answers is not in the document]
fbarotov: thanks. since I am working on windows, to run checks locally I had put windows. Will change to linux before pushing commits.

CHANGELOG.md
fbarotov: since mv uses cp operations and cp depend on listing objects, doesn't this solve other related issues as well? I will check that.

storage/s3.go
```diff
@@ -156,6 +157,10 @@ func (s *S3) listObjectsV2(ctx context.Context, url *url.URL) <-chan *Object {
 		defer close(objCh)
 		objectFound := false
 
+		// keep track of unix timestamp, which is used not to iterate
+		// files, which have already been passed through
+		snapshotUx := time.Now().Unix()
```
igungor, line 162: let's rename this to simply `now`.

igungor: also, it's better to explicitly call `UTC()` to avoid any surprises.

storage/s3.go
```diff
@@ -234,6 +245,10 @@ func (s *S3) listObjects(ctx context.Context, url *url.URL) <-chan *Object {
...
```
igungor, line 250: same as above comments.

igungor:
Could you think of a test case to demonstrate the fix?

e2e/mv_test.go
```diff
@@ -311,3 +311,53 @@ func TestMoveMultipleS3ObjectsToS3(t *testing.T) {
...
+		cmpFuncs[i] = equals("mv s3://%v/%v %v%v", bucket, filename, dst, filename)
+	}
+	cmd := s5cmd("mv", src, dst)
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+
+	assertLines(t, result.Stdout(), cmpFuncs, sortInput(true))
+
+	fmt.Println(result.Stdout())
```
sonmezonur, line 351: we should remove logs.

fbarotov: thanks :)

e2e/mv_test.go
```diff
@@ -311,3 +311,51 @@ func TestMoveMultipleS3ObjectsToS3(t *testing.T) {
 		assert.Assert(t, ensureS3Object(s3client, bucket, "dst/"+filename, content))
 	}
 }
+
+// mv s3://bucket/single*.blob* s3://bucket/single/
+func TestMoveWildcardTwoThousandS3ObjectsToS3(t *testing.T) {
```
ilkinulas, line 316: This test does not fail on `master` branch. I was expecting a timeout when running this on master.
 Please double check if we are testing the right assumptions.

fbarotov: hmm, the example is in some sense close imitation of the one from [the related issue](https://github.com/peak/s5cmd/issues/168), but to test it I think any number of files above 1000 would do, so let me try some lower number than 2000.

e2e/mv_test.go
```diff
@@ -311,3 +311,51 @@ func TestMoveMultipleS3ObjectsToS3(t *testing.T) {
...
 	}
 }
+
+// mv s3://bucket/single*.blob* s3://bucket/single/
+func TestMoveWildcardThousandAndTenS3ObjectsToS3(t *testing.T) {
+	t.Parallel()
+
+	bucket := s3BucketFromTestName(t)
+
+	s3client, s5cmd, cleanup := setup(t)
```
sonmezonur, line 321: It took nearly 37seconds to run this test when I forked and run this test locally :) We use `boltdb` as a default backend for e2e tests. But for this test, it is more convenient to use in-memory backend since we are working with a large number of files. 
```suggestion
	s3client, s5cmd, cleanup := setup(t, withS3Backend("mem"))
```

fbarotov: wow, awesome, thanks. let me try it :)

e2e/mv_test.go
```diff
@@ -311,3 +311,28 @@ func TestMoveMultipleS3ObjectsToS3(t *testing.T) {
 		assert.Assert(t, ensureS3Object(s3client, bucket, "dst/"+filename, content))
 	}
 }
+
+// mv s3://bucket/single*.blob* s3://bucket/single/
+func TestMoveSourceAndDestinationMatchWildcardWithTenThousandAndOneS3ObjectsToS3(t *testing.T) {
```
igungor, line 316: i can't see how this test asserts that new files that are uploaded after the ListObjects request are ignored byt the List call.

fbarotov: hmm, normally, if this test is executed it will lead to infinite loop/repetition, so we get timeout error; but following the changes the test case terminates, meaning files uploaded after ListObjects call are ignored. Otherwise, even if one is not ignored it will cause infinite repetition.

PR #189, opened 2020-07-13
Retry on closed connection

eminugurkenar:
We upload large amount of data to s3 and after upgrading version 1.0.0 ,  we experienced failures for large uploads , randomly we get errors like below.

`RequestError: send request failed
caused by: Put "https://testing.s3.amazonaws.com/xxxxx&partNumber=n&uploadId=id": write tcp 10.xx.xx.xx:41654->52.216.8.115:443: use of closed network connection`

After digging AWS SDK `ShouldRetry` method, we noticed that for some cases it does return `false` but for some cases it returns `true` even if base error says `use of closed network connection` in both cases, you can see below

`SerializationError: failed to unmarshal error message
        status code: 500, request id: IDDDDDDD, host id: 3231sdsadasdasdI=
caused by: UnmarshalError: failed to unmarshal error message
caused by: read tcp 10.xx.xx.xx:39860->52.216.166.67:443: use of closed network connection: true`

`RequestError: send request failed
caused by: Put "https://testing.s3.amazonaws.com/xxxx?partNumber=n&uploadId=id--": write tcp 10.xx.xx.xx:41654->52.216.8.115:443: use of closed network connection: false`

So we implemented this patch to retry on any error that contains `use of closed network connection` ,  as golang does not export this error `https://github.com/golang/go/issues/4373` , we had to check error msg.

### DIFF
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -627,6 +627,11 @@ func (c *customRetryer) ShouldRetry(req *request.Request) bool {
 	if errHasCode(req.Error, "InternalError") {
 		return true
 	}
+
+	if errContains(req.Error, "use of closed network connection") {
+		return true
+	}
+
 	return c.DefaultRetryer.ShouldRetry(req)
 }
 
@@ -666,6 +671,22 @@ func errHasCode(err error, code string) bool {
 	var multiUploadErr s3manager.MultiUploadFailure
 	if errors.As(err, &multiUploadErr) {
 		return errHasCode(multiUploadErr.OrigErr(), code)
+	}
+
+	return false
+
+}
+
+func errContains(err error, msg string) bool {
+	if err == nil || msg == "" {
+		return false
+	}
+
+	var awsErr awserr.Error
+	if errors.As(err, &awsErr) {
+		if strings.Contains(awsErr.Error(), msg) {
+			return true
+		}
 	}
 
 	return false
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -690,7 +690,6 @@ func errContains(err error, msg string) bool {
 	}
 
 	return false
-
 }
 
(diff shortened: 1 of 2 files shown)


storage/s3.go
```diff
@@ -627,6 +627,11 @@ func (c *customRetryer) ShouldRetry(req *request.Request) bool {
 	if errHasCode(req.Error, "InternalError") {
 		return true
 	}
+
+	if errContains(req.Error, "use of closed network connection") {
```
igungor, line 631: could you update S3Retry test?

storage/s3.go
```diff
@@ -672,6 +677,22 @@ func errHasCode(err error, code string) bool {
...
+
+	var awsErr awserr.Error
+	if errors.As(err, &awsErr) {
+		if strings.Contains(awsErr.Error(), msg) {
+			return true
+		}
+	}
+
+	return false
+
```
igungor, line 693: remove trailing newline

storage/s3.go
eminugurkenar: updated

storage/s3.go
eminugurkenar: removed

storage/s3_test.go
```diff
@@ -283,6 +283,10 @@ func TestS3Retry(t *testing.T) {
 			name: "RequestError",
 			err:  awserr.New(request.ErrCodeRequestError, "request error", nil),
 		},
+		{
+			name: "RequestError",
+			err:  awserr.New(request.ErrCodeRequestError, "use of closed network connection", nil),
```
igungor, line 288: could you use a unique test name please?

eminugurkenar: updated

PR #190, opened 2020-07-13
command/{expand, expand_test}, e2e/{cp, rm, run}: fixed rm operation issue on non-existent file

### LINKED ISSUE #69: Remove operation on non-existent file return success
```
tmp  Δ aws s3 ls 's3://bucket/notfound'
tmp  Δ echo $?
1
tmp  Δ s5cmd rm 's3://bucket/notfound'
2020/02/25 13:08:38 +OK "rm s3://bucket/notfound"
tmp  Δ echo $?
0
tmp  Δ
```

Removing a non-existent object returns `0` exit code. Also the log is misleading. It says worker is finished successfully but we're not interested in that actually.

I expect the status code a non-zero value. Also I don't need an `+OK` log cause it's not what it says.

fbarotov:
Resolves #69

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,14 +41,7 @@ jobs:
     - run: make test
 
   qa:
-    strategy:
-      matrix:
-        os:
-          - macos
-          - ubuntu
-          - windows
-
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: windows-latest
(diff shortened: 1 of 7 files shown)


igungor:
This issue requires some thinking. Also it's a breaking change. Let's not merge this. Could you close this PR @fbarotov ?

PR #192, opened 2020-07-15
readme.md: added benchmarks section

fbarotov:
not linking to the related issue, since it may require much more updates and thus, may not necessarily meet the expectations of readme file


README.md
```diff
@@ -277,6 +294,30 @@ ERROR "cp s3://somebucket/file.txt file.txt": object already exists
...
     }
 ```
+## Benchmarks
+Some benchmarks regarding the performance of `s5cmd` are introduced below. For more
+details refer to this [post](https://medium.com/@joshua_robinson/s5cmd-for-high-performance-object-storage-7071352cc09d)
+which is the source of the benchmarks to be presented.
+
+*Upload/download of single large file*
+
+![get/put performance graph](https://miro.medium.com/max/770/0*aWOwrmXJN9XXUHf9)
```
aykutfarsak, line 304: Could we put the images to s5cmd repository in order not to be affected by the deletion of them from the source?

fbarotov: nice point, thanks.

igungor:
rebase please.

sonmezonur:
can we resize benchmark images and make them a little bit smaller? :D

PR #194, opened 2020-07-17
command/cp: add cross-region transfer support

### LINKED ISSUE #155: Add cross region transfer support
Currently s5cmd doesn't support cp/mv operations between different s3 buckets. We need a way to specify the source region via a flag.

`awscli` provides `--source-region <value>` flag to support this feature.

fbarotov:
Resolves #155


command/app.go
```diff
@@ -19,6 +19,10 @@ const (
 	appName = "s5cmd"
 )
 
+// AppStorageOptions will be overridden by inner command flags
+// such as if provided `cp -region` will override s5cmd -region flag value.
+var AppStorageOptions storage.StorageOptions
```
sonmezonur, line 24: Instead of keeping global storage options in `command/app.go`, I think it is better to store them in `storage` package. You may pass these options by using` storage.Init(s3opts)`.

command/app.go
```diff
@@ -55,6 +59,14 @@ var app = &cli.App{
 			Name:  "install-completion",
 			Usage: "install completion for your shell",
 		},
+		&cli.StringFlag{
```
sonmezonur, line 62: These flags should not be part of global options since it is not used by all subcommands. It is only required by `cp`. Other commands including `ls`, `du`, `rm` will not use these options.

command/cat.go
```diff
@@ -67,11 +67,14 @@ var catCommand = &cli.Command{
 
 // Cat prints content of given source to standard output.
 func Cat(ctx context.Context, src *url.URL) error {
-	client := storage.NewClient(src)
+	client, err := storage.NewClient(src, AppStorageOptions)
```
sonmezonur, line 70: After passing global options to storage package you may use:

```
client, err := storage.NewClient(src)
```

storage/s3.go
```diff
@@ -32,6 +32,11 @@ var _ Storage = (*S3)(nil)
 
 var sentinelURL = urlpkg.URL{}
 
+var (
+	mutex            = &sync.Mutex{}
```
sonmezonur, line 36: If you use sync.Mutex for session management, read operations will block each other. We should allow multiple goroutines to access same session at the same time. Please try to use [sync.RWMutex ](https://golang.org/pkg/sync/#RWMutex)

storage/s3.go
```diff
@@ -32,6 +32,11 @@ var _ Storage = (*S3)(nil)
...
+	mapS3optsSession = map[S3Options]*session.Session{}
```
sonmezonur, line 37: Instead of keeping global mutex and map, we can create a new type like:
```go
type Session struct{
      sync.RWMutex
      sessions map[S3Options]*session.Session{}
}

func (s Session) Read(opts S3Options) *session.Session {
     s.RLock()
     defer s.RUnlock()
     return s.sessions[opts]
}

func (s *Session) Write(opts S3Options) error {
      s.Lock()
      defer s.Unlock()

      awsSession, err := newSession(opts)
      if err != nil {
	  return err
      }

      s.sessions[opts] = awsSession
      return nil
}

func (s *Session) Clear() {
        s.Lock()
	defer s.Unlock()
	s.sessions = map[S3Options]*session.Session{}
}
```

storage/s3.go
```diff
@@ -44,23 +49,14 @@ const (
...
 type S3 struct {
-	api         s3iface.S3API
-	downloader  s3manageriface.DownloaderAPI
-	uploader    s3manageriface.UploaderAPI
-	endpointURL urlpkg.URL
+	api           s3iface.S3API
+	downloader    s3manageriface.DownloaderAPI
+	uploader      s3manageriface.UploaderAPI
+	endpointURL   urlpkg.URL
+	destinationS3 *S3
```
sonmezonur, line 59: Why do we need to embed destination S3? If the purpose here is to use destination session on [S3 Copy operation](https://github.com/peak/s5cmd/blob/7d53d22338a8315e28942bb5e44214e3cf013c41/storage/s3.go#L342), that means we can safely create destination client rather than source client on [doCopy()](https://github.com/peak/s5cmd/blob/master/command/cp.go#L446)

storage/s3.go
```diff
@@ -89,22 +85,48 @@ func parseEndpoint(endpoint string) (urlpkg.URL, error) {
...
 		downloader:  s3manager.NewDownloader(awsSession),
 		uploader:    s3manager.NewUploader(awsSession),
 		endpointURL: endpointURL,
+	}
+	if opts.SourceRegion == "" || opts.SourceRegion == aws.StringValue(awsSession.Config.Region) {
+		s3Storage.destinationS3 = s3Storage
+		return s3Storage, nil
+	}
+
+	dstSession, err := newSession(S3Options{
```
sonmezonur, line 115: As I wrote in the above comment, I could not understand why we need to create two sessions with one `NewS3Storage()` call.

command/cp.go
```diff
@@ -113,6 +119,14 @@ var copyCommandFlags = []cli.Flag{
 		Value:   defaultPartSize,
 		Usage:   "size of each part transferred between host and remote server, in MiB",
 	},
+	&cli.StringFlag{
+		Name:  "source-region",
+		Usage: "connect to a specific region of the remote object storage service",
+	},
+	&cli.StringFlag{
```
sonmezonur, line 126: Why two flags are needed for cross-region support? We either create source or destination client. `awscli` only has `          [--source-region <value>]`  flag to support this functionality. Maybe you may want to investigate how they achieve this.

sonmezonur (changes requested):

command/app.go
fbarotov: i thought this might be useful for scenarios where we are running in batch mode and we have multiple copy commands, with all but a few having different source and destination buckets. so user will enter app-level flags and will override for specific ones if desired

storage/s3.go
fbarotov: hmm, here if session has already been created it immediately returns. if we use rwmutex, two different goroutines might read the map at the 'same' time and if the map doesn't contain session, each will try to create new session one after the other

command/app.go
sonmezonur: Hmm. It would be useful but doesn't it mean we should add this flag to run command rather than app options?

storage/s3.go
fbarotov: hmm, i thought if we don't keep source and destination clients attached we might need to pass destination options for each function call from upper level command down to Copy() call.

storage/s3.go
fbarotov: so, if multiple commands are entered using batch mode, for each copy/move operation we would need to know source and destination clients

command/cp.go
fbarotov: sorry for the confusion, similar flag options are [here](https://github.com/aws/aws-cli/issues/2274#issuecomment-258010608). Normally, by default s5cmd takes region as us-east-1 by default, so if some other region is desired we can provide using --source-region flag. it can be useful for scenario where we have many operations and total number of unique regions (source+destination) are more than 2. e.g, (us-west-1 -> us-north-1) and (eu-east-1 -> eu-south-2)

command/app.go
fbarotov: ooh, that's more logical and the way to go. thanks

storage/s3.go
sonmezonur: 👍

storage/s3.go
```diff
@@ -71,6 +64,21 @@ type S3Options struct {
 	NoVerifySSL bool
 }
 
+// NewS3Options returns new S3Options object by extracting
+// its fields from the provided context.
+func NewS3Options(c *cli.Context, isSrc bool) S3Options {
+	region := c.String("source-region")
+	if !isSrc && c.String("region") != "" {
+		region = c.String("region")
```
sonmezonur, line 72: It would be better if we do not pass `cli.context` to storage package.

storage/s3.go
```diff
@@ -587,9 +595,28 @@ func (s *S3) MakeBucket(ctx context.Context, name string) error {
 	return err
 }
 
+// Session holds session.Session according to s3Opts
+// and it synchronizes access/modification.
+type Session struct {
```
sonmezonur, line 600: Can we make this type unexported since it is only used internally and not called by another package?

sonmezonur (commented):
we should add multiple e2e tests

command/cp.go
sonmezonur: The reason why this part confused me is that we can already define source aws region with either environment variables or aws profile. That is why I thought it would be enough to add only one flag for the destination region of S3 object.

```
AWS_REGION=ap-northeast-2 s5cmd ls s3://asiatest
AWS_REGION=ap-northeast-2 s5cmd cp --region us-east-1 s3://asiatest/object s3://as
```

or
```
// .aws/config
[asia]
region=ap-northeast-2 
[east]
region=us-east-1
```

I guess this flag is only needed for batch mode since all operations will share same AWS profile and environment variables if we don't provide an optional flag for source region. Thanks for your clarification. Could you please add e2e test and update the documentation?

command/run.go
```diff
@@ -31,11 +31,22 @@ Examples:
...
+		Name:  "region",
+		Usage: "[default] region of the destination bucket for cp/mv operations; defaulted to source-region",
+	},
+}
 
 var runCommand = &cli.Command{
 	Name:               "run",
 	HelpName:           "run",
 	Usage:              "run commands in batch",
+	Flags:              runCommandFlags, // to be passed to children subcommands
```
sonmezonur, line 49: Have you tried these flags for run command? I suspect it won't work because the context of run command is actually not the parent context of other subcommands. We parse the given file and create a new context with new flagset ([ref](https://github.com/peak/s5cmd/blob/master/command/run.go#L121)).  You might not be able to access through
```
c.String("region")
c.String("source-region")
```
Can you try and confirm that it works as expected?

fbarotov: Yeah, it seems `run` command flags do get passed into subcommands. I did some manual tests using command line with aws s3. Ex:

 -  go run main.go run commands.txt

ERROR "ls s3://bucket-test-3/": AuthorizationHeaderMalformed: The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'us-west-2' status code: 400, request id: FED09052D9A38085, host id: Tb8h/jUhVtBZd/Yr1OAMt316k4RE8p37sjG54JfMeQ+tqXNtfsiFfbwcLmSlStRbwoL8lcd7rC4=

- go run main.go run --source-region=us-west-2 commands.txt

    2020/07/28 12:53:38              1082 LICENSE
2020/07/18 11:50:25               938 go.mod

command/cp.go
```diff
@@ -150,11 +164,12 @@ var copyCommand = &cli.Command{
 			flatten:          c.Bool("flatten"),
 			followSymlinks:   !c.Bool("no-follow-symlinks"),
 			storageClass:     storage.StorageClass(c.String("storage-class")),
-			concurrency:      c.Int("concurrency"),
```
sonmezonur, line 153: These flags are still needed. I guess you removed them by mistake.

fbarotov: Oh yeah, thanks for pointing it out. In previous commits, these values were handled using different wrapper type, after removing them I should have restored these.

command/run.go
fbarotov: sorry, you were right. flags do get transferred to child contexts from parents, but if same flags are used they do not.

command/app.go
```diff
@@ -122,3 +114,28 @@ func Main(ctx context.Context, args []string) error {
 
 	return app.RunContext(ctx, args)
 }
+
+// s3opts returns new S3Options object by extracting
+// its fields from the provided context.
+func s3opts(c *cli.Context, isSrc bool) storage.S3Options {
```
ilkinulas, line 120: We can remove the boolean parameter by introducing another function:

```go
func s3opts(c *cli.Context) storage.S3Options {
	region := c.String("source-region")
	if region == "" {
		region = c.String("default-source-region")
	}
	return storage.S3Options{
		MaxRetries:  c.Int("retry-count"),
		Endpoint:    c.String("endpoint-url"),
		NoVerifySSL: c.Bool("no-verify-ssl"),
		Region:      region,
	}
}

func dstS3opts(c *cli.Context) storage.S3Options {
	dstRegion := c.String("region")
	if dstRegion == "" {
		dstRegion = c.String("default-region")
	}
	return storage.S3Options{
		MaxRetries:  c.Int("retry-count"),
		Endpoint:    c.String("endpoint-url"),
		NoVerifySSL: c.Bool("no-verify-ssl"),
		Region:      dstRegion,
	}
}
```

This way it will be more readable on the usage side (with the cost of a little code duplication)
```go
opts := s3opts(c, false)
// vs
opts := dstS3opts(c)
```
I prefer readability.

storage/s3.go
```diff
@@ -593,9 +584,28 @@ func (s *S3) MakeBucket(ctx context.Context, name string) error {
...
 
+// s3Session holds session.Session according to s3Opts
+// and it synchronizes access/modification.
+type s3Session struct {
+	sync.Mutex
+	sessions map[S3Options]*session.Session
+}
+
+// Sessions returns allSessions singleton.
+func Sessions() *s3Session {
```
ilkinulas, line 595: Do we need this function? `allSessions` is only accessed from the same package.

fbarotov: I think we don't need, will remove it. I put there just in case we needed for e2e testing, but its not needed.

storage/s3.go
```diff
@@ -658,9 +663,27 @@ func newSession(opts S3Options) (*session.Session, error) {
 		sess.Config.Region = aws.String(endpoints.UsEast1RegionID)
 	}
 
+	s.sessions[opts] = sess
+
 	return sess, nil
 }
 
+// NumOfSessions returns number of sessions currently active.
+func NumOfSessions() int {
```
sonmezonur, line 672: If `NumOfSessions()` and `ClearSessions()` are only intended for TestNumOfSessions test, I think it is better to remove them :) Without them, the code will be much more readable and clean.

fbarotov: thanks. they are also used in `TestNewSessionWithRegionSetViaEnv`, when all tests are run together one session created by default region blocks another test from creating a new session with empty region (and set via environment variable).
so, shall i remove them?

sonmezonur: I am not sure. You may keep these tests if you wish but since these functions are only needed for test suites, I think they should not be part of `storage/s3.go`, maybe they can be added to `storage/s3_test.go`. Also, clearing the global session from test suite is not elegant. If we decide to run these tests parallel in the future, tests will not work since they both share the same global session variable (one may remove session while other trying to access). Unfortunately, I don't have any good solution for now. We may either remove these tests or find a better way to keep separate (nonglobal) session for each test.

sonmezonur: I am not sure. You may keep these tests if you wish but since these functions are only needed for test suites, I think they should not be part of `storage/s3.go`, maybe they can be added to `storage/s3_test.go`. Also, clearing the global session from test suite is not elegant. If we decide to run these tests parallel in the future, tests will not work since they both share the same global session variable (one may remove session while other trying to access). Unfortunately, I don't have any good solution for now. We may either remove these tests or find a better way to keep separate (nonglobal) session for each test.

ilkinulas (approved):

command/mv.go
```diff
@@ -46,16 +46,20 @@ var moveCommand = &cli.Command{
 			dst:          c.Args().Get(1),
 			op:           c.Command.Name,
 			fullCommand:  givenCommand(c),
-			deleteSource: true, // delete source
+			deleteSource: true, // don't delete source
```
igungor, line 49: don't?

storage/s3_test.go
```diff
@@ -79,27 +79,6 @@ func TestNewSessionPathStyle(t *testing.T) {
 	}
 }
 
-func TestNewSessionWithRegionSetViaEnv(t *testing.T) {
```
igungor, line 82: why did we remove this test?

storage/s3_test.go
```diff
@@ -699,3 +678,82 @@ func TestS3listObjectsV2(t *testing.T) {
...
+func TestSessionCreateAndCachingWithDifferentRegions(t *testing.T) {
+	const bucketRegionPath = "CreateBucketConfiguration.LocationConstraint"
+
+	testcases := []struct {
+		region         string
+		expectedRegion string // to check if `create session` request with specific region was executed
+
+		alreadyCreated bool // sessions should not be created again if they already have been created before
+	}{
+		{},
```
igungor, line 691: does an empty test have a meaning?

storage/s3.go
```diff
@@ -24,14 +24,15 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager/s3manageriface"
-
```
igungor, line 27: don't remove this newline :)

command/cp.go
```diff
@@ -113,6 +119,14 @@ var copyCommandFlags = []cli.Flag{
 		Value:   defaultPartSize,
 		Usage:   "size of each part transferred between host and remote server, in MiB",
 	},
+	&cli.StringFlag{
+		Name:  "source-region",
+		Usage: "connect to a specific region of the remote object storage service",
+	},
+	&cli.StringFlag{
+		Name:  "region",
```
igungor, line 127: Why do we need this flag? I still don't understand why do we need to "attach" source and destination sessions.

igungor (changes requested):

storage/s3_test.go
fbarotov: It was removed following https://github.com/peak/s5cmd/pull/194#discussion_r462100909 

Normally, environment variables are read when region is not provided explicitly. When we run all tests together, sessions created, and thus cached, for one test affects the other. For this test case, since there is already a session with empty/default region created, session is not recreated by reading environment variable; making the test case fail.

Should we add an option to remove cached sessions?

storage/s3_test.go
fbarotov: It is used to create session with default region, for when it is not provided. should we add comments or make it more explicit?

storage/s3.go
fbarotov: when we run `go fmt`, it removes this empty line

command/cp.go
fbarotov: when running in batch mode, say `s5cmd run commands.txt` where `commands.txt` consists of 

`cp s3://bucket/region_X/ s3://bucket/region_Y/`
`cp s3://bucket/region_K/ s3://bucket/region_P/`

since we have buckets in four different regions, I thought maybe we should somehow specify those.

command/app.go
```diff
@@ -4,12 +4,13 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/peak/s5cmd/storage"
```
sonmezonur, line 7: it would be better if we group imports :)

command/cp.go
```diff
@@ -70,6 +70,7 @@ Examples:
 
 	12. Perform KMS-SSE of the object(s) at the destination using customer managed Customer Master Key (CMK) key id
 		> s5cmd {{.HelpName}} -sse aws:kms -sse-kms-key-id <your-kms-key-id> s3://bucket/object s3://target-bucket/prefix/object
+
```
sonmezonur, line 73: we can remove empty line

storage/s3.go
```diff
@@ -658,9 +679,27 @@ func newSession(opts S3Options) (*session.Session, error) {
 		sess.Config.Region = aws.String(endpoints.UsEast1RegionID)
 	}
 
+	// get region of the bucket and create session accordingly
+	if opts.bucket != "" {
+		region, err := s3manager.GetBucketRegion(context.Background(), sess, string(opts.bucket), "")
```
sonmezonur, line 684: Since creating sessions needs a network call after this update, we should use the main context instead of `context.Background()` (requests should be cancellable). You can update NewClient function to take context from parameters.   

```
func NewClient(ctx context.Context, url *url.URL) (Storage, error)
 ```

**Comment by sonmezonur** [comment]
@fbarotov Hey ✋ Sorry for the late feedback. I personally found this approach better than #209 because of 2 reasons:

1) Getting region information with API call is almost costless. It parses region information from the response after sending HEAD request ([ref](https://github.com/aws/aws-sdk-go/blob/b601f141d1a828b07251867463e42699dbfa4d74/service/s3/s3manager/bucket_region.go#L104-L149)). I wrote [simple program](https://gist.github.com/sonmezonur/b0f93edf42e36022155829daa75f6084) to benchmark GetBucketRegion call: 
 ```
 AWS_PROFILE=personal hyperfine --show-output --min-runs 20 './testgetregion --bucket asiaonurtest >/dev/null'
 Benchmark #1: ./testgetregion --bucket asiaonurtest >/dev/null
  Time (mean ± σ):     722.0 ms ±  38.0 ms    [User: 57.9 ms, System: 23.6 ms]
  Range (min … max):   690.4 ms … 833.6 ms    20 runs
  ```
  I think it is quite negligible since we make this call once and use the cached session afterward.  

2) IMHO the code will be more readable on this PR. In #209, each command needs to parse and handle retryable errors separately (e.g. [1](https://github.com/peak/s5cmd/blob/88ed840a35d2cfa4b5e2cec069dc6f6b9c2254c0/storage/s3.go#L113-L118), [2](https://github.com/peak/s5cmd/blob/88ed840a35d2cfa4b5e2cec069dc6f6b9c2254c0/storage/s3.go#L227-L229)). While adding a new command to s5cmd, we also need to handle retryable errors in order to get region info. If we use this approach and get region info first, it will prevent nested error handling patterns.


What do you think? Dou you want to work on this PR and resolve conflicts before we conclude which implementation is better? 😸

**Comment by fbarotov** [comment]
> @fbarotov Hey ✋ Sorry for the late feedback. I personally found this approach better than #209 because of 2 reasons:
> ...

Hi @sonmezonur 🖐

Thanks a lot for the feedback. 

Awesome, I also ran [your program](https://gist.github.com/sonmezonur/b0f93edf42e36022155829daa75f6084) on my laptop and got the following results:
1. when the default session and bucket have the _same_ regions:  `411.65 +- 303.64 ms`
2. when the default session and bucket have _different_ regions: `597.10 +- 528.81 ms`

indeed, as you mentioned they are negligible.

Yes, the approach in this pr is more readable in my opinion as we would be having more loose coupling between session logic and operations logic.

Let me try to resolve the conflicts :)

File: storage/storage.go [hunk 24]
```diff
@@ -46,13 +46,21 @@ func NewLocalClient(opts Options) *Filesystem {
 	return &Filesystem{dryRun: opts.DryRun}
 }
 
-func NewRemoteClient(_ *url.URL, opts Options) (*S3, error) {
-	return newS3Storage(opts, cachedSession)
+func NewRemoteClient(ctx context.Context, url *url.URL, opts Options) (*S3, error) {
+	newOpts := Options{
+		MaxRetries:  opts.MaxRetries,
+		Endpoint:    opts.Endpoint,
+		Region:      opts.Region,
```
sonmezonur, line 53: I guess we don't have --region option anymore. Can we remove it?

storage/s3.go
```diff
@@ -678,9 +688,29 @@ func newSession(opts Options) (*session.Session, error) {
 		sess.Config.Region = aws.String(endpoints.UsEast1RegionID)
 	}
 
+	// get region of the bucket and create session accordingly
+	// if the region is not provided, it means we want region-independent session
+	// for operations such as listing buckets, making a new bucket, ...
+	if opts.bucket != "" {
```
sonmezonur, line 694: URL creation fails if the bucket is empty (https://github.com/peak/s5cmd/blob/master/storage/url/url.go#L87-L89). Since we use bucket name in URL, do we still need this check for getting region info? Can we assume bucket is always a non-zero value?

fbarotov: Since when getting a remote client, session for the bucket that the client will operate on is either created or fetched from cache, we face with issue when getting client for make-bucket operation as we would not be interacting with an existing bucket. Therefore, we have the [following change](https://github.com/peak/s5cmd/pull/194/files#diff-60ad7baa580083a1e939d08b56510758R71) which makes empty bucket receival possible.

storage/s3.go
```diff
@@ -43,19 +42,18 @@ const (
...
-var cachedSession func() *session.Session
+var cachedSessions func() *s3Session
 
 // Init creates a new global S3 session.
-func Init(opts Options) error {
-	sess, err := newSession(opts)
-	if err != nil {
-		return err
+func init() {
+	s3Sess := &s3Session{
```
sonmezonur, line 49: Why not just:

```go
func init() {
	sessionProvider = &s3Session{ // or maybe sessionFactory
		Mutex:    sync.Mutex{},
		sessions: map[Options]*session.Session{},
	}
}
```

and access it without function closure 
```go
awsSession, err := sessionProvider.newSession(ctx, opts) 
// rather than cachedSessions().newSession(ctx, opts)
```

CHANGELOG.md
```diff
@@ -3,10 +3,12 @@
 ## not released yet
 
 #### Features
-- Added global `--dry-run` option. It displays which command(s) will be executed without actually having a side effect. ([#90](https://github.com/peak/s5cmd/issues/90))
-- Added `--stat` option for `s5cmd` and it displays program execution statistics before the end of the program output. ([#148](https://github.com/peak/s5cmd/issues/148))
+
```
sonmezonur, line 6: empty line

storage/s3.go
sonmezonur: makes sense. thanks for the clarification 👍

storage/s3.go
fbarotov: thanks. I had tried to mitigate potential harms of global variable by wrapping it in a function, since for example in [this test](https://github.com/peak/s5cmd/pull/194/files#diff-2c7a8270898140e540697cb13ca5406bR91) we have to modify the variable. Wrapping the variable with function might give slight warning when making such calls. But it might not be a big issue, so I changed according to your proposition.

command/cp.go
fbarotov: since we 'infer' bucket regions and this flag has been removed, I think it is resolved :)

storage/s3_test.go
fbarotov: When resolving conflicts I undid removal of this test, but we need the [following call](https://github.com/peak/s5cmd/pull/194/files#diff-2c7a8270898140e540697cb13ca5406bR91). We can remove this test case if everyone is ok with it.

storage/s3.go
```diff
@@ -87,13 +80,16 @@ func parseEndpoint(endpoint string) (urlpkg.URL, error) {
 }
 
 // NewS3Storage creates new S3 session.
-func newS3Storage(opts Options, sessProvider func() *session.Session) (*S3, error) {
+func newS3Storage(ctx context.Context, opts Options, sessionProvider *s3Session) (*S3, error) {
```
sonmezonur, line 83:
```suggestion
func newS3Storage(ctx context.Context, opts Options) (*S3, error) {
```

Since sessionProvider is global, it can be directly accessed.

sonmezonur (approved):

command/mv.go
```diff
@@ -60,6 +60,7 @@ var moveCommand = &cli.Command{
 			ifSizeDiffer:     c.Bool("if-size-differ"),
 			ifSourceNewer:    c.Bool("if-source-newer"),
 			flatten:          c.Bool("flatten"),
+			followSymlinks:   !c.Bool("no-follow-symlinks"),
```
igungor, line 63: 👍

storage/s3.go
```diff
@@ -43,19 +42,14 @@ const (
...
 // Init creates a new global S3 session.
-func Init(opts Options) error {
-	sess, err := newSession(opts)
-	if err != nil {
-		return err
-	}
-
-	cachedSession = func() *session.Session {
-		return sess
+func init() {
```
igungor, line 48: we can initialize `sessionProvider` directly without an `init` function.

storage/s3.go
```diff
@@ -678,9 +684,29 @@ func newSession(opts Options) (*session.Session, error) {
 		sess.Config.Region = aws.String(endpoints.UsEast1RegionID)
 	}
 
+	// get region of the bucket and create session accordingly
+	// if the region is not provided, it means we want region-independent session
+	// for operations such as listing buckets, making a new bucket, ...
+	if opts.bucket != "" {
```
igungor, line 690: As far as I understand from the code, we always fetch bucket region of all given distinct buckets. Am I right?

It should be enough to make a get-bucket-region request **if** it's a transfer request (cp/mv s3->s3) **and** given urls have different bucket names.

fbarotov: thanks for the feedback. yes, we fetch bucket region for each distinct bucket. 

If we make get-bucket-region request only for transfer operations (cp/mv, as you suggested), we will still face an issue for batch operations. For instance,

`s5cmd ls | awk '{printf "ls %s/\n", $3}' | s5cmd run` will throw an error if all buckets reside on different regions.

igungor: You're handling `ls` requests for region enabled S3 buckets too. My assumption of the change was based on the PR title (cross region transfer).

fbarotov: oh, ok, thanks :)

storage/s3.go
```diff
@@ -43,19 +42,9 @@ const (
...
-	if err != nil {
-		return err
-	}
-
-	cachedSession = func() *session.Session {
-		return sess
-	}
-	return nil
+var sessionProvider = &s3Session{
+	Mutex:    sync.Mutex{},
```
igungor, line 46: Zero value of sync.Mutex is an unlocked mutex, thus creating a new value seems unnecessary.

igungor (approved):
Thanks for pursuing this.

CHANGELOG.md
```diff
@@ -3,10 +3,11 @@
 ## not released yet
 
 #### Features
-- Added global `--dry-run` option. It displays which command(s) will be executed without actually having a side effect. ([#90](https://github.com/peak/s5cmd/issues/90))
-- Added `--stat` option for `s5cmd` and it displays program execution statistics before the end of the program output. ([#148](https://github.com/peak/s5cmd/issues/148))
+- Added cross-region transfer support. Bucket regions are inferred, thus, supporting cross-region transfers and multiple regions in batch mode. ([#155](https://github.com/peak/s5cmd/issues/155))
 
 #### Improvements
+- Added global `--dry-run` option. It displays which command(s) will be executed without actually having a side effect. ([#90](https://github.com/peak/s5cmd/issues/90))
```
sonmezonur, line 9: Why we move `--dry-run` and `--stat` flags to the improvements section rather than features?

fbarotov: when this pr was made, I think `--dry-run` and `--stat` were not in the master branch. During conflict resolution their place must have been shifted. fixed it. thanks for pointing it out.

PR #199, opened 2020-07-22
all: print errors as they occur

### LINKED ISSUE #136: s5cmd rm command errors should be printed as the error happens.
I came across this behaviour when I tried to delete  thousands of s3 keys where I do not have permission to delete. (built s5cmd from master branch `1c8d8274e1f44`)

s5cmd prints out the error messages as a final step of its command execution.  If there are lots of s3 objects to delete, the error message is printed very late.

s5cmd version 0.7.0 reports the error quickly, so I can cancel the rm operation and fix the access error.

Is it possible to report rm errors as the error happens?

fbarotov:
Resolves #136

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
-#### Bugfixes
+## not released yet
+
+#### Improvements
 
 - For some operations errors were printed at the end of the program execution. Now, errors are displayed immediately after being detected. ([#136](https://github.com/peak/s5cmd/issues/136))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+#### Bugfixes
+
+- For some operations errors were printed at the end of the program execution. Now, errors are displayed immediately after being detected. ([#136](https://github.com/peak/s5cmd/issues/136))
+
 ## v1.1.0 - 22 Jul 2020
 
 With this release, Windows is supported.
--- a/command/cat.go
+++ b/command/cat.go
@@ -32,41 +32,11 @@ var catCommand = &cli.Command{
 	Usage:              "print remote object's contents to stdout",
 	CustomHelpTemplate: catHelpTemplate,
 	Before: func(c *cli.Context) error {
-		op := c.Command.Name
-		fullCommand := givenCommand(c)
-
-		if c.Args().Len() != 1 {
-			err := fmt.Errorf("expected only one argument")
-			printError(fullCommand, op, err)
-			return err
-		}
-
-		src, err := url.New(c.Args().Get(0))
-
+		err := validateCatCommand(c)
 		if err != nil {
-			printError(fullCommand, op, err)
-			return err
+			printError(givenCommand(c), c.Command.Name, err)
 		}
-
-		if !src.IsRemote() {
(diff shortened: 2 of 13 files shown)


command/cp.go
```diff
@@ -270,7 +267,7 @@ func (c Copy) Run(ctx context.Context) error {
 	waiter.Wait()
 	<-errDoneCh
 
-	return merror
+	return nil
```
sonmezonur, line 270: If you don't return an error here, exit code of all commands will be 0 (success). Old versions of s5cmd(<v1.0.0) were working like this and we changed this behavior after version 1.0.0. Please find a better way.

command/ls.go
```diff
@@ -127,19 +130,24 @@ func List(
...
-			showStorageClass: showStorageClass,
+			showEtag:         l.showEtag,
+			showHumanized:    l.humanize,
+			showStorageClass: l.showStorageClass,
 		}
 
 		log.Info(msg)
 	}
-
+	if merror != nil {
```
sonmezonur, line 146: Could not get why we need custom error with the conditional displayErr flag. If the purpose is not logging them at the end, you can simply remove [ExitErrHandler](https://github.com/peak/s5cmd/blob/master/command/app.go#L98-L101) callback and directly return  `merror` after each command.

fbarotov: Thanks! Your suggestion is elegant way of doing this and its great. It seems we would need some extra care in order not to print errors repeatedly (e.g., print in current function and print again in calling function after return).  I will do modifications accordingly.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+#### Bugfixes
```
sonmezonur, line 3: Let's rename this to `Improvements` since it is not a bug. Also, add `## not released yet`  title at the top of this section. We will replace it with `## v1.2.0 - date` (like other release logs) after the release.

command/cat.go
```diff
@@ -32,46 +32,76 @@ var catCommand = &cli.Command{
 	Usage:              "print remote object's contents to stdout",
 	CustomHelpTemplate: catHelpTemplate,
 	Before: func(c *cli.Context) error {
+		op := c.Command.Name
+		fullCommand := givenCommand(c)
+
 		if c.Args().Len() != 1 {
-			return fmt.Errorf("expected only one argument")
+			err := fmt.Errorf("expected only one argument")
+			printError(fullCommand, op, err)
```
sonmezonur, line 40: how about defining validate function and logging these validation errors in only one place just like [cp](https://github.com/peak/s5cmd/blob/cf9fb156b02fca5af419121a7cef85e665e927bd/command/cp.go#L631)? 

```go
func validate(c *cli.Context) error {
	if c.Args().Len() != 1 {
		return fmt.Errorf("expected only one argument")
	}

	src, err := url.New(c.Args().Get(0))
	if err != nil {
		return err
	}

	if !src.IsRemote() {
		return fmt.Errorf("source must be a remote object")
	}

	if src.IsBucket() || src.IsPrefix() {
		return fmt.Errorf("remote source must be an object")
	}

	if src.HasGlob() {
		return fmt.Errorf("remote source %q can not contain glob characters", src)
	}

	return nil
}

// caller
err := validate(c)
if err != nil {
	printError(givenCommand(c), c.Command.Name, err)
}
return err

```

PR #201, opened 2020-07-24
command/: add stat flag for s5cmd

### LINKED ISSUE #148: Implement stats logic
`master` branch is missing the stats logic which can be handy for error situations.

fbarotov:
Resolves #148


statutil/stat.go
```diff
@@ -0,0 +1,73 @@
...
+// StatDisplay will output statistics that has been collected so far.
+func StatDisplay() {
+	if !enabled {
+		return
+	}
+	for path, cnt := range stats[totalCount].mapStrInt64 {
+		cntSucc := stats[succCount].mapStrInt64[path]
+		cntErr := cnt - cntSucc
+		avgTime := float64(stats[timeInfo].mapStrInt64[path]) / float64(cnt)
+		fmt.Printf("=> visited %q %d time with %d ERROR and %d SUCCESS returns; average execution time: %f msec.\n",
```
sonmezonur, line 70: Please implement [log.Message](https://github.com/peak/s5cmd/blob/master/log/message.go#L12) interface and use logger package while printing stats. If you don't, the output will be mixed (will include both structured and unstructured logs). If `-json` flag is given by user, all logs including info, error, stats must be in JSON format. Otherwise, it cannot be parsed and used for tooling.

log/stat/stat.go
```diff
@@ -0,0 +1,97 @@
...
+	for i := range stats {
+		stats[i] = syncMap{
+			mu:          sync.Mutex{},
+			mapStrInt64: map[string]int64{},
+		}
+	}
+}
+
+// syncMap is a statically typed and synchronized map.
+type syncMap struct {
```
ilkinulas, line 36: Let's embed this `mutex` into syncMap
```go
type syncMap struct {
   sync.Mutex
   mapStrInt64 map[string]int64
}
```
and update `add` method to use `lock, defer unlock` pattern

```go
func (s *syncMap) add(key string, val int64) {
	s.Lock()
	defer s.Unlock()

	s.mapStrInt64[key] += val
}
```

log/stat/stat.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Stat implements log.Message interface.
+type Stat struct {
+	Visited     string  `json:"visited"`
+	SuccVisits  int64   `json:"success visits"`
+	ErrVisits   int64   `json:"error visits"`
+	AvgExecTime float64 `json:"avg. execution time"`
+}
+
+func (s Stat) String() string {
```
ilkinulas, line 55: It is better to have more readable stats output.
Current implementation dumps stats as below:
```
"newSession" 1 time with 0 ERROR and 1 SUCCESS returns; average execution time: 0.000000 msec.
"S3.Get" 11 time with 0 ERROR and 11 SUCCESS returns; average execution time: 591.909091 msec.
"Copy.doDownload" 11 time with 0 ERROR and 11 SUCCESS returns; average execution time: 592.545455 msec.
"Copy.Run" 1 time with 0 ERROR and 1 SUCCESS returns; average execution time: 1492.000000 msec.
```

I am suggesting the below tabular format: 
```
Operation        Total Error Success  Duration(ms)
==================================================
newSession           1     0       1             0
S3.Get              11     0      11           591
Copy.doDownload     11     0      11           592
Copy.Run             1     0       1          1492
```
Since this is for us (humans) it should be as readable/simple as it can. What do you think?

command/cat.go
```diff
@@ -66,12 +68,13 @@ var catCommand = &cli.Command{
 }
 
 // Cat prints content of given source to standard output.
-func Cat(ctx context.Context, src *url.URL) error {
+func Cat(ctx context.Context, src *url.URL) (err error) {
+	defer stat.Collect("command=>Cat", time.Now(), &err)()
```
ilkinulas, line 72:
```suggestion
	defer stat.Collect("Cat", time.Now(), &err)()
```

log/stat/stat.go
fbarotov: thanks, tabular format is certainly more pleasant and better

command/cat.go
```diff
@@ -66,12 +68,13 @@ var catCommand = &cli.Command{
 }
 
 // Cat prints content of given source to standard output.
-func Cat(ctx context.Context, src *url.URL) error {
+func Cat(ctx context.Context, src *url.URL) (err error) {
+	defer stat.Collect("Cat", time.Now(), &err)()
```
sonmezonur, line 72: Instead of adding stat logic into each command, why we cannot collect stats in `app.ExitErrHandler()`? Smth like:

```go
ExitErrHandler: func(c *cli.Context, err error) {
        stat.Collect(c.Command.FullName(), err)
},
```

Also, I personally found it unnecessary to calculate elapsed time for each command execution. Users may access this information through :

```
time s5cmd ...
```

I think stat mode is useful for showing the number of successful/failed operations, especially in batch mode.

log/stat/stat.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Stat is for storing a particular statistics.
+type Stat struct {
+	Visited    string `json:"visited"`
+	SuccVisits int64  `json:"success visits"`
+	ErrVisits  int64  `json:"error visits"`
+}
+
+// Collect collects function execution data.
+func Collect(path string, err *error) func() {
```
sonmezonur, line 56: Why not `Collect(path string, err error)`? Can you state why we need to return function and use pointer error?

log/stat/stat.go
```diff
@@ -0,0 +1,107 @@
...
+	for _, stat := range s {
+		fmt.Fprintf(w, "%s\t%d\t%d\t%d\t\n", stat.Visited, stat.ErrVisits+stat.SuccVisits, stat.ErrVisits, stat.SuccVisits)
+	}
+
+	w.Flush()
+	return b.String()
+}
+
+func (s Stats) JSON() string {
+	return strutil.JSON(s)
```
sonmezonur, line 87: When I test --json flag, it prints following output:

```
[{"visited":"ls","success visits":2,"error visits":0}]
```

Can we use a more readable format like {"operation": "ls", "success": 2, "error": 0}? Also, it returns an array of jsons ( single json would be better).

log/stat/stat.go
```diff
@@ -0,0 +1,107 @@
...
+type Stats []Stat
+
+func (s Stats) String() string {
+	b := bytes.Buffer{}
+
+	w := tabwriter.NewWriter(&b, 5, 0, 5, ' ', tabwriter.AlignRight)
+
+	fmt.Fprintf(w, "\n%s\t%s\t%s\t%s\t\n", "Operation", "Total", "Error", "Success")
+	fmt.Fprintf(w, "%s\t%s\t%s\t%s\t\n", "=========", "=====", "=====", "=======")
+	for _, stat := range s {
```
sonmezonur, line 78: Can you check the size of the stats before preparing a tabular format (if it is empty, we can return empty string)? If there is no stat present, it prints out the following output:

```
./s5cmd --stat
NAME:
   s5cmd - Blazing fast S3 and local filesystem execution tool

USAGE:
   s5cmd [global options] command [command options] [arguments...]

COMMANDS:
   ls       list buckets and objects
   cp       copy objects
   rm       remove objects
   mv       move/rename objects
   mb       make bucket
   du       show object size usage
   cat      print remote object's contents to stdout
   run      run commands in batch
   version  print version
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --json                         enable JSON formatted output (default: false)
   --numworkers value             number of workers execute operation on each object (default: 256)
   --retry-count value, -r value  number of times that a request will be retried for failures (default: 10)
   --endpoint-url value           override default S3 host for custom services
   --no-verify-ssl                disable SSL certificate verification (default: false)
   --log value                    log level: (debug, info, error) (default: "info")
   --install-completion           install completion for your shell (default: false)
   --stat                         collect statistics of program execution and display it at the end (default: false)
   --help, -h                     show help (default: false)

     Operation     Total     Error     Success
     =========     =====     =====     =======
```

log/stat/stat.go
sonmezonur: You can ignore it. I forgot deferred calls :)

log/stat/stat.go
fbarotov: thanks. done.

log/stat/stat.go
fbarotov: thanks. done.

log/stat/stat.go
```diff
@@ -0,0 +1,117 @@
...
+	if len(s) == 0 {
+		return ""
+	}
+
+	b := bytes.Buffer{}
+
+	w := tabwriter.NewWriter(&b, 5, 0, 5, ' ', tabwriter.AlignRight)
+
+	fmt.Fprintf(w, "\n%s\t%s\t%s\t%s\t\n", "Operation", "Total", "Error", "Success")
+	fmt.Fprintf(w, "%s\t%s\t%s\t%s\t\n", "=========", "=====", "=====", "=======")
```
seruman, line 82: IMHO it would be better to remove this separator line. Without it one can treat the output as TSV and integrate it with existing CSV tools.

log/stat/stat.go
```diff
@@ -0,0 +1,117 @@
...
+		stats[totalCount].add(path, 1)
+	}
+}
+
+// Stats implements log.Message interface.
+type Stats []Stat
+
+func (s Stats) String() string {
+	if len(s) == 0 {
+		return ""
```
seruman, line 74: Same comments as below, if it just prints headers without data, it can be treated as TSV.

command/run.go
```diff
@@ -44,7 +44,7 @@ var runCommand = &cli.Command{
 		}
 		return err
 	},
-	Action: func(c *cli.Context) error {
+	Action: func(c *cli.Context) (err error) {
```
sonmezonur, line 47: named return is not needed here.

CHANGELOG.md
```diff
@@ -2,10 +2,14 @@
 
 ## not released yet
 
+#### Features
+- Added `-stat` option for `s5cmd` and it displays program execution statistics before the end of the program output. ([#148](https://github.com/peak/s5cmd/issues/148))
```
igungor, line 6: `--stat`

e2e/app_test.go
```diff
@@ -66,3 +69,19 @@ func TestAppRetryCount(t *testing.T) {
...
+	defer cleanup()
+
+	cmd := s5cmd("--stat", "ls")
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Expected{ExitCode: 0})
+
+	out := result.Stdout()
+
+	assert.Assert(t, strings.Contains(out, "Operation"))
```
igungor, line 84: let's make sure that the output is a valid tsv. (exact match instead of substring match)

log/stat/stat.go
```diff
@@ -0,0 +1,112 @@
...
+	totalCount = iota
+	succCount
+)
+
+var (
+	enabled bool
+	stats   statistics
+)
+
+type statistics [2]syncMap
```
igungor, line 23: why 2?

log/stat/stat.go
```diff
@@ -0,0 +1,112 @@
...
+func (s *syncMap) add(key string, val int64) {
+	s.Lock()
+	defer s.Unlock()
+
+	s.mapStrInt64[key] += val
+}
+
+// Stat is for storing a particular statistics.
+type Stat struct {
+	Visited    string `json:"operation"`
```
igungor, line 51: Is it `operation` or `Visited`?  I'm confused.

log/stat/stat.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		stats[totalCount].add(path, 1)
+	}
+}
+
+// Stats implements log.Message interface.
+type Stats []Stat
+
+func (s Stats) String() string {
+	b := bytes.Buffer{}
```
igungor, line 73: `var buf bytes.Buffer`

log/stat/stat.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, stat := range s {
+		fmt.Fprintf(w, "%s\t%d\t%d\t%d\t\n", stat.Visited, stat.ErrVisits+stat.SuccVisits, stat.ErrVisits, stat.SuccVisits)
+	}
+
+	w.Flush()
+	return b.String()
+}
+
+func (s Stats) JSON() string {
+	builder := strings.Builder{}
```
igungor, line 87: `var builder strings.Builder`

log/stat/stat.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Statistics will return statistics that has been collected so far.
+func Statistics() Stats {
+	if !enabled {
+		return make([]Stat, 0)
+	}
+
+	result := make([]Stat, 0)
+	for path, cnt := range stats[totalCount].mapStrInt64 {
```
igungor, line 102: `path, total`

log/stat/stat.go
```diff
@@ -0,0 +1,112 @@
...
+		cntSucc := stats[succCount].mapStrInt64[path]
```
igungor, line 103: cntSucc -> success

log/stat/stat.go
```diff
@@ -0,0 +1,112 @@
...
+
+type statistics [2]syncMap
+
+// InitStat initializes collecting program statistics.
+func InitStat() {
+	enabled = true
+	for i := range stats {
+		stats[i] = syncMap{
+			Mutex:       sync.Mutex{},
+			mapStrInt64: map[string]int64{},
```
igungor, line 31: this name doesn't provide any information about what it actually stores.

e2e/app_test.go
```diff
@@ -3,8 +3,11 @@ package e2e
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"testing"
 
+	"gotest.tools/v3/assert"
+
```
igungor, line 10: no newline. could you use `goimports` for formatting please?

command/cat.go
```diff
@@ -38,7 +39,9 @@ var catCommand = &cli.Command{
 		}
 		return err
 	},
-	Action: func(c *cli.Context) error {
+	Action: func(c *cli.Context) (err error) {
+		defer stat.Collect(c.Command.FullName(), &err)()
+
 		src, err := url.New(c.Args().Get(0))
```
igungor, line 45: not sure if this is going to work. the `err` returned from `url.New` is actually the named return, or is it a new variable? if it's a new one, what will `stat.Collect` receive?

log/stat/stat.go
fbarotov: one is for success count and the other for total count

log/stat/stat.go
fbarotov: since there is no restriction for from where `stat collect` might be called, `visited` is probably ok option, but currently `stat collect` is being called only from within operations; `operation` seems correct one. thanks for pointing it out.

e2e/app_test.go
fbarotov: i am using `goimports`, but it does not remove empty lines and does only the sorting. thanks for pointing out.

command/cat.go
fbarotov: for named returns it seems binding occurs before the return, so for `_, e := ur.New(); return e` value of `e` will be assigned to `err` before the return, and since `stat.Collect` has pointer to `err` it will receive error `e`

log/stat/stat.go
```diff
@@ -0,0 +1,112 @@
...
+		builder.WriteString(strutil.JSON(stat) + "\n")
+	}
+
+	return builder.String()
+}
+
+// Statistics will return statistics that has been collected so far.
+func Statistics() Stats {
+	if !enabled {
+		return make([]Stat, 0)
```
sonmezonur, line 98:
```go
return Stats{}
```

log/stat/stat.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	result := make([]Stat, 0)
```
sonmezonur, line 101:
```go
var result Stats
```

command/cat.go
igungor: 👍

PR #209, opened 2020-08-12
command/cp: added cross-region transfer support

### LINKED ISSUE #155: Add cross region transfer support

fbarotov:
Resolves #155


command/rm.go
```diff
@@ -71,7 +71,10 @@ func Delete(
 	}
 	srcurl := srcurls[0]
 
-	client := storage.NewClient(srcurl)
+	client, err := storage.NewClient(srcurl)
```
fbarotov, line 74: if we have `s5cmd rm s3://bucket/pre/*` operation, this client will be used first to list the remote objects and using the same client `multi-delete` command will be issued. If `bucket` resides in a wrong region, we will get and handle `region error` twice inside client storage.

one solution might be to keep a map of client pointer to another client pointer, and at each client operation call, use that map to avoid making s3 requests which would fail with wrong region error.

fbarotov: or we can change signature of `func expandSources(ctx context.Context, client storage.Storage, followSymlinks bool, srcurls ...*url.URL)` as to accept `*url.URL` and create the client itself rather than taking `storage.Storage`. it will solve our problem without changing consistency of the code.

storage/s3.go
```diff
@@ -767,3 +917,21 @@ func (a *writeAtAdapter) Write(p []byte) (int, error) {
 	a.offset += int64(n)
 	return n, nil
 }
+
+// getRegion extracts correct region of the bucket
+// from the provided error.
+func getRegion(err error) (string, error) {
+	if awsErr, ok := err.(s3.RequestFailure); ok {
+
+		re := regexp.MustCompile("expecting '.*'")
```
sonmezonur, line 926: Is there any other way to get the region from an error message? I think this approach is not safe. Any error may match with `expecting  `.*'` regex (it is too generic).

command/rm.go
fbarotov: for rm operation `s5cmd` first lists all objects concurrently from the given urls, and makes multi-delete request, which can carry up to 1000 urls of objects to be deleted. so let's say we have 

`s5cmd rm s3://bucket_at_region_X/* s3://bucket_at_region_Y/*`

here we face the problem of grouping urls from different bucket regions and use of one client to perform multi-delete operations.

storage/s3.go
fbarotov: yes, indeed this regex is too generic, I will try to find a better way. thanks for pointing it out.

command/rm.go
fbarotov: last commit addresses the above issue(s), but since it slightly changes structure of `rm` operation, it can be considered as experimental and we can undo the last commit.

fbarotov: this issue exists in the [other](https://github.com/peak/s5cmd/pull/194) cross-region transfer PR as well.

igungor:
Closing the issue since we decided to go with #194.

command/rm.go
fbarotov: @igungor @sonmezonur should we consider this issue as separate pr or should we ignore it, assuming commands similar to 

`s5cmd rm s3://bucket_at_region_X/* s3://bucket_at_region_Y/*`

will not be given to `s5cmd` in practice.

sonmezonur: I am not sure if we should allow variadic removal operations with multiple buckets. Because I don't see any advantage of using 
```
s5cmd rm s3://bucket_at_region_X/* s3://bucket_at_region_Y/*
```
instead of 
```
s5cmd rm s3://bucket_at_region_X/* 
s5cmd rm s3://bucket_at_region_Y/*
```

sonmezonur: For the same bucket usage, it will increase performance since it will make single batch delete request. (explained in [advanced usage section](https://github.com/peak/s5cmd#beast-mode-s5cmd))
```
s5cmd rm s3://bucket_at_same_region/a/* s3://bucket_at_same_region/b/*
```
But if different buckets are given, we already need to create a storage client for each bucket and objects cannot be combined in a single batch request. That is why I think we should not allow users to prompt multiple buckets in a single `rm` command.

fbarotov: hmm, understood. that is very logical, thanks for clarification :)

sonmezonur: We should update the validation logic for rm command (if different buckets given, we should raise an error). Happy to see your PR if you are interested :)

fbarotov: yeah, of course, asap i will try to :)

PR #210, opened 2020-08-15
storage: add --dry-run option

### LINKED ISSUE #90: add --dry-run option
Hi, 
it would be nice if you could add `--dry-run` to the command line options.

fbarotov:
Resolves #90


command/app.go
```diff
@@ -55,9 +55,14 @@ var app = &cli.App{
 			Name:  "install-completion",
 			Usage: "install completion for your shell",
 		},
+		&cli.BoolFlag{
+			Name:  "dry-run",
```
sonmezonur, line 59: why --dry-run is an app flag? not all operations have a side effect. we discussed it before but I could not find a reference to our previous discussion.

storage/storage.go
```diff
@@ -19,6 +20,8 @@ var (
 
 	// ErrNoObjectFound indicates there are no objects found from a given directory.
 	ErrNoObjectFound = fmt.Errorf("no object found")
+
+	optionSingle Options
```
sonmezonur, line 24: Can you please try not to keep any global S3 option? Currently, we have both global and non-global options mixed. I think it is quite confusing and difficult to maintain.

storage/s3.go
```diff
@@ -366,17 +379,8 @@ func (s *S3) Get(
 	concurrency int,
 	partSize int64,
 ) (int64, error) {
-	if concurrency == 1 {
```
sonmezonur, line 369: Why this block is removed?

storage/s3.go
```diff
@@ -751,19 +772,3 @@ func errContains(err error, msg string) bool {
 func IsCancelationError(err error) bool {
 	return errHasCode(err, request.CanceledErrorCode)
 }
-
-// writerAtAdapter is an 'io.Writer' adapter for 'io.WriterAt'.
-type writeAtAdapter struct {
```
sonmezonur, line 756: Why this block is removed? Cat command uses this struct.

command/app.go
fbarotov: we had discussed it with @ilkinulas that making dry-run global might be better when moving its implementation to storage. For example, someone might want to perform `s5cmd --dry-run run commands.txt` and it encompasses all subcommands that might have a side-effect

storage/storage.go
fbarotov: since `dry-run` is global, i think our option is either to merge `--dry-run` with S3 struct or keep a global option. For `cross-region transfer support` if we will have to keep global options, merging `--dry-run` with those might be better than carrying it with each client. do you have suggestions?

storage/s3.go
fbarotov: I did some refactoring so that passing all storage logic to `storage` is done while keeping interface size as small as possible

storage/s3.go
fbarotov: it was being used only for one scenario with `storage.Get()` where is concurrency is 1, since we took that into a seperate function, this was not necessary to keep

storage/storage.go
sonmezonur: In [cross-region PR ](https://github.com/peak/s5cmd/pull/194), you proved that commands can access the flag of parent contexts.

> For cross-region transfer support if we will have to keep global options, merging --dry-run with those might be better than carrying it with each client

I honestly think the opposite. Why do you think this is bad:

```
client := storage.NewClient(src, s3opts)
```

We may discuss it with other contributors to decide which approach is better :D

fbarotov: thanks for your suggestion and help. we discussed it with others and decided to do as you suggested - pass --dry-run flag value as an argument rather than keeping a global variable.

command/mv.go
```diff
@@ -64,6 +64,13 @@ var moveCommand = &cli.Command{
 			encryptionMethod: c.String("sse"),
 			encryptionKeyID:  c.String("sse-kms-key-id"),
 			acl:              c.String("acl"),
+
+			storageOpts: storage.Options{
```
ilkinulas, line 68:
```suggestion
			storageOpts: NewStorageOpts(c),
```
command/app.go:
```go
func NewStorageOpts(c *cli.Context) storage.Options {
	return storage.Options{
		MaxRetries:  c.Int("retry-count"),
		Endpoint:    c.String("endpoint-url"),
		NoVerifySSL: c.Bool("no-verify-ssl"),
		DryRun:      c.Bool("dry-run"),
	}
}
```

ilkinulas (approved):
LGTM

storage/s3.go
sonmezonur: thanks. I overlooked `storage.Open()` implementation.

command/cp.go
```diff
@@ -367,7 +380,14 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL)
 		return err
 	}
 
-	f, err := os.Create(dsturl.Absolute())
+	r, err := dstClient.Make(ctx, storage.MakeOpts{
```
igungor, line 383: MakeOpts is unnecessary. `path` is the mandatory argument, and only `directory` is optional. This is better:

`func (fs *Filesystem) Make(ctx context.Context, path string, isDirectory bool)`

Another thing to not is that, I still don't like this method because it does 3 whole different things, and for only 1 scenario caller has to call Close method. I'd prefer 3 methods, each one doing one thing (and seem obvious from the signature). Let's merge this as it is and talk about it for the next one.

storage/storage.go
```diff
@@ -227,3 +240,27 @@ func (m Metadata) SetSSEKeyID(kid string) Metadata {
...
 }
+
+type ReadCloserFile struct {
+	rc io.ReadCloser
+	f  *os.File
+}
+
+func (rcf *ReadCloserFile) ReadCloser() (io.ReadCloser, error) {
+	if rcf.rc == nil {
+		return nil, errors.New("404, ReadCloser not found")
```
sonmezonur, line 251:
```suggestion
		return nil, os.ErrNotExist
```

storage/storage.go
```diff
@@ -227,3 +240,27 @@ func (m Metadata) SetSSEKeyID(kid string) Metadata {
...
+	}
+	return rcf.rc, nil
+}
+
+func (rcf *ReadCloserFile) File() (*os.File, error) {
+	if rcf.f == nil {
+		return nil, errors.New("404, File not found")
```
sonmezonur, line 258:
```suggestion
		return nil, os.ErrNotExist
```

command/mb.go
```diff
@@ -41,38 +41,49 @@ var makeBucketCommand = &cli.Command{
...
 
-// MakeBucket creates bucket.
-func MakeBucket(
-	ctx context.Context,
-	op string,
-	fullCommand string,
-	src string,
-) error {
-	bucket, err := url.New(src)
+type Bucket struct {
```
igungor, line 54: Let's call this `MakeBucket`.

command/cp.go
fbarotov: hmm, ok, thanks.

storage/storage.go
```diff
@@ -51,18 +51,30 @@ type Storage interface {
 	// filtered.
 	ListBuckets(ctx context.Context, prefix string) ([]Bucket, error)
 
-	// MakeBucket creates given bucket.
-	MakeBucket(ctx context.Context, bucket string) error
+	// Make creates bucket for remote operations and dir/file for local.
+	Make(ctx context.Context, opts MakeOpts) (ReadCloserFile, error)
+
+	// Open opens the given source. Return value can be either a readable and/or writable.
+	Open(ctx context.Context, src *url.URL) (ReadCloserFile, error)
```
sonmezonur, line 58: While `fs.Open()` actually opens a file, `s3.Open()` retrieves S3 object and reads its content. I suspect sharing a common interface for these functions might be confusing. Instead, maybe we can keep separate two methods for fs and s3. What do you think?

fbarotov: Yeah, I also think `fs.Open()` and `s3.Open()` do different things, and there is similar problem with 
`fs.Make()` and `s3.Make()`. First I thought of introducing new methods as you have suggested, but in that case we would have an interface with 11 methods, where `s3` does not implement(returns `notimplemented` error) 2 methods while `fs` not implements 4 methods. This reinforces the issue of making one interface implementation not be totally replaceable with the other.

One option might be to change storage interface as to contain only the methods common for both `s3` / `fs` and make clients work directly with `s3` or `fs` objects. This removes client side interface-abstraction, but it also removes surprises. Currently we are making an implicit assumption that for `storage.Storage{ local filesystem }`, users of interface will not call `storage.ListBuckets`. And in the future if `s5cmd` supports more s3 operations, dissimilarity of `s3` and `fs` will increase more.

So, shall we remove/modify `Open`, `Make` (also `ReadCloserFile`) and introduce new functions?

sonmezonur: let's discuss this later :) We can keep `Open()` and `Make()` for now.

igungor:
Closing this. #216 is better.

PR #220, opened 2020-08-25
command/cp: fix mime type inference

### LINKED ISSUE #214: cp command incorrectly infers MIME type
When using s5cmd and the `cp` command on a set of statically-generated web files to upload to an S3 bucket, the MIME types are not being correctly inferred. This causes some of them (ex. CSS files) to be fetched with the wrong response headers being returned. When using awscli and the `sync` command, for example, this is not an issue and the files are uploaded correctly.

Produces error:
`s5cmd cp -s -n -u <local_dir> s3://bucket_name`

Success:
`aws s3 sync --size-only --delete <local_dir> s3://bucket_name`

fbarotov:
Resolves #214


command/cp_test.go
```diff
@@ -0,0 +1,62 @@
...
+					<html>
+						<head>
+							<title>Hello World</title>
+						</head>
+						<body>
+							<p>Hello, World! I am s5cmd :)</p>
+						</body>
+					</html>
+					`,
+			expectedContentType: "text/plain; charset=utf-8",
```
igungor, line 41: I'm surprised to see `text/plain` here instead of `text/html`. Please take a look at https://github.com/peak/s5cmd/blob/master/e2e/cp_test.go#L657.

ilkinulas (approved):

command/cp_test.go
fbarotov: thanks, yes it should be `text/html` rather than `text/plain`.

igungor (approved):

PR #223, opened 2020-08-26
dockerfile: add dockerfile

### LINKED ISSUE #219: Provide Dockerfile

fbarotov:
Resolves #219

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM golang:1.14-alpine as build
 COPY . /s5cmd/
 RUN cd /s5cmd/ && \
-    go build -o s5cmd main.go
+    go build -mod=vendor -o s5cmd main.go
 
 FROM alpine:3.12
 COPY --from=build /s5cmd/s5cmd .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,8 @@
-FROM golang:1.14-alpine
-WORKDIR $GOPATH/src/s5cmd/
-COPY . .
-RUN go build -o s5cmd main.go
-CMD ["./s5cmd"]
+FROM golang:1.14-alpine as build
+COPY . /s5cmd/
+RUN cd /s5cmd/ && \
+    go build -o s5cmd main.go
+
+FROM alpine:3.12
+COPY --from=build /s5cmd/s5cmd .
+ENTRYPOINT ["./s5cmd"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM golang:1.14-alpine
-
-RUN \
-    apk add --no-cache git && \
-    git clone https://github.com/peak/s5cmd && cd s5cmd && \
-    go get ./... && \
-    go build -o s5cmd main.go
+WORKDIR $GOPATH/src/s5cmd/
+COPY . .
+RUN go build -o s5cmd main.go
+CMD ["./s5cmd"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 FROM golang:1.14-alpine as build
 COPY . /s5cmd/
-RUN cd /s5cmd/ && \
-    go build -mod=vendor -o s5cmd main.go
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang:1.14-alpine
+
+RUN \
```
aykutfarsak, line 3: We can separate env setup, git clone, and go build operations with multiple `RUN`.

Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+    apk add --no-cache git && \
+    git clone https://github.com/peak/s5cmd && cd s5cmd && \
+    go get ./... && \
+    go build -o s5cmd main.go
```
aykutfarsak, line 7: I think we should add `CMD` to use `s5cmd` as an entry point.

aykutfarsak (changes requested):

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang:1.14-alpine
+
+RUN \
+    apk add --no-cache git && \
+    git clone https://github.com/peak/s5cmd && cd s5cmd && \
```
seruman, line 5: Instead of cloning master, I think it would be better to add sources into container then build. This way one can build Docker images from revisions other than master without changing anything. To exclude sources in the final image [multistage build](https://docs.docker.com/develop/develop-images/multistage-build/) could be used.

seruman (changes requested):

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.14-alpine
```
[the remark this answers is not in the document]
fbarotov: thanks for the suggestion and help :)

seruman (approved):

ilkinulas:
@fbarotov could you please add a new section to README which shows how to build and run s5cmd docker image.
- docker build ...
- docker run ( `--env-file` flag from credentials)

We will update it after we publish the image to a public repo.

fbarotov:
> @fbarotov could you please add a new section to README which shows how to build and run s5cmd docker image.
> 
> * docker build ...
> * docker run ( `--env-file` flag from credentials)
> 
> We will update it after we publish the image to a public repo.

thanks for the help and suggestion. done.

ilkinulas (approved):

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:1.14-alpine as build
+COPY . /s5cmd/
+RUN cd /s5cmd/ && \
+    go build -o s5cmd main.go
```
igungor, line 4: Please use `-mod=vendor` flag.

fbarotov: thanks, done.

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golang:1.14-alpine as build
+COPY . /s5cmd/
+RUN cd /s5cmd/ && \
+    go build -mod=vendor -o s5cmd main.go
```
igungor, line 4: hmm, binary is missing the correct version information. Let's fix it in this PR.

- We have a `build` target in Makefile. It's better to use it instead of plain `go build`. 
- That target doesn't embed the correct tag and git commit information.

```
VERSION := `git describe --abbrev=0 --tags || echo "0.0.0"`
BUILD := `git rev-parse --short HEAD`
LDFLAGS=-ldflags "-X=github.com/peak/s5cmd/version.Version=$(VERSION) -X=github.com/peak/s5cmd/version.GitCommit=$(BUILD)"

.PHONY: build
build:
	@go build ${GCFLAGS} ${LDFLAGS} -mod=vendor .
```
should do the job.

fbarotov: thanks, done.

igungor: 🚀

igungor (approved):

igungor:
@aykutfarsak could you re-review please?

aykutfarsak (approved):
LGTM 👍

PR #224, opened 2020-08-27
command/cp: add stdin pipe option for cp

### LINKED ISSUE #182: Add option to stdin to cp
Basically the other end of https://github.com/peak/s5cmd/issues/20, which resulted in `s5cmd cat`. It doesn't look like using s5cmd run or cp has the magic "-" option to take in a piped input from stdin.

Use cases: Pipe a tar archive (or any streamed command output) into s5cmd without needing a intermediary temp file. This also solves cases where we want to preserve timestamps, permissions, etc.

Example:
`tar -cf - foo | s5cmd cp - s3://bar/foo.tar`

or if using `s5cmd run`

`tar -cf - foo | echo "cp - s3://bar/foo.tar" | s5cmd run`

fbarotov:
Resolves #182


igungor:
`s5cmd v1` is considered stable. Meaning we won't break current usage as far as possible. Removing the `cat` command is gonna break that promise 🙀

Let's discuss the issue first and implement second.

igungor (changes requested):
Blocking the merge until we have a consensus.

sonmezonur:
Removing `cat` command should not be an option because of backward compatibility. But supporting this feature would be a great improvement. Maybe we can add a new command (`pipe` or similar).

igungor:
I'm closing the PR. Once we have an agreement about the implementation, we can reevaluate this.

PR #227, opened 2020-08-28
vendor/: upgrade aws-sdk-go

### LINKED ISSUE #208: Retry on closed connection issue is handled by SDK
#189 is also fixed by [AWS SDK](https://github.com/aws/aws-sdk-go/pull/3476/files). When we upgrade the SDK version, remove the retry logic on our side but keep the relevant test.

fbarotov:
Resolves #208


storage/s3.go
```diff
@@ -95,8 +95,25 @@ func newS3Storage(opts Options, sessProvider func() *session.Session) (*S3, erro
...
 	awsSession := sessProvider()
 
+	// for copy operation aws-sdk-go expects a valid
+	// deserializable response and thus tries to unmarshal request.HTTPResponse.Body
+	// into an xml and this stage causes
+	// 'SerializationError: empty response payload status code: 200'
+	// if the response body is empty.
+	// For our tests, we need the following handler.
+	s3api := s3.New(awsSession)
+	s3api.Handlers.Unmarshal.PushBack(func(r *request.Request) {
```
igungor, line 105: I know that this is required for tests but not 100% sold if this is the only way, honestly.

igungor (changes requested):

storage/s3.go
fbarotov: thanks. yes, I also think there should be a better way, let me explore other possibilities.

fbarotov: The only clean way of solving this issue seems to make changes to gofakes3 that we use for tests, so it can return same response message as aws s3 for copy operations. I can fork and work on [igungor/gofakes3](https://github.com/igungor/gofakes3). Shall we proceed as such? @igungor

igungor: That'd be great!

igungor: @fbarotov is your latest commit final? should i review?

fbarotov: @igungor yes it is final, but there are some manual changes to `vendor/gofakes3`. Probably this fix should come after a new tag release from `igungor/gofakes3` following [this](https://github.com/igungor/gofakes3/pull/1) pull request.

igungor: I somehow missed that pull request. It is merged now and released with tag `v0.0.5`. Could you update go.mod and vendor accordingly?

fbarotov: done :)

storage/s3_test.go
```diff
@@ -523,6 +523,15 @@ func TestS3CopyEncryptionRequest(t *testing.T) {
 				}
 				assert.Equal(t, aclVal, tc.expectedAcl)
 			})
+			mockApi.Handlers.Unmarshal.PushBack(func(r *request.Request) {
```
igungor, line 526: is this still required, even after the gofakes3 change?

fbarotov: yess, for this test we do not interact with gofakes3 but return an empty response with [this](https://github.com/peak/s5cmd/pull/227/files#diff-2ba9e7f40440a8e598a45a16616a9ec760279956b775da2d7dc7b9f393e2e354L501) handler.

igungor: ah ok.

igungor (approved):

sonmezonur (approved):

PR #241, opened 2020-10-25
command/rm: batch rm with different buckets not allowed

fbarotov:


e2e/rm_test.go
```diff
@@ -501,6 +503,37 @@ func TestVariadicRemoveS3Objects(t *testing.T) {
...
+
+	cmd := s5cmd(
+		"rm",
+		"s3://"+bucket+"/file1",
+		"s3://"+someOtherBucket+"/file2",
+	)
+	result := icmd.RunCmd(cmd)
+
+	assertLines(t, result.Stderr(), map[int]compareFunc{
+		0: contains(command.ErrDifferentBucket.Error()),
```
igungor, line 530: let's check the string literal and remove the `command` import.

command/rm.go
```diff
@@ -142,6 +150,22 @@ func newURLs(sources ...string) ([]*url.URL, error) {
 	return urls, nil
 }
 
+// srcURLsBucketVal checks whether given urls all correspond to the same
+// bucket or not.
+func srcURLsBucketVal(urls []*url.URL) error {
```
igungor, line 155: `hasSameBuckets` is more readable. Also better return bool, and do error stuff at the callsite.

command/rm.go
```diff
@@ -80,7 +78,8 @@ func (d Delete) Run(ctx context.Context) error {
 		printError(d.fullCommand, d.op, err)
 		return err
 	}
-	if err := srcURLsBucketVal(srcurls); err != nil {
+	if !hasSameBuckets(srcurls) {
+		err := errors.New("more than one bucket detected")
```
igungor, line 82: could we print more user-friendly message? If I try to remove objects from different buckets (using variadic arguments of course), I'll get this message, which looks like a warning instead of a message that explains what is wrong with the command.

command/rm.go
```diff
@@ -77,6 +78,12 @@ func (d Delete) Run(ctx context.Context) error {
 		printError(d.fullCommand, d.op, err)
 		return err
 	}
+	if !hasSameBuckets(srcurls) {
+		err := errors.New("one rm command cannot be used for object removal of more than one bucket")
```
sonmezonur, line 82:
```suggestion
		err := fmt.Errorf("one rm command cannot be used for object removal of more than one bucket")
```

We did not use `errors` package before. For consistency purposes, let's use `fmt.Errorf`.

command/rm.go
```diff
@@ -77,6 +78,12 @@ func (d Delete) Run(ctx context.Context) error {
 		printError(d.fullCommand, d.op, err)
 		return err
 	}
+	if !hasSameBuckets(srcurls) {
```
sonmezonur, line 81: I think we should not do this validation in  `command.Run()`. Like other validations, it should be a part of `command.Before()` (validateRMCommand).

command/rm.go
```diff
@@ -174,5 +190,17 @@ func validateRMCommand(c *cli.Context) error {
 		return fmt.Errorf("expected at least 1 object to remove")
 	}
 
-	return sourcesHaveSameType(c.Args().Slice()...)
+	src := c.Args().Slice()
+	if err := sourcesHaveSameType(src...); err != nil {
```
sonmezonur, line 194: Let's change the signature of this function and make it the same with `hasSameBuckets`
```
func hasSameTypes(urls []*url.URL) error 
```

No need to create storage.URL twice ([ref](https://github.com/peak/s5cmd/blob/333370fa083048a424066f1acffc8f01a409e484/command/rm.go#L165))

fbarotov: with `func hasSameTypes(urls []*url.URL) error` it would be returning error whereas `hasSameBuckets` function returns a boolean (we might have slight inconsistency in naming pattern). maybe we can make it `func ensureSameTypes(urls []*url.URL) error`, is that ok?

sonmezonur: Sorry. Return value should be boolean rather than error. How about:

```
func hasSameTypes(urls []*url.URL) bool
```

fbarotov: hmm, since `sourcesHaveSameType` returns more than one type of error messages, I am not sure how we can take those outside the function.

sonmezonur: Hmm, I thought it only validates source types (by considering the method name).

I think we should refactor and improve this validation logic:

1) In the current design, we need to iterate over all source URLs. I think it is better to validate URLs in one phase.

2) Method name of sourcesHaveSameType is wrong. It also validates whether the source URL is prefix or bucket.  

3) [Validation error here](https://github.com/peak/s5cmd/blob/333370fa083048a424066f1acffc8f01a409e484/command/rm.go#L172) is misleading. It raises an error if sources are bucket or prefix. However, the error message is "source argument must contain wildcard character", which is actually wrong.

sonmezonur: Does it make sense to you?

fbarotov: Oh yeah 💯, that makes sense. thanks a lot.

e2e/rm_test.go
```diff
@@ -271,37 +271,6 @@ func TestRemoveS3PrefixWithoutSlash(t *testing.T) {
 	})
 }
 
-// rm s3://bucket/prefix/
-func TestRemoveS3Prefix(t *testing.T) {
```
sonmezonur, line 275: Why this test is removed? Cant we just update the error message rather than removing it?

command/rm.go
```diff
@@ -14,6 +14,12 @@ import (
 	"github.com/peak/s5cmd/storage/url"
 )
 
+const (
```
sonmezonur, line 17: I know this is intended for reusing error messages in tests. However, I would prefer to rewrite them in tests rather than declaring global variables.

command/rm.go
sonmezonur: Thanks for your effort. This implementation is much better 👍

e2e/rm_test.go
fbarotov: since we have refactored to make all `rm` validations inside one function and `TestRemoveS3Prefix` test content is already [in the](https://github.com/peak/s5cmd/pull/241/files#diff-fdb5fae95cc63262dac0eb174dc8eb812e177a9c10a18d4c60ebe0a96e7e30e9R35) `TestValidateRMCommand`, I thought removing one might be better.

command/rm.go
fbarotov: hmm, thanks :)

command/rm.go
fbarotov: thank you for your time and help :)

e2e/rm_test.go
sonmezonur: 👍

command/rm.go
sonmezonur: I don't have any special reasoning for this :)  I might be wrong but we should be consistent while handling errors. Since other commands don't have global error messages,  I thought these error messages should not be global as well. Either way is okay for me.

sonmezonur (approved):

command/rm.go
```diff
@@ -165,14 +171,14 @@ func sourcesHaveSameType(sources ...string) error {
...
-
-func validateRMCommand(c *cli.Context) error {
-	if !c.Args().Present() {
-		return fmt.Errorf("expected at least 1 object to remove")
+		if i == 0 {
+			firstBucket = srcurl.Bucket
+			continue
+		}
+		if srcurl.Bucket != firstBucket {
+			return fmt.Errorf("one rm command cannot be used for object removal of more than one bucket")
```
igungor, line 179: how about this? `removal of objects with different buckets in a single command is not allowed`

igungor (approved):

igungor:
Thanks!

PR #256, opened 2020-12-01
Dockerfile: disable CGO

### LINKED ISSUE #250: statically link the s5cmd binary
We use your s5cmd utility along with the kubectl utility to perform backups/restores of object stores provisioned by Rook Ceph in our Kubernetes cluster. The object buckets are represented by Object Bucket Claims and Object Buckets. To implement this, we need to have a container with the `s5cmd` utility, the `kubectl` utility, and `bash` to script the various tasks. Presently, the s5cmd contains a couple of dynamic library references which make this a little more complex:
```
ldd s5cmd
    /lib/ld-musl-x86_64.so.1 (0x7f9f5df96000)
    libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x7f9f5df96000)
```
Would you consider statically linking the s5cmd executable so that it can be more easily copied to a new container and combined with other utilities as descrbed?

seruman:
Fixes #250

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM golang:1.14-alpine as build
 COPY . /s5cmd/
 RUN apk add --no-cache git make && \
     cd /s5cmd/ && \
-    make build
+    CGO_ENABLED=0 make build
 
 FROM alpine:3.12
 COPY --from=build /s5cmd/s5cmd .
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM golang:1.14-alpine as build
 COPY . /s5cmd/
```
igungor, line 2: should we bump go image to 1.15?

seruman: bumped

igungor (approved):

igungor (commented):
please update changelog.

seruman:
@igungor updated changelog

aykutfarsak (approved):

Issue #244, opened 2020-11-02
s5cmd slow startup


raz-adroll:
Hi,

I noticed that using s5cmd was slow for me, doing a simple `s5cmd ls` would take ~20 seconds to start.

I decided to look into and I found why s5cmd was so slow when it ran in my machine, but ran ok in an ec2 instance, well, turns out that by default has a retry of 10 requests. Doing `s5cmd -r 0 ls` would actually be fast.

After some digging, I found out that it would make 10 requests (throttling included) for getting the credentials. So, it would make 10 calls to `169.254.169.254/latest/api/token` and then fallback to `169.254.169.254/latest/meta-data/iam/security-credentials/`

Then I Google similar issues and I got this: https://github.com/aws/aws-sdk-go/issues/2972

Maybe it should have a less aggressive retry for getting tokens?

arl:
In certain use cases this makes `s5cmd` unusable.
Now that I've understood the problem and aliased `s5cmd` to `s5cmd -r 0` I can switch to `s5cmd` back from `aws s3`, again.

tomberek:
Yes, ran into this when using mock credential providers.

Another referencing issue about the endpoint: https://github.com/aws/aws-sdk-go/issues/2972

bartek-siudeja:
It seems I still need `-r 0` to speed up start on s5cmd `v1.4.0-d7a0dda`.

strophy:
This issue occurs when running s5cmd in a Docker container on an EC2 instance due to the hop limit of 1 on requesting the EC2Metadata. The correct way to resolve it is to set `http_put_response_hop_limit = 2` in the instance metadata. 

Reference: https://github.com/aws/aws-sdk-go/issues/2972#issuecomment-747781100

PR #261, opened 2020-12-18
storage/s3: propagate retryable errors from sdk to the user

sonmezonur:
I think it would be useful to inform the caller for retryable errors. If retryable errors are received during command execution, users should be able to get information about it.

### DIFF
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -744,8 +744,8 @@ func (c *customRetryer) ShouldRetry(req *request.Request) bool {
 
 	if shouldRetry && req.Error != nil {
 		err := fmt.Errorf("retryable error: %v", req.Error)
-		msg := log.ErrorMessage{Err: err.Error()}
-		log.Error(msg)
+		msg := log.DebugMessage{Err: err.Error()}
+		log.Debug(msg)
 	}
 
 	return shouldRetry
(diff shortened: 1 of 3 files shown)


sonmezonur:
Since s5cmd does not print any logs during the retry period, it misleads users (they don't know what actually happens during execution). For example, this issue is not due to the slow startup of s5cmd, it is because aws sdk getting errors and we don't send any information to the user. ( https://github.com/peak/s5cmd/issues/244)

storage/s3.go
```diff
@@ -737,11 +737,18 @@ func newCustomRetryer(maxRetries int) *customRetryer {
...
-		return true
+	shouldRetry := errHasCode(req.Error, "InternalError") || errHasCode(req.Error, "RequestTimeTooSkewed")
+	if !shouldRetry {
+		shouldRetry = c.DefaultRetryer.ShouldRetry(req)
 	}
 
-	return c.DefaultRetryer.ShouldRetry(req)
+	if shouldRetry && req.Error != nil {
+		err := fmt.Errorf("retryable error: %v", req.Error)
+		msg := log.ErrorMessage{Err: err.Error()}
```
igungor, line 747: We should definitely log the message but I'm not sure about the severity since we still retry.

sonmezonur: Maybe Debug? We don't actively use this severity level 

only here: https://github.com/peak/s5cmd/blob/358a6c4261ec19d4f7305da9129dd0d6c65441c9/command/cp.go#L423

igungor: debug is 👍

ilkinulas (approved):

storage/s3.go
sonmezonur: done

igungor (approved):

PR #270, opened 2021-01-20
command/cp: Allow overriding bucket region

### LINKED ISSUE #262: Region setting not working
I'm currently using `s5cmd` to backup files to CTClouds (It provides S3-compitable storage, running by China Telecom).
The `s5cmd` binary I'm using is compiled locally using `go get`.

I'm copying files to a bucket on region "sh2", and this information is specified in `~/.aws/config`. However, when uploading, s5cmd always tries to upload to `us-east-1`, thus causing a `AuthorizationHeaderMalformed` error. Setting env `AWS_REGION` or `AWS_DEFAULT_REGION` to "sh2" won't change anything.

I noticed that since 1.2.0, s5cmd includes a feature "automatically determines region information of destination buckets". I'm wondering if it has anything to do with this feature, or is there any way I can confirm that, since adding `--log debug` won't give me any additional information.

kemege:
Add two parameters to `cp` and set region in session
This can be one way to resolve #262


storage/s3.go
```diff
@@ -678,8 +678,13 @@ func (sc *SessionCache) newSession(ctx context.Context, opts Options) (*session.
 	// get region of the bucket and create session accordingly. if the region
 	// is not provided, it means we want region-independent session
 	// for operations such as listing buckets, making a new bucket etc.
-	if err := setSessionRegion(ctx, sess, opts.bucket); err != nil {
-		return nil, err
+	// Only get region when it is specified
```
aykutfarsak, line 681:
```suggestion
	// only get bucket region when it is not specified.
```

command/cp.go
```diff
@@ -128,6 +128,14 @@ var copyCommandFlags = []cli.Flag{
 		Name:  "acl",
 		Usage: "set acl for target: defines granted accesses and their types on different accounts/groups",
 	},
+	&cli.StringFlag{
+		Name:  "src-region",
+		Usage: "set the region of source bucket: only useful when the correct region info cannot be fetched",
```
sonmezonur, line 133: `only useful when the correct region info cannot be fetched` might be misleading. some users may prefer to set `--source-region` (even auto-fetch logic works) since it sends additional API request. How about something like:
```
set the region of source bucket; the region of the source bucket will be automatically discovered if --source-region is not specified
```

command/cp.go
```diff
@@ -128,6 +128,14 @@ var copyCommandFlags = []cli.Flag{
...
+	},
+	&cli.StringFlag{
+		Name:  "dst-region",
+		Usage: "set the region of destination bucket: only useful when the correct region info cannot be fetched",
```
sonmezonur, line 137: same as above `--source-region` comment

sonmezonur (commented):
Thanks for the contribution and sorry for the late feedback. LGTM aside from some minor comments

command/cp.go
rileytg: @sonmezonur does that mean after this pull is merged, if i explicitly set `--source-region` i can save a roundtrip request on each file? 

(for context: I'm doing millions of small files using `run` with `cp`)

sonmezonur: > @sonmezonur does that mean after this pull is merged, if i explicitly set `--source-region` i can save a roundtrip request on each file?
> 
> (for context: I'm doing millions of small files using `run` with `cp`)


Actually, it does not. `s5cmd` retrieves bucket location during session creation and re-uses these sessions (from global cache) afterward. For each bucket, we are sending [HeadBucket](https://docs.aws.amazon.com/en_us/AmazonS3/latest/API/API_HeadBucket.html) request to determine its region. So if you are copying files to `same bucket`, you can save only one roundtrip request. But if you have multiple destination buckets in your run file, you may expect more.

kemege: @sonmezonur I agree with you on that. Your statement looks much better.

sonmezonur (approved):

aykutfarsak (approved):

storage/storage.go
```diff
@@ -67,12 +68,18 @@ func NewClient(ctx context.Context, url *url.URL, opts Options) (Storage, error)
...
+	NoVerifySSL bool
+	DryRun      bool
 	NoSignRequest bool
-	bucket        string
+	bucket      string
+	region      string
+}
+
+func (o Options) SetRegion(region string) Options {
+	o.region = region
```
igungor, line 81: SetRegion won't work because receiver is not a pointer. https://play.golang.org/p/y_6xsO3zB9L

storage/s3.go
```diff
@@ -684,8 +684,13 @@ func (sc *SessionCache) newSession(ctx context.Context, opts Options) (*session.
 	// get region of the bucket and create session accordingly. if the region
 	// is not provided, it means we want region-independent session
 	// for operations such as listing buckets, making a new bucket etc.
-	if err := setSessionRegion(ctx, sess, opts.bucket); err != nil {
-		return nil, err
+	// only get bucket region when it is not specified.
+	if len(opts.region) > 0 {
```
igungor, line 688:
```suggestion
	if opts.region != "" {
```

command/cp.go
```diff
@@ -473,6 +496,10 @@ func (c Copy) doUpload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) er
 }
 
 func (c Copy) doCopy(ctx context.Context, srcurl, dsturl *url.URL) error {
+	// override destination region if set
+	if len(c.dstRegion) > 0 {
```
igungor, line 500:
```suggestion
	if c.dstRegion != "" {
```

command/cp.go
```diff
@@ -430,6 +449,10 @@ func (c Copy) doUpload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) er
 		return err
 	}
 
+	// override destination region if set
+	if len(c.dstRegion) > 0 {
```
igungor, line 453:
```suggestion
	if c.dstRegion != "" {
```

command/cp.go
```diff
@@ -219,6 +234,10 @@ func (c Copy) Run(ctx context.Context) error {
 		return err
 	}
 
+	// override source region if set
+	if len(c.srcRegion) > 0 {
```
igungor, line 238:
```suggestion
	if c.srcRegion) != "" {
```

command/cp.go
```diff
@@ -131,6 +131,14 @@ var copyCommandFlags = []cli.Flag{
 		Name:  "acl",
 		Usage: "set acl for target: defines granted accesses and their types on different accounts/groups",
 	},
+	&cli.StringFlag{
+		Name:  "source-region",
+		Usage: "set the region of source bucket; the region of the source bucket will be automatically discovered if --source-region is not specified",
+	},
+	&cli.StringFlag{
+		Name:  "destination-region",
+		Usage: "set the region of destination bucket: the region of the destination bucket will be automatically discovered if --source-region is not specified",
```
igungor, line 140:
```suggestion
		Usage: "set the region of destination bucket: the region of the destination bucket will be automatically discovered if --destination-region is not specified",
```

igungor (changes requested):

storage/storage.go
kemege: Actually it works here since I'm using `opt = opt.SetRegion(...)`. Anyway, I will change this to make it consistent with other setters.

kemege (commented):
I've changed the SetRegion function to use pointers, along with the typo in the description for `destination-region`

igungor (approved):

igungor:
> I've changed the SetRegion function to use pointers, along with the typo in the description for `destination-region`

Thank you.

Issue #279, opened 2021-04-01
File not found if S3 key contains special characters


khacminh:
### Issue:

S3 key may contain special characters as mention in this [document](https://docs.aws.amazon.com/AmazonS3/latest/dev-retired/UsingMetadata.html). In this case, downloading S3 object throws 404 error.

### Solution:

The S3 key need to be URL encoded before download.

ilkinulas:
Hi @khacminh, thank you for reporting this.

Could you please give an example (and valid) s3 prefix where download s3 object request throws 404 error?

khacminh:
Hi @ilkinulas 

The source bucket has the following objects:

```txt
test-dir/object-2021-04-04+1.txt
test-dir/object-2021-04-04:2.txt
```

Then I would like to copy them to the new bucket with the following command

```sh
s5cmd cp srcBucket/* dstBucket/
```

---

According to AWS [document](https://docs.aws.amazon.com/AmazonS3/latest/dev-retired/UsingMetadata.html), the following character must be handled by URL enccoding or referenced as HEX:

- Ampersand ("&")
- Dollar ("$")
- 'At' symbol ("@")
- Equals ("=")
- Semicolon (";")
- Colon (":")
- Plus ("+")
- Space
- Comma (",")
- Question mark ("?")

Encoding the url component fix this issue. However, it breaks another test cases using wildcard.

PR #280, opened 2021-04-01
Implement #279

khacminh:
Implement Issue #279 

### Issue:
S3 key may contain special characters as mention in this document. In this case, downloading S3 object throws 404 error.

### Solution:
The S3 key need to be URL encoded before download.


storage/url/url.go
```diff
@@ -178,7 +179,11 @@ func (u *URL) remoteURL() string {
 	}
 
 	if u.Path != "" {
-		s += "/" + u.Path
+		pathElements := regexp.MustCompile(`/`).Split(u.Path, -1)
```
ilkinulas, line 182: S3 is returning `404 Not Found` because s5cmd is not escaping the s3 key. 

Could you please add escape logic to `s3.go` where we are creating `s3.CopyObjectInput` ?
https://github.com/peak/s5cmd/blob/6b97a9eeb01c4d7a2658613aead539702197bbd4/storage/s3.go#L319

Updaing `url.go` may break wildcard support.

ilkinulas (changes requested):

storage/url/url.go
khacminh: Hi @ilkinulas
I updated as your suggestion.

---

## About the test

I tested with the real s3 buckets, and it works.

I tried a new unit test. However, it doesn't work as expected. 

- If I `putFile()` an object with key: `a+1.txt`, running the cmd `cp s3://srcBucket/* s3://dstBucket/` will throw an 404 error because the key `a%2B1.txt` doesn't exist.
- The same error with key `a%2B1.txt` is sent to `putFile()`. Because the key is not converted to `a+1.txt` as the real S3 behavior. So, executing the `cp` command, the object key will be escaped one more time to `a%252B1.txt`

I think that's why the test `TestRunSpecialCharactersInPrefix()` was passed.

storage/s3.go
```diff
@@ -347,6 +348,14 @@ func (s *S3) Copy(ctx context.Context, from, to *url.URL, metadata Metadata) err
 	return err
 }
 
+func escapeKey(sourceKey string) string {
+	sourceKeyElements := regexp.MustCompile(`/`).Split(sourceKey, -1)
```
ilkinulas, line 352: why not `strings.Split(sourceKey, "/")` ?

and could you please add a unit test for this function.

storage/url/url.go
ilkinulas: I have also tried to write a test that uses `putFile(...)` and test passed. Could you please share your test where copy command throws an 404 error?

storage/s3.go
khacminh: I intended for other special-characters but it is not necessary because they could be either in:
- safe group: no need to handle, except `/`
- should avoid group.

I will update it as your suggestion

storage/url/url.go
khacminh: @ilkinulas 

I pushed my test. Here are the error:

---
## Case 1:

Screen Shot 2021-04-08 at 23 38 17

the following errors occurs:

```txt
Stderr:   ERROR "cp s3://bucket/sub&@$/test+1.txt s3://dstbucket/sub&@$/test+1.txt": NoSuchKey:  status code: 404, request id: 000000000000000A, host id: MDAwMDAwMDAwMDAwMDAwQTAwMDAwMDAwMDAwMDAwMEEwMDAwMDAwMDAwMDAwMDBBMDAwMDAwMDAwMDAwMDAwQQ==
        ERROR "cp s3://bucket/test&@$:,?;= 3.txt s3://dstbucket/test&@$:,?;= 3.txt": NoSuchKey:  status code: 404, request id: 000000000000000B, host id: MDAwMDAwMDAwMDAwMDAwQjAwMDAwMDAwMDAwMDAwMEIwMDAwMDAwMDAwMDAwMDBCMDAwMDAwMDAwMDAwMDAwQg==
        ERROR "cp s3://bucket/sub:,?/test; =2.txt s3://dstbucket/sub:,?/test; =2.txt": NoSuchKey:  status code: 404, request id: 000000000000000D, host id: MDAwMDAwMDAwMDAwMDAwRDAwMDAwMDAwMDAwMDAwMEQwMDAwMDAwMDAwMDAwMDBEMDAwMDAwMDAwMDAwMDAwRA==
```

---

## Case 2:

Screen Shot 2021-04-08 at 23 40 05

the following errors occurs:

```txt
Stderr:   ERROR "cp s3://bucket/sub%3A%2C%3F/test%3B+%3D2.txt s3://dstbucket/sub%3A%2C%3F/test%3B+%3D2.txt": NoSuchKey:  status code: 404, request id: 000000000000000A, host id: MDAwMDAwMDAwMDAwMDAwQTAwMDAwMDAwMDAwMDAwMEEwMDAwMDAwMDAwMDAwMDBBMDAwMDAwMDAwMDAwMDAwQQ==
        ERROR "cp s3://bucket/test%26%40%24%3A%2C%3F%3B%3D+3.txt s3://dstbucket/test%26%40%24%3A%2C%3F%3B%3D+3.txt": NoSuchKey:  status code: 404, request id: 000000000000000B, host id: MDAwMDAwMDAwMDAwMDAwQjAwMDAwMDAwMDAwMDAwMEIwMDAwMDAwMDAwMDAwMDBCMDAwMDAwMDAwMDAwMDAwQg==
        ERROR "cp s3://bucket/sub%26%40%24/test%2B1.txt s3://dstbucket/sub%26%40%24/test%2B1.txt": NoSuchKey:  status code: 404, request id: 000000000000000C, host id: MDAwMDAwMDAwMDAwMDAwQzAwMDAwMDAwMDAwMDAwMEMwMDAwMDAwMDAwMDAwMDBDMDAwMDAwMDAwMDAwMDAwQw==
```

e2e/cp_test.go
```diff
@@ -1646,6 +1646,52 @@ func TestCopySingleS3ObjectIntoAnotherBucketWithPrefix(t *testing.T) {
...
+
+	src := fmt.Sprintf("s3://%v/*", srcbucket)
+	dst := fmt.Sprintf("s3://%v/", dstbucket)
+
+	cmd := s5cmd("cp", src, dst)
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+
+	assertLines(t, result.Stdout(), map[int]compareFunc{
```
ilkinulas, line 1683: This assertion is not necessary. assertion on line 1687 is enough to check the output of the cp commands.

e2e/cp_test.go
```diff
@@ -1646,6 +1646,52 @@ func TestCopySingleS3ObjectIntoAnotherBucketWithPrefix(t *testing.T) {
...
+		0: equals(`cp %v %v`, src, dst),
+	})
+
+	assertLines(t, result.Stdout(), map[int]compareFunc{
+		0: equals(`cp s3://%v/sub&@$/test+1.txt s3://%v/sub&@$/test+1.txt`, srcbucket, dst),
```
ilkinulas, line 1688:
```suggestion
		0: equals(`cp s3://%v/sub&@$/test+1.txt s3://%v/sub&@$/test+1.txt`, srcbucket, dstbucket),
```

storage/url/url.go
ilkinulas: I am still trying to find out why test is failing with 404, fyi

e2e/cp_test.go
khacminh: I removed these assertions as your suggestion.

storage/s3.go
```diff
@@ -347,6 +347,14 @@ func (s *S3) Copy(ctx context.Context, from, to *url.URL, metadata Metadata) err
 	return err
 }
 
+func escapeKey(sourceKey string) string {
```
igungor, line 350: Could you move this function into our `url` package?

e2e/cp_test.go
```diff
@@ -1646,6 +1646,46 @@ func TestCopySingleS3ObjectIntoAnotherBucketWithPrefix(t *testing.T) {
...
+	for filename, content := range filesToContent {
+		putFile(t, s3client, srcbucket, filename, content)
+	}
+
+	src := fmt.Sprintf("s3://%v/*", srcbucket)
+	dst := fmt.Sprintf("s3://%v/", dstbucket)
+
+	cmd := s5cmd("cp", src, dst)
+	result := icmd.RunCmd(cmd)
+
```
ilkinulas, line 1680: Could you please add an assertion for run command result ?
```suggestion
result.Assert(t, icmd.Success)
```

go.mod
```diff
@@ -22,4 +22,4 @@ require (
 	gotest.tools/v3 v3.0.2
 )
 
-replace github.com/johannesboyne/gofakes3 => github.com/igungor/gofakes3 v0.0.5
\ No newline at end of file
+replace github.com/johannesboyne/gofakes3 => github.com/igungor/gofakes3 v0.0.5
```
ilkinulas, line 25: We can now switch to https://github.com/johannesboyne/gofakes3 (upstream). With the latest gofakes3 tests will pass. 

Could you please remove the `replace` directive  and set gofakes3 version to 4b6b67a85d38ec538153d0f21b36378a38edb143(latest commit)?

storage/s3.go
khacminh: Moved to `url.go`

https://github.com/peak/s5cmd/blob/cd5170dd8c7abad0d75de47335a7f1def3711fc1/storage/url/url.go#L352-L359

e2e/cp_test.go
khacminh: I added the assertion

go.mod
khacminh: I switched to the latest commit, then run `go mod vendor`. Running on my computer (macOS 11.2.3), the test is passed but failed on ci/test

Screen Shot 2021-04-15 at 23 39 26

storage/s3.go
igungor: 👍

go.mod
ilkinulas: This is also unexpected for us. We are trying to figure out the root cause of these flaky tests. I will update the issue as soon as we find something.

ilkinulas: Hi @khacminh 

Could you please update gofakes3 dependency to https://github.com/igungor/gofakes3/releases/tag/v0.0.6  by re-adding the replace directive?

```
replace github.com/johannesboyne/gofakes3 => github.com/igungor/gofakes3 v0.0.6
```

We decided to continue to use our fork, because with gofakes3 upstream, tests are failing randomly while running in parallel.

storage/url/url.go
```diff
@@ -347,3 +348,12 @@ func parseNonBatch(prefix string, key string) string {
...
 	return strings.ContainsAny(s, globCharacters)
 }
+
+func (u *URL) GetCopySource() string {
+	sourceKey := strings.TrimPrefix(u.String(), "s3://")
+	sourceKeyElements := strings.Split(sourceKey, "/")
+	for i, element := range sourceKeyElements {
+		sourceKeyElements[i] = url.QueryEscape(element)
+	}
+	return strings.Join(sourceKeyElements, `/`)
```
sonmezonur, line 358: let's use single-line string representation.

```suggestion
	return strings.Join(sourceKeyElements, "/")
```

storage/url/url.go
```diff
@@ -347,3 +348,12 @@ func parseNonBatch(prefix string, key string) string {
 func hasGlobCharacter(s string) bool {
 	return strings.ContainsAny(s, globCharacters)
 }
+
+func (u *URL) GetCopySource() string {
```
sonmezonur, line 352: Since `storage.URL` is a generic abstraction for all storage operations, I recommend using a more generic method name such as `EscapedPath()`, etc. 

```suggestion
func (u *URL) EscapedPath() string {
```

sonmezonur (commented):
Thanks for your contribution.

storage/url/url.go
khacminh: I updated as your suggestion

storage/url/url.go
khacminh: I updated as your suggestion

go.mod
khacminh: Yay! The tests are passed now

khacminh: hi @ilkinulas 
I also revert the go.mod file to minimize the file changes. Re-run the tests and they are passed.

sonmezonur (approved):

ilkinulas (approved):
Thank you for your contribution, it is good to go!

PR #298, opened 2021-06-14
Treat "connection reset" errors as retryable.

### LINKED ISSUE #294: Got "read: connection reset by peer" error when using version 1.0.0+
Tested on OpenStack Swift using s5cmd version 1.0.0 and 1.2.1 with the following command:

```
s5cmd --endpoint-url http://objstor.testserver.com --no-verify-ssl  cp -n -s -u  aaaatest2 s3://aaaatest01/aaaajfg006/
```

And got errors like this:

```
ERROR "cp aaaatest2/2ms/2m_file_1 s3://aaaatest01/aaaajfg006/aaaatest2/2ms/2m_file_1": RequestError: send request failed caused by: Head "http://objstor.testserver.com/aaaatest01/aaaajfg006/aaaatest2/2ms/2m_file_1": read tcp xxx.xxx.xxx.xxx:38924->xxx.xxx.xxx.xxx:80: read: connection reset by peer
```

However, it's all good using s5cmd v0.7.0 with the same command. No error returned at all.

I aslo found an issue on aws-sdk-go repo (see below) complainting about the same error.

https://github.com/aws/aws-sdk-go/issues/3027

Is it possible that it's the new version of aws-sdk-go causes the problem?

nelhage:
Per
https://github.com/aws/aws-sdk-go/pull/2926#issuecomment-553196888,
the upstream SDK doesn't retry connection reset errors on reads,
because at that point we can't tell if the server has already applied
the operation.

For S3 operations, I think pretty much all requests should be safely
idempotent, so I think we should retry these errors.

I saw some of my own long-running jobs fail this weekend due to
network weather and some dropped connections that s5cmd did not retry,
which I found very surprising.

Per
https://github.com/aws/aws-sdk-go/commit/c3d27102088451890a9ba5be0dc1f2e0425f0fc3,
the `strings.Contains` check does seem to be the best way to check for
this error.

This might fix #294, or there might be something deeper going on there.

### DIFF
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -737,7 +737,7 @@ func newCustomRetryer(maxRetries int) *customRetryer {
 // ShouldRetry overrides SDK's built in DefaultRetryer, adding custom retry
 // logics that are not included in the SDK.
 func (c *customRetryer) ShouldRetry(req *request.Request) bool {
-	shouldRetry := errHasCode(req.Error, "InternalError") || errHasCode(req.Error, "RequestTimeTooSkewed")
+	shouldRetry := errHasCode(req.Error, "InternalError") || errHasCode(req.Error, "RequestTimeTooSkewed") || strings.Contains(req.Error.Error(), "connection reset")
 	if !shouldRetry {
 		shouldRetry = c.DefaultRetryer.ShouldRetry(req)
 	}


igungor (approved):

igungor:
Thank you.

Issue #51, opened 2020-02-12
Add console progress bar for multipart upload/download


sonmezonur:
Currently, s5cmd does not show the progress of multipart operations. It would be nice to add an option to display the progress for large files.

geosigno:
Any update or workaround to get a progress event?

ochaton:
It would be nice if s5cmd would show progress bar for `cp` command

leng-yue:
Any update?

sorenwacker:
So far this feature is not available.

mwtoews:
It would be excellent if this option were also available for `sync` too.

PR #300, opened 2021-06-16
Implement s5cmd select

### LINKED ISSUE #299: SelectObjectContent support
I prototyped `SelectObjectContent` support (with wildcards) for `s5cmd`, and it works pretty well! I did make some assumptions, and it's not quite ready for primetime, and it'd take a bit of work to make it ready to go. I'm happy to open a PR if this is something you're open to supporting.

```sh
$ s5cmd select --query "SELECT * FROM S3Object s WHERE s.query_name LIKE '%.github.com.'" --compression-type gzip s3://example-bucket/AWSLogs/*
```

skeggse:
Fixes #299.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ storage services and local filesystems.
 - Set Server Side Encryption using AWS Key Management Service (KMS)
 - Set Access Control List (ACL) for objects/files on the upload, copy, move.
 - Print object contents to stdout
+- Select JSON records from objects using SQL expressions
 - Create or remove buckets
 - Summarize objects sizes, grouping by storage class
 - Wildcard support for all operations
@@ -190,7 +191,7 @@ are not supported by `s5cmd` and result in error (since we have 2 different buck
     rm s3://bucket-foo/object
     rm s3://bucket-bar/object
 
-more details and examples on `s5cmd run` are presented in a [later section](./README.md#L208).
+more details and examples on `s5cmd run` are presented in a [later section](./README.md#L224).
 
 #### Copy objects from S3 to S3
 
@@ -203,6 +204,24 @@ folder hierarchy.
 
 ⚠️ Copying objects (from S3 to S3) larger than 5GB is not supported yet. We have
 an [open ticket](https://github.com/peak/s5cmd/issues/29) to track the issue.
+
+#### Select JSON object content using SQL
+
+`s5cmd` supports the `SelectObjectContent` S3 operation, and will run your
+[SQL query](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-glacier-select-sql-reference.html)
+against objects matching normal wildcard syntax and emit matching JSON records via stdout. Records
+from multiple objects will be interleaved, and order of the records is not guaranteed (though it's
+likely that the records from a single object will arrive in-order, even if interleaved with other
+records).
+
+    $ s5cmd select --compression GZIP \
+      --query "SELECT timestamp, hostname FROM S3Object WHERE ip_address LIKE '10.%' OR application='unprivileged'" \
+      s3://bucket-foo/object/2021/*
+    {"timestamp":"2021-07-08T18:24:06.665Z","hostname":"application.internal"}
+    {"timestamp":"2021-07-08T18:24:16.095Z","hostname":"api.github.com"}
+
+At the moment this operation _only_ supports JSON records selected with SQL. S3 calls this
+lines-type JSON, but it seems that it works even if the records aren't line-delineated. YMMV.
 
 #### Count objects and determine total size
 
@@ -406,7 +425,7 @@ For a more practical scenario, let's say we have an [avocado prices](https://www
 
 ## Beast Mode s5cmd
 
-`s5cmd` allows to pass in some file, containing list of operations to be performed, as an argument to the `run` command as illustrated in the [above](./README.md#L199) example. Alternatively, one can pipe in commands into
+`s5cmd` allows to pass in some file, containing list of operations to be performed, as an argument to the `run` command as illustrated in the [above](./README.md#L224) example. Alternatively, one can pipe in commands into
 the `run:`
 
     BUCKET=s5cmd-test; s5cmd ls s3://$BUCKET/*test | grep -v DIR | awk ‘{print $NF}’
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ likely that the records from a single object will arrive in-order, even if inter
 records).
 
     $ s5cmd select --compression GZIP \
-      --query "SELECT timestamp, hostname FROM S3Object WHERE ip_address LIKE '10.%' OR application='unprivileged'" \
+      --query "SELECT s.timestamp, s.hostname FROM S3Object s WHERE s.ip_address LIKE '10.%' OR s.application='unprivileged'" \
       s3://bucket-foo/object/2021/*
     {"timestamp":"2021-07-08T18:24:06.665Z","hostname":"application.internal"}
     {"timestamp":"2021-07-08T18:24:16.095Z","hostname":"api.github.com"}
--- a/command/select.go
+++ b/command/select.go
@@ -1,0 +1,163 @@
+package command
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/urfave/cli/v2"
+
+	"github.com/peak/s5cmd/log/stat"
+	"github.com/peak/s5cmd/parallel"
+	"github.com/peak/s5cmd/storage"
+	"github.com/peak/s5cmd/storage/url"
+)
+
+var selectHelpTemplate = `Name:
+	{{.HelpName}} - {{.Usage}}
+
+Usage:
+	{{.HelpName}} [options] argument
+
+Options:
+	{{range .VisibleFlags}}{{.}}
+	{{end}}
+Examples:
+	01. Search for all JSON objects with the foo property set to 'bar' and spit them into stdout
+		 > s5cmd {{.HelpName}} --compression-type gzip --query "SELECT * FROM S3Object s WHERE s.foo='bar'" s3://bucket/*
+`
+
+var selectCommandFlags = []cli.Flag{
+	&cli.StringFlag{
+		Name:    "query",
+		Aliases: []string{"e"},
+		Usage:   "SQL expression to use to select from the objects",
+	},
+	&cli.StringFlag{
+		Name:  "compression-type",
+		Usage: "Type of compression used in storage",
+	},
+}
+
+var selectCommand = &cli.Command{
+	Name:               "select",
+	HelpName:           "select",
+	Usage:              "select objects containing JSON using a SQL query",
+	Flags:              selectCommandFlags,
+	CustomHelpTemplate: selectHelpTemplate,
+	Before: func(c *cli.Context) error {
+		err := validateSelectCommand(c)
+		if err != nil {
+			printError(givenCommand(c), c.Command.Name, err)
+		}
+		return err
+	},
+	Action: func(c *cli.Context) (err error) {
+		defer stat.Collect(c.Command.FullName(), &err)()
+
+		return Select{
+			src:         c.Args().Get(0),
+			op:          c.Command.Name,
+			fullCommand: givenCommand(c),
+			// flags
+			query:           c.String("query"),
+			compressionType: c.String("compression-type"),
+
+			storageOpts: NewStorageOpts(c),
+		}.Run(c.Context)
+	},
+}
+
+// Select holds select operation flags and states.
+type Select struct {
+	src         string
+	op          string
+	fullCommand string
+
+	deleteSource bool
+
+	query           string
+	compressionType string
+
+	// s3 options
+	storageOpts storage.Options
+}
+
+// Run starts copying given source objects to destination.
+func (s Select) Run(ctx context.Context) error {
+	srcurl, err := url.New(s.src)
+	if err != nil {
+		printError(s.fullCommand, s.op, err)
+		return err
+	}
+
+	client, err := storage.NewRemoteClient(ctx, srcurl, s.storageOpts)
+	if err != nil {
+		printError(s.fullCommand, s.op, err)
+		return err
+	}
+
+	// TODO: fix prefix support?
+	objch, err := expandSource(ctx, client, false, srcurl)
+	if err != nil {
+		printError(s.fullCommand, s.op, err)
+		return err
+	}
+
+	err = s.doSelect(ctx, objch)
+	if err != nil {
+		printError(s.fullCommand, s.op, err)
+	}
+	return err
+}
+
+// doSelect
+func (s Select) doSelect(ctx context.Context, objch <-chan *storage.Object) error {
+	// set as remote storage
+	url := &url.URL{Type: 0}
+	client, err := storage.NewRemoteClient(ctx, url, s.storageOpts)
+	if err != nil {
+		return err
+	}
+
+	query := storage.Query{
+		ExpressionType:  "SQL",
+		Expression:      s.query,
+		CompressionType: s.compressionType,
+	}
+	err = client.Select(ctx, objch, os.Stdout, &query, parallel.Size())
+	if err != nil {
+		return err
+	}
+
+	// msg := log.InfoMessage{
+	// 	Operation:   c.op,
+	// 	Source:      srcurl,
+	// 	Destination: dsturl,
+	// 	Object: &storage.Object{
+	// 		Size: size,
+	// 	},
+	// }
+	// log.Info(msg)
+
+	return nil
+}
+
+func validateSelectCommand(c *cli.Context) error {
+	if c.Args().Len() != 1 {
+		return fmt.Errorf("expected source argument")
+	}
+
+	src := c.Args().Get(0)
+
+	srcurl, err := url.New(src)
+	if err != nil {
+		return err
+	}
+
+	if !srcurl.IsRemote() {
+		return fmt.Errorf("source must be remote")
+	}
+
+	return nil
+}
--- a/command/select.go
+++ b/command/select.go
@@ -39,12 +39,12 @@ var selectCommandFlags = []cli.Flag{
 	},
 	&cli.StringFlag{
 		Name:  "compression",
-		Usage: "Type of compression used in storage",
+		Usage: "input compression format",
 		Value: "NONE",
 	},
 	&cli.StringFlag{
 		Name:  "format",
-		Usage: "Format of input data in storage",
+		Usage: "input data format",
 		Value: "JSON",
 	},
 }
@@ -52,7 +52,7 @@ var selectCommandFlags = []cli.Flag{
 var selectCommand = &cli.Command{
 	Name:               "select",
 	HelpName:           "select",
-	Usage:              "select objects containing JSON using a SQL query",
+	Usage:              "run SQL queries on objects",
 	Flags:              selectCommandFlags,
 	CustomHelpTemplate: selectHelpTemplate,
 	Before: func(c *cli.Context) error {
--- a/command/select.go
+++ b/command/select.go
@@ -172,6 +172,7 @@ func (s Select) Run(ctx context.Context) error {
 
 	waiter.Wait()
 	<-errDoneCh
+	<-writeDoneCh
 
 	return merror
 }
--- a/command/select.go
+++ b/command/select.go
@@ -84,8 +84,6 @@ type Select struct {
 	op          string
 	fullCommand string
 
-	deleteSource bool
-
 	query           string
 	compressionType string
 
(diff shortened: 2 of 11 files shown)


command/select.go
```diff
@@ -0,0 +1,163 @@
...
+
+Options:
+	{{range .VisibleFlags}}{{.}}
+	{{end}}
+Examples:
+	01. Search for all JSON objects with the foo property set to 'bar' and spit them into stdout
+		 > s5cmd {{.HelpName}} --compression-type gzip --query "SELECT * FROM S3Object s WHERE s.foo='bar'" s3://bucket/*
+`
+
+var selectCommandFlags = []cli.Flag{
```
skeggse, line 30: It might make sense to explicitly require the format here - I'm not sure I have the momentum to add full support for parquet and CSV formats, but I could at least add the flag? Let me know if CSV + parquet support seem mandatory for landing this.

command/select.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	client, err := storage.NewRemoteClient(ctx, srcurl, s.storageOpts)
+	if err != nil {
+		printError(s.fullCommand, s.op, err)
+		return err
+	}
+
+	// TODO: fix prefix support?
+	objch, err := expandSource(ctx, client, false, srcurl)
```
skeggse, line 101: In testing, I wasn't able to just provide a prefix. Maybe it's not statically possible to determine whether you need to recurse, or maybe I should just see if there's a trailing `'/'` and if so append a `'*'`?

command/select.go
```diff
@@ -0,0 +1,163 @@
...
+
+	err = s.doSelect(ctx, objch)
+	if err != nil {
+		printError(s.fullCommand, s.op, err)
+	}
+	return err
+}
+
+// doSelect
+func (s Select) doSelect(ctx context.Context, objch <-chan *storage.Object) error {
```
skeggse, line 115: I do need a bunch more doc comments, for sure.

command/select.go
```diff
@@ -0,0 +1,163 @@
...
+	// set as remote storage
+	url := &url.URL{Type: 0}
+	client, err := storage.NewRemoteClient(ctx, url, s.storageOpts)
+	if err != nil {
+		return err
+	}
+
+	query := storage.Query{
+		ExpressionType:  "SQL",
```
skeggse, line 124: This is currently the only supported expression type. Not clear to me that it's worth adding a flag for this.

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+		defer writer.Close()
+		// TODO: wire ctx up to eventStream.StreamCloser.Close
+		for event := range resp.EventStream.Reader.Events() {
+			switch e := event.(type) {
+			case *s3.RecordsEvent:
+				// TODO: how should we handle errors here?!
+				writer.Write(e.Payload)
+			// How the heck are these different?
+			case *s3.StatsEvent:
+			case *s3.ProgressEvent:
```
skeggse, line 409: I think it'd be cool to try and output progress to stderr, but I didn't get to that in this PR.

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+			}
+		}
+	}()
+
+	decoder := json.NewDecoder(reader)
+	// decoder.DontCopyRawMessage()
```
skeggse, line 415: I tried doing some zerocopy stuff, but it looks like there's an internal buffer that gets reused, and doing that plus the delay between writing to the channel and reading from the channel causes data corruption.

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+		resultch <- &SelectRecord{
+			Err: err,
+		}
+	}
+}
+
+func (s *S3) selectWorker(ctx context.Context, input s3.SelectObjectContentInput, objch <-chan *Object, resultch chan<- *SelectRecord) {
+	for object := range objch {
+		// TODO: reintroduce
+		if object.Type.IsDir() { // || errorpkg.IsCancelation(object.Err) {
```
skeggse, line 442: I can't `import "errorpkg"` because it causes an import loop. If you have a preferred way of resolving this, lmk.

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+	}
+}
+
+type Query struct {
+	ExpressionType  string
+	Expression      string
+	CompressionType string
+}
+
+func writev(to io.Writer, vector ...[]byte) (err error) {
```
skeggse, line 482: A little misleading, seeing as it doesn't actually use the `writev` syscall.

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+		// 	// TODO: conditionally enable? for larger objects? ?
+		// 	Enabled: false,
+		// },
+	}
+
+	workerCtx, cancel := context.WithCancel(ctx)
+
+	resultch := make(chan *SelectRecord, concurrency)
+	var wg sync.WaitGroup
+	wg.Add(concurrency)
```
skeggse, line 519: There's a natural split in the design of this command: some `s5cmd select` calls will match very few lines out of a lot of files, and would benefit significantly from more goroutines. However, some calls match many lines across fewer files, and in the interest of avoiding excess context switching it doesn't make sense to have the same huge number of goroutines there. Maybe there's a way to set this up such that there are fewer reader goroutines and many "discovery" goroutines, where the readers actually do the `json` message blob extraction and the discovery goroutines handle sending the `SelectObjectContent` request? I worry that in some cases that'll cause a significant delay between when we make the request and start reading from it, and that that'll cause timeouts on the S3 side.

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+			if fatalError != nil {
+				continue
+			}
+			if err := writev(to, record.Record, []byte{'\n'}); err != nil {
+				cancel()
+				fatalError = err
+				os.Stderr.Write([]byte(fmt.Sprintf("error writing: %v\n", err)))
+			}
+			// TODO: reinstate
+		} else if !IsCancelationError(record.Err) { //if !errorpkg.IsCancelation(record.Err) {
```
skeggse, line 549: Same `import`-related problem.

skeggse, line 549: There's a particular case where I think the network drops momentarily and a connection fails, and we don't currently retry. That causes the entire `select` command to fail.

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+
+	resultch := make(chan *SelectRecord, concurrency)
+	var wg sync.WaitGroup
+	wg.Add(concurrency)
+	for worker := 0; worker < concurrency; worker++ {
+		go func() {
+			defer wg.Done()
+			s.selectWorker(workerCtx, baseInput, objch, resultch)
+		}()
+	}
```
skeggse, line 525: It seems like it'd be nice to use `parallel` for the workers here, but when I looked at it it looked like I couldn't stream tasks into it without buffering as many tasks as would exist. I'm not convinced my understanding is correct, though. I'm happy to rewrite to use `parallel` (I was just trying to get this working), but I'd like to figure out the appropriate concurrency model for the aforementioned design problem around wanting to select from many many files in the case that there are fewer matches (to try and vastly speed up needle-in-a-haystack selects).

command/select.go
```diff
@@ -0,0 +1,163 @@
...
+		Expression:      s.query,
+		CompressionType: s.compressionType,
+	}
+	err = client.Select(ctx, objch, os.Stdout, &query, parallel.Size())
+	if err != nil {
+		return err
+	}
```
skeggse, line 131: It seems like there's probably a some use-case that writes to a file or set of files instead of `os.Stdout`. Thus far, `os.Stdout` has been more than sufficient for my use-cases.

skeggse (commented):
I'm looking for feedback on the right concurrency pattern, module import resolution, and maybe a little bit of error handling feedback. I've got _some_ exposure to Go, but not a ton. Best practices that I'm missing are welcome.

igungor:
Thanks for the PR and thorough comments! I'll try to review today.

storage/s3.go
igungor: we have an open ticket for this #51.

command/select.go
igungor: It's ok.

storage/s3.go
igungor: Let's do the optimization after this PR, if necessary of course.

command/select.go
igungor: I think of the `select` command like the `get` command but with server-side filtering. So writing to stdout sounds natural to me.

We can think about writing to file(s) (may be with a cli option) after a certain amount of time. Not sure.

command/select.go
igungor: I think we can do the JSON first and add others later, one by one. CSV is very common but I don't think its critical to support it at first pass.

A flag sounds good to me.

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+	// }()
+
+	var merror error
+	var fatalError error
+	for record := range resultch {
+		if record.Err == nil {
+			if fatalError != nil {
+				continue
+			}
+			if err := writev(to, record.Record, []byte{'\n'}); err != nil {
```
igungor, line 543: `writev` function looks unnecessary because you can write:

```
_, err := to.Write(append(record.Record, '\n'))
```

command/select.go
```diff
@@ -0,0 +1,163 @@
...
+`
+
+var selectCommandFlags = []cli.Flag{
+	&cli.StringFlag{
+		Name:    "query",
+		Aliases: []string{"e"},
+		Usage:   "SQL expression to use to select from the objects",
+	},
+	&cli.StringFlag{
+		Name:  "compression-type",
```
igungor, line 37: `--compression` might be better.

command/select.go
```diff
@@ -0,0 +1,163 @@
...
+		Usage: "Type of compression used in storage",
```
igungor, line 38:
```suggestion
		Usage: "input compression format",
```

command/select.go
```diff
@@ -0,0 +1,163 @@
...
+	},
+}
+
+var selectCommand = &cli.Command{
+	Name:               "select",
+	HelpName:           "select",
+	Usage:              "select objects containing JSON using a SQL query",
```
igungor, line 45:
```suggestion
	Usage:              "run SQL queries on objects",
```

storage/s3.go
```diff
@@ -14,6 +15,8 @@ import (
 	"sync"
 	"time"
 
+	// "github.com/segmentio/encoding/json"
```
igungor, line 18: https://github.com/peak/s5cmd/pull/300#discussion_r654318052

Please remove this line and run `go mod tidy && go mod vendor`. `go.mod` looks stale.

storage/s3.go
igungor: Will be unused after https://github.com/peak/s5cmd/pull/300#discussion_r654325903

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+		InputSerialization: &s3.InputSerialization{
+			CompressionType: aws.String(query.CompressionType),
+			JSON: &s3.JSONInput{
+				Type: aws.String("Lines"),
+			},
+		},
+		OutputSerialization: &s3.OutputSerialization{
+			JSON: &s3.JSONOutput{},
+		},
+		// RequestProgress: &s3.RequestProgress{
```
igungor, line 509: let's remove progress bar code.

storage/s3.go
skeggse: I guess I'm not terribly sure about this, but my impression was that this could be faster than copying all of `record.Record`. Probably just me (needlessly) microoptimizing.

command/select.go
igungor: Do you mean`s5cmd select -e '...' s3://bucket/prefix/` to expand to all objects under that prefix?

storage/s3.go
igungor: I sketched a prototype to demonstrate `parallel.Run` usage [here](https://github.com/peak/s5cmd/tree/select-proto1).

There may be some rough edges but overall, I think we can use the same structure as the `cp` command. Please take a look.

storage/s3.go
igungor: We need to see the generated code to see if this code needs optimization. I don't think it's worth it at this stage.

storage/s3.go
igungor: `errorpkg` is (implicitly) meant to be used at the command layer. Almost all of the storage errors bubble up to the callsite, handled there.

storage/s3.go
igungor: SDK retries all API calls. Are there other cases that we need to be aware of?

command/select.go
sonmezonur: Can you make  `--compression` flag optional?  (default: none)`select` command fails if we don't provide this flag

```
MissingRequiredParameter: CompressionType is required. Please check the service documentation and try again.
```

storage/s3.go
```diff
@@ -382,6 +386,175 @@ func (s *S3) Get(
...
+	}
+	return
+}
+
+func (s *S3) Select(ctx context.Context, objch <-chan *Object, to io.Writer, query *Query, concurrency int) error {
+	if s.dryRun {
+		return nil
+	}
+
+	baseInput := s3.SelectObjectContentInput{
```
ilkinulas, line 497: If we  move `s3.SelectObjectContentInput` creation  to `select.go` where `storage.Query` is built, adding support for other input/output serializations will be easier.

ilkinulas: > If we move `s3.SelectObjectContentInput` creation to `select.go` where `storage.Query` is built, adding support for other input/output serializations will be easier.

Ignore this suggestion please, `s3` details will be leaked to command level if we move `s3.SelectObjectContentInput` to `select.go`.

storage/s3.go
skeggse: LOL so removing `writev` results in a ~17% speedup for IO-heavy `select` calls on my machine (N=8)

storage/s3.go
skeggse: I've seen the select command fail on terabyte-scale datasets after a few minutes due to what seemed at the time like a network error. I have a stable gigabit connection, so I'd expect this to not happen if the SDK retries API calls...

command/select.go
skeggse: Yep!

storage/s3.go
skeggse: Okay, pulled that in and refactored to correctly segment records.

command/select.go
igungor: `s5cmd` expands the source if it contains a wildcard, just like the `cp` command.

`s5cmd cp s3://bucket/object .` -> `download object`
`s5cmd cp s3://bucket/prefix .` -> `error: no such key`
`s5cmd cp s3://bucket/prefix/ .` -> `error: source argument must contain wildcard character`

storage/s3.go
igungor: Thanks.

command/select.go
```diff
@@ -0,0 +1,213 @@
...
+	},
+}
+
+// Select holds select operation flags and states.
+type Select struct {
+	src         string
+	op          string
+	fullCommand string
+
+	deleteSource bool
```
igungor, line 87: unused field

storage/s3.go
```diff
@@ -382,6 +383,79 @@ func (s *S3) Get(
...
+
+				switch e := event.(type) {
+				case *s3.RecordsEvent:
+					writer.Write(e.Payload)
+				}
+			}
+		}
+	}()
+
+	decoder := json.NewDecoder(reader)
```
igungor, line 443: do we really need to unmarshall the payload?

command/select.go
```diff
@@ -0,0 +1,213 @@
...
+		return err
+	}
+
+	client, err := storage.NewRemoteClient(ctx, srcurl, s.storageOpts)
+	if err != nil {
+		printError(s.fullCommand, s.op, err)
+		return err
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
```
igungor, line 110: Could you fix the static analyzer errors please? In case you can't see the CI output, here's the `staticcheck` output:

```
Run make vet
# github.com/peak/s5cmd/command
command/select.go:110:2: the cancel function is not used on all paths (possible context leak)
command/select.go:115:3: this return statement may be reached without using the cancel var defined on line 110
# github.com/peak/s5cmd/storage
storage/s3.go:456:2: unreachable code
make: *** [Makefile:31: vet] Error 2
Error: Process completed with exit code 2.
```

storage/s3.go
igungor: Have you seen the error message at the time?

igungor:
@skeggse Hi. Have you had time the time to look into this?

skeggse:
Hey, thanks for your patience! I'll try and take a look later today.

storage/s3.go
skeggse: The idea is that we need to find record delimiters, and the byte slices we get back from the `SelectObjectContent` operation don't align to record boundaries. This is a relatively straightforward way to find those boundaries and output the entire record at once, without having multiple concurrent select operations interleave their output.

storage/s3.go
skeggse: I didn't record the error message, no :(

command/select.go
skeggse: 672b7ac

storage/s3.go
igungor: Good to know. Thanks.

storage/s3.go
igungor: I understand your concern but I don't have a solution other than introducing yet another flag to change *discovery* and *consumer* goroutine count.

igungor:
I think this PR is good to go. The only thing missing is the tests. Since `gofakes3` package doesn't support `SelectObject` calls, I'm okay with with PR as it is.

igungor (approved):

ilkinulas:
I am ok with this PR too, thank you @skeggse 
Just a little request before merging it:
 - Let's add an example usage to README
 - And it would be good to add a note saying we only support json file format.

storage/s3.go
skeggse: I don't think we need to solve this now; just commenting that this functionality is relatively important to make this command reasonably support a variety of use-cases (across the fraction space representing data scanned over data retrieved).

skeggse:
Updated to address @ilkinulas' comments!

README.md
```diff
@@ -204,6 +205,24 @@ folder hierarchy.
...
+`s5cmd` supports the `SelectObjectContent` S3 operation, and will run your
+[SQL query](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-glacier-select-sql-reference.html)
+against objects matching normal wildcard syntax and emit matching JSON records via stdout. Records
+from multiple objects will be interleaved, and order of the records is not guaranteed (though it's
+likely that the records from a single object will arrive in-order, even if interleaved with other
+records).
+
+    $ s5cmd select --compression GZIP \
+      --query "SELECT timestamp, hostname FROM S3Object WHERE ip_address LIKE '10.%' OR application='unprivileged'" \
+      s3://bucket-foo/object/2021/*
```
ilkinulas, line 219: I got an InvalidKeyPath error with this sample select query.
Query worked after updating as below:

```suggestion
    $ s5cmd select --compression GZIP \
      --query "SELECT s.timestamp, s.hostname FROM S3Object s WHERE s.ip_address LIKE '10.%' OR s.application='unprivileged'" \
      s3://bucket-foo/object/2021/*
```

ilkinulas (approved):

command/select.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		printError(s.fullCommand, s.op, err)
+		return err
+	}
+
+	var merror error
+
+	waiter := parallel.NewWaiter()
+	errDoneCh := make(chan bool)
+	writeDoneCh := make(chan bool)
```
sonmezonur, line 121: If I'm not missing something, this channel is not used. It is closed by goroutine but we don't check whether it is closed. Shouldn't we need to wait in parent just like we did in`errDoneCh`?

```go
waiter.Wait()
<-errDoneCh
<-writeDoneCh

return merror
```

command/select.go
```diff
@@ -0,0 +1,212 @@
...
+	srcurl, err := url.New(src)
+	if err != nil {
+		return err
+	}
+
+	if !srcurl.IsRemote() {
+		return fmt.Errorf("source must be remote")
+	}
+
+	if strings.ToUpper(c.String("format")) != "JSON" {
```
sonmezonur, line 207:
```suggestion
      if !strings.EqualFold(c.String("format"), "JSON") {
```

command/select.go
skeggse: Yes, I think you're right.

sonmezonur (approved):

igungor:
Thank you!

PR #311, opened 2021-07-04
Add steps for installation using MacPorts

manojkarthick:
Hi! I'm a macOS user and I noticed that this tool is available on Homebrew but was missing from MacPorts. So I took the liberty of adding the package. This PR updates the README with installation steps. 

Relevant links: 
* MacPorts: https://www.macports.org/
* Installation Config: https://github.com/macports/macports-ports/blob/master/net/s5cmd/Portfile

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,15 @@ For macOS, a [homebrew](https://brew.sh) tap is provided:
     brew tap peak/s5cmd https://github.com/peak/s5cmd
     brew install s5cmd
 
+### MacPorts
+
+You can also install `s5cmd` from [MacPorts](https://ports.macports.org/port/s5cmd/summary) on macOS:
+
+    sudo port selfupdate
+    sudo port install s5cmd
+
+NOTE: MacPorts is not officially supported. The versions might be out of date compared to Homebrew.
+
 ### Build from source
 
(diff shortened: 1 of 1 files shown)


sonmezonur (approved):
@manojkarthick Hello! Thank you for your contribution.

igungor:
Our release process (`goreleaser` under the hood) **automatically** creates tarballs, updates Docker images and brew formula. Unfortunately [it doesn't support `macports`](https://github.com/goreleaser/goreleaser/issues/1556). That's why it's missing from the readme. Also the fact that the whole team uses `brew` :)

I don't know much about `macports` but I see that `s5cmd` is in the index, so the Portfile needs to be updated after each release, right? I'm not sure we'd do that. If we merge this, we should mention that `macports` is not officially supported, so updates may fall behind.

igungor:
>  If we merge this, we should mention that macports is not officially supported, so updates may fall behind.

Hi @manojkarthick. Is the above approach sound good to you? If so, would you update the readme please?

manojkarthick:
> > If we merge this, we should mention that macports is not officially supported, so updates may fall behind.
> 
> Hi @manojkarthick. Is the above approach sound good to you? If so, would you update the readme please?

Sounds good @igungor - I've updated the PR with a note on support.

igungor:
Thanks!

PR #312, opened 2021-07-06
command/cp: add `force-glacier-transfer` flag

### LINKED ISSUE #206: when object is restored from glacier, cant download it.
We restored some items from glacier, and wanted to download it (console allows download)
But s5cmd throws error because of:
https://github.com/peak/s5cmd/blob/601ee14f130b21739c4c315218ccdfb71ac41afa/command/cp.go#L248

ocakhasan:
resolves #206. Tests are not included. Not sure how to do it.

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -133,7 +133,7 @@ var copyCommandFlags = []cli.Flag{
 	},
 	&cli.BoolFlag{
 		Name:  "force-glacier-transfer",
-		Usage: "try to copy files which is restored from GLACIER storage class",
+		Usage: "force transfer of GLACIER objects whether they are restored or not",
 	},
 	&cli.StringFlag{
 		Name:  "source-region",
@@ -293,7 +293,6 @@ func (c Copy) Run(ctx context.Context) error {
 			continue
 		}
 
-		
 		if object.StorageClass.IsGlacier() && !c.forceGlacierTransfer {
 			err := fmt.Errorf("object '%v' is on Glacier storage", object)
 			printError(c.fullCommand, c.op, err)
--- a/command/cp.go
+++ b/command/cp.go
@@ -76,6 +76,9 @@ Examples:
 
 	13. Perform KMS-SSE of the object(s) at the destination using customer managed Customer Master Key (CMK) key id
 		> s5cmd {{.HelpName}} --sse aws:kms --sse-kms-key-id <your-kms-key-id> s3://bucket/object s3://target-bucket/prefix/object
+	
+	14. Force transfer of GLACIER objects whether they are restored or not
+		> s5cmd {{.HelpName}} --force-glacier-transfer s3://bucket/prefix target-directory/
 `
 
 var copyCommandFlags = []cli.Flag{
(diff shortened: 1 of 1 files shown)


sonmezonur (approved):

command/cp.go
```diff
@@ -131,6 +131,10 @@ var copyCommandFlags = []cli.Flag{
 		Name:  "acl",
 		Usage: "set acl for target: defines granted accesses and their types on different accounts/groups",
 	},
+	&cli.BoolFlag{
+		Name:  "force-glacier-transfer",
+		Usage: "omit checks for glacier objects",
```
seruman, line 136: A sentence stating that it would try to issue copy for files with GLACIER storage class whether they're restored or not would be better.

seruman (changes requested):

command/cp.go
```diff
@@ -287,7 +293,9 @@ func (c Copy) Run(ctx context.Context) error {
 			continue
 		}
 
-		if object.StorageClass.IsGlacier() {
+		// if force glacier transfer flag is set, it will omit check for the if block
+		// if not set, then it will check whether the object is glacier.
+		if !c.forceGlacierTransfer && object.StorageClass.IsGlacier() {
```
seruman, line 298:
```suggestion
		if object.StorageClass.IsGlacier() && !c.forceGlacierTransfer {
```

IMHO it's easier on the eyes when the checks are swapped and the checks are pretty self explanatory making comment redundant.

command/cp.go
ocakhasan: changed to `try to copy files which is restored from GLACIER storage class`.

command/cp.go
ocakhasan: changed the order of conditions and removed the comment lines.

command/cp.go
seruman: It kinda sounds like the option only works on restored objects. With force option, if ile is not restored, S3 would still raise an error;
```
InvalidObjectState: The operation is not valid for the object's storage class status
```
I think we should indicate we do not check restore status and issue a copy request anyways.

seruman (commented):
Lastly, could you also add a usage example to help output?

seruman (approved):

command/cp.go
```diff
@@ -76,6 +76,9 @@ Examples:
 
 	13. Perform KMS-SSE of the object(s) at the destination using customer managed Customer Master Key (CMK) key id
 		> s5cmd {{.HelpName}} --sse aws:kms --sse-kms-key-id <your-kms-key-id> s3://bucket/object s3://target-bucket/prefix/object
+	
+	14. Force transfer of GLACIER objects whether they are restored or not
+		> s5cmd {{.HelpName}} --force-glacier-transfer s3://bucket/prefix target-directory/
```
igungor, line 81: This command won't work. `s5cmd` doesn't expand prefix if no wildcard is given.

igungor (changes requested):

command/cp.go
ocakhasan: Changed to `s5cmd {{.HelpName}} --force-glacier-transfer s3://bucket/prefix/* target-directory/` with explanation
`Force transfer of GLACIER objects with a prefix whether they are restored or not`.

igungor (approved):

sonmezonur:
🚀

PR #315, opened 2021-07-08
command/cp&rm: add `--raw` flag

### LINKED ISSUE #235: disable wildcard and use "*" as literal?
The S3 objects I'm going to download has "\*" in object key. When I use `cp` to download, `s5cmd` expands it and download multiple objects more than the particular object. Is there a way to escape wildcard "\*" or disable wildcard? Thanks.

ocakhasan:
- command/cp&rm:  new flag `--raw` which disables the wildcard operations.
- also changed the `rb` command's helper text to correct form, based on [this commit](https://github.com/peak/s5cmd/commit/68f797b0de6b26f311076012928a9aa3aaad2713).

resolves #235

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -139,6 +139,10 @@ var copyCommandFlags = []cli.Flag{
 		Name:  "destination-region",
 		Usage: "set the region of destination bucket: the region of the destination bucket will be automatically discovered if --destination-region is not specified",
 	},
+	&cli.BoolFlag{
+		Name:  "raw",
+		Usage: "disable the wildcard operations, useful with filenames that contains glob characters.",
+	},
 }
 
 var copyCommand = &cli.Command{
@@ -175,9 +179,11 @@ var copyCommand = &cli.Command{
 			encryptionMethod: c.String("sse"),
 			encryptionKeyID:  c.String("sse-kms-key-id"),
 			acl:              c.String("acl"),
+			raw:              c.Bool("raw"),
+
 			// region settings
-			srcRegion:        c.String("source-region"),
-			dstRegion:        c.String("destination-region"),
+			srcRegion: c.String("source-region"),
+			dstRegion: c.String("destination-region"),
 
 			storageOpts: NewStorageOpts(c),
 		}.Run(c.Context)
@@ -203,6 +209,7 @@ type Copy struct {
 	encryptionMethod string
 	encryptionKeyID  string
 	acl              string
+	raw              bool
 
 	// region settings
 	srcRegion string
@@ -223,12 +230,16 @@ increase the open file limit or try to decrease the number of workers with
 // Run starts copying given source objects to destination.
 func (c Copy) Run(ctx context.Context) error {
 	srcurl, err := url.New(c.src)
+	fmt.Printf("srcurl : %#v\n", srcurl)
+
 	if err != nil {
 		printError(c.fullCommand, c.op, err)
 		return err
 	}
 
 	dsturl, err := url.New(c.dst)
+	fmt.Printf("dsturl : %#v\n", dsturl)
+
 	if err != nil {
 		printError(c.fullCommand, c.op, err)
 		return err
@@ -244,7 +255,7 @@ func (c Copy) Run(ctx context.Context) error {
 		return err
 	}
 
-	objch, err := expandSource(ctx, client, c.followSymlinks, srcurl)
+	objch, err := expandSource(ctx, client, c.followSymlinks, srcurl, c.raw)
 	if err != nil {
 		printError(c.fullCommand, c.op, err)
 		return err
@@ -271,6 +282,7 @@ func (c Copy) Run(ctx context.Context) error {
 		}
 	}()
 
+	// use raw flag to handle glob operations.
 	isBatch := srcurl.HasGlob()
 	if !isBatch && !srcurl.IsRemote() {
 		obj, _ := client.Stat(ctx, srcurl)
@@ -278,6 +290,7 @@ func (c Copy) Run(ctx context.Context) error {
 	}
 
 	for object := range objch {
+		fmt.Printf("object is %#v\n", object.URL)
 		if object.Type.IsDir() || errorpkg.IsCancelation(object.Err) {
 			continue
 		}
@@ -348,7 +361,7 @@ func (c Copy) prepareDownloadTask(
 		if err != nil {
 			return err
 		}
-
+		fmt.Println("I am here 1.")
 		err = c.doDownload(ctx, srcurl, dsturl)
 		if err != nil {
 			return &errorpkg.Error{
@@ -385,6 +398,7 @@ func (c Copy) prepareUploadTask(
 
 // doDownload is used to fetch a remote object and save as a local object.
 func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) error {
+	fmt.Println("I am inside doDownload.")
 	srcClient, err := storage.NewRemoteClient(ctx, srcurl, c.storageOpts)
 	if err != nil {
 		return err
@@ -403,6 +417,8 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) 
 	}
 
 	file, err := dstClient.Create(dsturl.Absolute())
+	fmt.Println("dsturl.Absolute()=", dsturl.Absolute())
+	fmt.Println("file create error is ", err)
 	if err != nil {
 		return err
 	}
@@ -640,6 +656,7 @@ func prepareLocalDestination(
 	isBatch bool,
 	storageOpts storage.Options,
 ) (*url.URL, error) {
+	fmt.Println("I am here prepare local destination.")
 	objname := srcurl.Base()
 	if isBatch && !flatten {
 		objname = srcurl.Relative()
@@ -649,6 +666,7 @@ func prepareLocalDestination(
 
 	if isBatch {
 		err := client.MkdirAll(dsturl.Absolute())
+		fmt.Println("Error in mkdirAll=", err)
 		if err != nil {
 			return nil, err
 		}
@@ -660,8 +678,14 @@ func prepareLocalDestination(
 	}
 
 	if isBatch && !flatten {
+		fmt.Println(objname)
 		dsturl = dsturl.Join(objname)
+		fmt.Println("dsturl=", dsturl)
+		fmt.Println(dsturl.Dir())
+		fmt.Println("dsturl.Dir()=", dsturl.Dir())
+
 		err := client.MkdirAll(dsturl.Dir())
+		fmt.Println("Error in mkdirAll2=", err)
 		if err != nil {
 			return nil, err
 		}
@@ -669,6 +693,7 @@ func prepareLocalDestination(
 
 	if err == storage.ErrGivenObjectNotFound {
 		err := client.MkdirAll(dsturl.Dir())
+		fmt.Println("Error in mkdirAll3=", err)
 		if err != nil {
 			return nil, err
 		}
--- a/command/cp.go
+++ b/command/cp.go
@@ -230,7 +230,6 @@ increase the open file limit or try to decrease the number of workers with
 // Run starts copying given source objects to destination.
 func (c Copy) Run(ctx context.Context) error {
 	srcurl, err := url.New(c.src)
-	fmt.Printf("srcurl : %#v\n", srcurl)
 
 	if err != nil {
 		printError(c.fullCommand, c.op, err)
@@ -238,7 +237,6 @@ func (c Copy) Run(ctx context.Context) error {
 	}
 
 	dsturl, err := url.New(c.dst)
-	fmt.Printf("dsturl : %#v\n", dsturl)
 
 	if err != nil {
 		printError(c.fullCommand, c.op, err)
@@ -284,14 +282,12 @@ func (c Copy) Run(ctx context.Context) error {
 
 	// use raw flag to handle glob operations.
 	isBatch := !c.raw && srcurl.HasGlob()
-	fmt.Println("isBatch", isBatch)
 	if !isBatch && !srcurl.IsRemote() {
 		obj, _ := client.Stat(ctx, srcurl)
 		isBatch = obj != nil && obj.Type.IsDir()
 	}
 
 	for object := range objch {
-		fmt.Printf("object is %#v\n", object.URL)
 		if object.Type.IsDir() || errorpkg.IsCancelation(object.Err) {
 			continue
 		}
@@ -362,7 +358,6 @@ func (c Copy) prepareDownloadTask(
 		if err != nil {
 			return err
 		}
-		fmt.Println("I am here 1.")
 		err = c.doDownload(ctx, srcurl, dsturl)
 		if err != nil {
 			return &errorpkg.Error{
@@ -399,7 +394,6 @@ func (c Copy) prepareUploadTask(
 
 // doDownload is used to fetch a remote object and save as a local object.
 func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) error {
-	fmt.Println("I am inside doDownload.")
 	srcClient, err := storage.NewRemoteClient(ctx, srcurl, c.storageOpts)
 	if err != nil {
 		return err
@@ -418,8 +412,6 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) 
 	}
 
 	file, err := dstClient.Create(dsturl.Absolute())
-	fmt.Println("dsturl.Absolute()=", dsturl.Absolute())
-	fmt.Println("file create error is ", err)
 	if err != nil {
 		return err
 	}
--- a/command/cp.go
+++ b/command/cp.go
@@ -258,12 +258,8 @@ increase the open file limit or try to decrease the number of workers with
 
 // Run starts copying given source objects to destination.
 func (c Copy) Run(ctx context.Context) error {
-	var urlMode url.URLMode
-	if c.raw {
-		urlMode = 1
-	}
-	srcurl, err := url.New(c.src, url.WithMode(urlMode))
 
+	srcurl, err := url.New(c.src, url.WithMode(url.URLMode(c.raw)))
 	if err != nil {
 		printError(c.fullCommand, c.op, err)
 		return err
--- a/command/cp.go
+++ b/command/cp.go
@@ -76,6 +76,9 @@ Examples:
 
 	13. Perform KMS-SSE of the object(s) at the destination using customer managed Customer Master Key (CMK) key id
 		> s5cmd {{.HelpName}} --sse aws:kms --sse-kms-key-id <your-kms-key-id> s3://bucket/object s3://target-bucket/prefix/object
+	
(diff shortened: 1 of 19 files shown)


command/cp.go
```diff
@@ -280,7 +288,8 @@ func (c Copy) Run(ctx context.Context) error {
 		}
 	}()
 
-	isBatch := srcurl.HasGlob()
+	// use raw flag to handle glob operations.
```
aykutfarsak, line 291:
```suggestion
	// use --raw flag to prevent glob operations.
```
or you can remove the comment entirely.

command/expand.go
```diff
@@ -16,22 +16,28 @@ func expandSource(
...
-	// if the source is local, we send a Stat call to know if  we have
-	// directory or file to walk. For remote storage, we don't want to send
-	// Stat since it doesn't have any folder semantics.
-	if !srcurl.HasGlob() && !srcurl.IsRemote() {
-		obj, err := client.Stat(ctx, srcurl)
-		if err != nil {
-			return nil, err
+
+	// If wildcard operations are disabled (raw flag is set), then program deals with 1 file.
+	// Therefore, there is no need to check the directory operations.
```
aykutfarsak, line 23:
```suggestion
	// if wildcard operations are disabled with the raw flag, then the program deals with 1 file.
	// therefore, there is no need to check if it is a directory.
```
I am not sure about the comment but let's not start with capital letters.

command/expand_test.go
```diff
@@ -328,6 +329,86 @@ func TestExpandSource_Do_Not_Follow_Symlinks(t *testing.T) {
 	assert.Equal(t, []string{workdirJoin}, expected)
 }
 
+func TestExpandSourceWithRawFlagTrue(t *testing.T) {
```
aykutfarsak, line 332:
```suggestion
func TestExpandSourceWithRawFlag(t *testing.T) {
```

command/expand_test.go
```diff
@@ -328,6 +329,86 @@ func TestExpandSource_Do_Not_Follow_Symlinks(t *testing.T) {
...
+	assert.Equal(t, len(ch), 1)
+	var expected []string
+	for obj := range ch {
+		expected = append(expected, obj.URL.Absolute())
+	}
+	workdirJoin := filepath.ToSlash(workdir.Join("a/file*.txt"))
+	assert.Equal(t, []string{workdirJoin}, expected)
+}
+
+func TestExpandSourceWithRawFlagFalse(t *testing.T) {
```
aykutfarsak, line 370:
```suggestion
func TestExpandSourceWithoutRawFlag(t *testing.T) {
```

ocakhasan: changed to `TestExpandSourceWithoutRawFlag`

command/expand_test.go
ocakhasan: changed to `TestExpandSourceWithRawFlag`.

command/expand.go
ocakhasan: fixed the comments as you suggested.

command/cp.go
ocakhasan: changed the comment as you suggested.

command/expand.go
```diff
@@ -16,22 +16,28 @@ func expandSource(
 	client storage.Storage,
 	followSymlinks bool,
 	srcurl *url.URL,
+	isRaw bool,
```
ilkinulas, line 19: There is another way to implement this `--raw` flag: 
Instead of introducing a new boolean parameter to this function we can create a new function that just returns a channel with single object. `cp` and `rm` commands could check `--raw` flag and decide to call either `expandSource` or this new function (`rawSource` for example).

```go
func rawSource(srcurl *url.URL, followSymlinks bool) <-chan *storage.Object {
	ch := make(chan *storage.Object, 1)
	if storage.ShouldProcessUrl(srcurl, followSymlinks) {
		ch <- &storage.Object{URL: srcurl}
	}
	close(ch)
	return ch
}
```

This implementation has following advantages:
- `expandSource` function does what it's name says
- No need to update `expandSource` tests

One drawback that comes to my mind is `cp` and `rm` commands will have one additional check for `--raw` flag, but it would not be more complicated that updating `expandSource` function. 

What do you think?

ocakhasan: I think this approach is better than with parameter, because in the other [PR](https://github.com/peak/s5cmd/pull/324), I also had to add another parameter (`exclude`) to `expandSource`, and if I add parameters to `expandSource` like this, then it would be very hard to follow. I will change to new function as you suggested.

command/expand.go
```diff
@@ -89,3 +90,25 @@ func expandSources(
...
+	ch := make(chan *storage.Object, 1)
+	if storage.ShouldProcessUrl(srcurl, followSymlinks) {
+		ch <- &storage.Object{URL: srcurl}
+	}
+	close(ch)
+	return ch
+}
+
+// raw source returns the only object with given path.
+func rawSourceUrls(srcurls []*url.URL, followSymlinks bool) <-chan *storage.Object {
```
ilkinulas, line 105: Let's try merging `rawSource` and `rawSources` functions into a single variadic function.

```suggestion
func rawSource(followSymlinks bool, srcurls ...*url.URL) <-chan *storage.Object {
```

We can use this function in both cp/rm commands and `expandSource`  function.

ocakhasan: created a new single variadic function.

command/expand.go
```diff
@@ -89,3 +90,15 @@ func expandSources(
 
 	return ch
 }
+
+// raw source returns the only object with given path.
+func rawSource(followSymlinks bool, srcurls ...*url.URL) <-chan *storage.Object {
```
ilkinulas, line 95: Let's use `rawSource`  to eliminate code duplication in  `expandSource`.

`expand.go` line 38
```go
return rawSource(followSymlinks, srcurl), nil
```

command/rm.go
```diff
@@ -85,7 +95,10 @@ func (d Delete) Run(ctx context.Context) error {
 		return err
 	}
 
-	objChan := expandSources(ctx, client, false, srcurls...)
+	objChan := rawSource(false, srcurls...)
```
ilkinulas, line 98: Let's revert this raw check to make it more readable.

```go
declare objch channel
if raw {
	objch = rawSource ...
} else {
	objch, err = expandSource ..
}
```

command/expand.go
ocakhasan: changed to `return rawSource(followSymlinks, srcurl), nil` to avoid code duplication.

command/rm.go
ocakhasan: changed as you suggested.

e2e/cp_test.go
```diff
@@ -2926,3 +2926,311 @@ func TestCopyS3ToDirDryRun(t *testing.T) {
...
+
+			// assert filesystem
+			expected := fs.Expected(t, tc.src...)
+			assert.Assert(t, fs.Equal(workdir.Path(), expected))
+		})
+	}
+
+}
+
+// When folder is uploaded with --raw flag, it gives an error.
```
sonmezonur, line 3049: Why? I think it should copy files in `a*` directory in raw mode.

sonmezonur (commented):
Thanks for the PR :) I tested your branch and it works perfectly fine but I think we can still make some improvements:

1 ) I think we should allow glob characters for destination URL if `--raw` flag is given. Currently we are getting error:
```
❯ ./s5cmd cp --raw test 's3://bigtmp/prefix/test*'
ERROR "cp test s3://bigtmp/prefix/test*": target "s3://bigtmp/prefix/test*" can not contain glob characters
```
2) In raw mode, we are still parsing wildcard strings and compiling regexes even if they are not used. (url.New()  makes lots of parsing and regex calculations([ref](https://github.com/peak/s5cmd/blob/master/storage/url/url.go#L103))). IMO, URL package should know if it is a raw mode and should not make unnecessary/expensive calculations. I prototyped URL package to get optional `URLMode` parameter and skip parsing if `URLMode` is raw ([Prototype](https://github.com/peak/s5cmd/blob/url_mode_prototype/storage/url/url.go)). If we prefer using this method, we don't need to update `command/expand.go` anymore.  If we decide to add raw mode to other commands (like ls, cat, etc.) in near future, it will be quite easy since URL package is used by all other commands (while expandSource is only used by cp and rm). Does this make sense to you?

3) Could you add tests for remote->remote copy operations? As far as I can see, we have tests for local->remote and remote-> local but could not see remote->remote.

e2e/cp_test.go
ocakhasan: Thanks for the feedback. I changed as you suggested and here is my thoughts.

1. Now we can have wildcard in the destination in `--raw` mode. 

2. Changing URL package is simpler than adding new functions or changing `expandSource` function. It is working good and adding `--raw` flag to other commands is much more easier now. 

3. Added tests for remote -> remote. 

Also, local directory `a*` is uploaded now without uploading other folders matching `a*` pattern.

storage/url/url.go
```diff
@@ -45,10 +44,26 @@ type URL struct {
 	relativePath string
 	filter       string
 	filterRegex  *regexp.Regexp
+	mode         URLMode
+}
+
+type URLMode int
```
ilkinulas, line 50: using a bool field instead of URLMode type will simplify checks below:
```go
    var urlMode url.URLMode
    if c.raw {
        urlMode = 1
    }
    srcurl, err := url.New(c.src, url.WithMode(urlMode))
```

Above sample can be written as this if we use a bool field.
```go
srcurl, err := url.New(c.src, url.WithMode(c.raw))
```

command/expand.go
ilkinulas: Please revert `expand.go`, implementation is moved to `url`.

storage/url/url.go
ocakhasan: changed as you suggested.

command/expand_test.go
```diff
@@ -328,6 +329,70 @@ func TestExpandSource_Do_Not_Follow_Symlinks(t *testing.T) {
 	assert.Equal(t, []string{workdirJoin}, expected)
 }
 
+func TestRawSourceWithoutSymLinks(t *testing.T) {
```
ilkinulas, line 332: Do we still need these tests?

ocakhasan: We use `rawSource` method in `expandSource` but it was due to eliminate code duplication. With the change in url package there is no need to use `rawSource` method, therefore I can remove it and change the `expandSource` method to the normal version.

command/cp.go
```diff
@@ -750,7 +759,7 @@ func validateCopyCommand(c *cli.Context) error {
 	}
 
 	// wildcard destination doesn't mean anything
-	if dsturl.HasGlob() {
+	if !c.Bool("raw") && dsturl.HasGlob() {
```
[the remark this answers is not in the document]
sonmezonur: Also, please create URLs with `urlMode` option in validateCopyCommand ([1](https://github.com/peak/s5cmd/blob/ae2bcea77d920afcfef1444a7749d51fe20f44ca/command/cp.go#L751), [2](https://github.com/peak/s5cmd/blob/ae2bcea77d920afcfef1444a7749d51fe20f44ca/command/cp.go#L756)). 

Validation fails if target URL has glob in --raw mode: 
```
❯ ./s5cmd cp --raw 'test*' 's3://bigtmp/test*'
ERROR "cp test* s3://bigtmp/test*": target "s3://bigtmp/test*" must be a bucket or a prefix
```

After updating with urlMode:
```
❯ ./s5cmd cp --raw 'test*' 's3://bigtmp/test*'
cp test* s3://bigtmp/test*
```

Also, we can add simple test for this scenario.

storage/url/url.go
```diff
@@ -291,7 +312,7 @@ func (u *URL) MarshalJSON() ([]byte, error) {
 
 // HasGlob reports whether if a string contains any wildcard chars.
 func (u *URL) HasGlob() bool {
```
sonmezonur, line 314:
```suggestion
func (u *URL) IsWildcard() bool {
```
At this stage, I think we can rename this function since it also checks raw mode. Not sure about naming :D

command/expand.go
```diff
@@ -17,6 +17,7 @@ func expandSource(
 	followSymlinks bool,
 	srcurl *url.URL,
 ) (<-chan *storage.Object, error) {
+
```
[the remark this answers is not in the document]
ocakhasan: deleted empty line.

command/cp.go
```diff
@@ -253,13 +258,15 @@ increase the open file limit or try to decrease the number of workers with
 
 // Run starts copying given source objects to destination.
 func (c Copy) Run(ctx context.Context) error {
-	srcurl, err := url.New(c.src)
+
```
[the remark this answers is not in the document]
ocakhasan: deleted empty line.

storage/url/url.go
```diff
@@ -1,4 +1,3 @@
-// Package url abstracts local and remote file URLs.
```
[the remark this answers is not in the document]
ocakhasan: added package comment.

command/cp.go
```diff
@@ -253,13 +258,15 @@ increase the open file limit or try to decrease the number of workers with
...
+	srcurl, err := url.New(c.src, url.WithRaw(c.raw))
 	if err != nil {
 		printError(c.fullCommand, c.op, err)
 		return err
 	}
 
 	dsturl, err := url.New(c.dst)
```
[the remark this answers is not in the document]
ocakhasan: added `url.WithRaw(c.raw)` to url creation.

command/cp.go
```diff
@@ -302,7 +311,8 @@ func (c Copy) Run(ctx context.Context) error {
 		}
 	}()
 
-	isBatch := srcurl.HasGlob()
+	// use --raw flag to prevent glob operations.
+	isBatch := !c.raw && srcurl.HasGlob()
```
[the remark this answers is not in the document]
ocakhasan: changed as you suggested.

command/cp.go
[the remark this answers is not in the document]
ocakhasan: changed as you suggested. and added a new test case in `cp_test.go`.

storage/url/url.go
ocakhasan: I am also not sure about the name, so if we can agree on a name, then I will change it 😄

command/rm.go
```diff
@@ -65,14 +74,16 @@ type Delete struct {
...
+	raw         bool
 
 	// storage options
 	storageOpts storage.Options
 }
 
 // Run remove given sources.
 func (d Delete) Run(ctx context.Context) error {
-	srcurls, err := newURLs(d.src...)
+
```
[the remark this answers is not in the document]
ocakhasan: deleted the empty line.

storage/url/url_test.go
```diff
@@ -402,3 +402,33 @@ func TestURLIsBucket(t *testing.T) {
 		}
 	}
 }
+
+func TestURLWithMode(t *testing.T) {
+	tests := []struct {
+		input          string
+		urlMode        bool
```
ilkinulas, line 409:
```suggestion
		raw        bool
```

ocakhasan: changed to `raw`.

e2e/cp_test.go
```diff
@@ -2926,3 +2926,460 @@ func TestCopyS3ToDirDryRun(t *testing.T) {
...
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+
+	assertLines(t, result.Stdout(), map[int]compareFunc{
+		0: equals("cp %v/file.txt %v/a*/file.txt", srcpath, dstpath),
+		1: equals("cp %v/file1.txt %v/a*/file1.txt", srcpath, dstpath),
+	}, sortInput(true))
+
+	// assert no files are in S3
```
ilkinulas, line 3093: we do not need this comment
```suggestion
```

ocakhasan: deleted the comment line.

storage/url/url.go
sonmezonur: My choice is `IsWildcard()`

e2e/cp_test.go
```diff
@@ -2926,3 +2926,459 @@ func TestCopyS3ToDirDryRun(t *testing.T) {
...
+		},
+		{
+			name:       "cp  a*/file.txt s3://bucket/",
+			src:        []string{"a*/file.txt", "a*/file1.txt", "a*/file2.txt"},
+			wantedFile: "a*/file.txt",
+			rawFlag:    "--raw",
+			expectedFiles: []fs.PathOp{
+				fs.WithFile("file.txt", fileContent),
+			},
+		},
```
ilkinulas, line 3159:
```suggestion
		{
			name:       "cp  a*/file.txt s3://bucket/",
			src:        []string{"a*/file*.txt", "a*b/file1.txt", "a*c/file2.txt"},
			wantedFile: "a*/file*.txt",
			rawFlag:    "--raw",
			expectedFiles: []fs.PathOp{
				fs.WithFile("file*.txt", fileContent),
			},
		},
```
This tests raw mode better, what do you think?

storage/url/url.go
ocakhasan: `HasGlob()` is changed to `IsWildcard()`

e2e/cp_test.go
ocakhasan: I think you are right. I tried to change other tests also.

e2e/cp_test.go
```diff
@@ -2926,3 +2926,424 @@ func TestCopyS3ToDirDryRun(t *testing.T) {
...
+				fs.WithFile("file*2.txt", "content"),
+			},
+			wantedFile:       "file*.txt",
+			dst:              "s3://bucket/",
+			expectedFiles:    []string{"file*.txt", "file*1.txt", "file*file.txt", "file*2.txt"},
+			nonExpectedFiles: []string{},
+			rawFlag:          "",
+		},
+		{
+			name: "cp  a*/file.txt s3://bucket/",
```
ilkinulas, line 2979:
```suggestion
			name: "cp  a*/file*.txt s3://bucket/",
```

e2e/cp_test.go
```diff
@@ -2926,3 +2926,424 @@ func TestCopyS3ToDirDryRun(t *testing.T) {
...
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			s3client, s5cmd, cleanup := setup(t)
+			defer cleanup()
+
+			createBucket(t, s3client, bucket)
+
+			workdir := fs.NewDir(t, tc.name, tc.src...)
```
ilkinulas, line 3012:
```suggestion
			workdir := fs.NewDir(t, "copy-raw-test", tc.src...)
```

e2e/cp_test.go
ocakhasan: changed as you suggested.

e2e/cp_test.go
ocakhasan: changed as you suggested.

ilkinulas (approved):

command/cp.go
```diff
@@ -302,7 +309,8 @@ func (c Copy) Run(ctx context.Context) error {
 		}
 	}()
 
-	isBatch := srcurl.HasGlob()
+	// use --raw flag to prevent glob operations.
```
sonmezonur, line 312: you can remove this comment since we don't check --raw here anymore.

ocakhasan: deleted comment line.

sonmezonur (approved):

PR #317, opened 2021-07-09
fix lost region

tombokombo:
Hi this PR fixes lost region, Options opts.region are used only internally for cp src dst region, there is no cmd flag for region,  so setSessionRegion is called for command like ls, run etc. If the region is set via env or config file is never used, func always fallback to aws-sdk s3manager.GetBucketRegion, but aws sdk will not recognize non aws regions like nl-ams (scaleway).

### DIFF
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -717,7 +717,9 @@ func (sc *SessionCache) clear() {
 }
 
 func setSessionRegion(ctx context.Context, sess *session.Session, bucket string) error {
-	if aws.StringValue(sess.Config.Region) == "" {
+	if aws.StringValue(sess.Config.Region) != "" {
+		return nil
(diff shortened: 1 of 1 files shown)


tombokombo:
@igungor @seruman guys, could you please review this bugfix? Thanks.

igungor:
Sorry for the late reply @tombokombo but I was on vacation. There're a couple of issues related with region-detection. To address them, there's an open ticket: https://github.com/peak/s5cmd/issues/320.

We're actively working on a pull-request to fix all of them I hope.

tombokombo:
great, thx

sonmezonur:
Closing this PR since https://github.com/peak/s5cmd/pull/331 addresses all region-related issues.

PR #318, opened 2021-07-09
add cache-control and expires options

tombokombo:
Hello, this PR add support for setting cache-control and expires header to s3 object, same as aws cli
```bash
s5cmd cp --acl "public-read" --expires "2024-10-01T20:30:00Z" --cache-control "public, max-age=345600, s-maxage=345600" foobar
  s3://example-bucket/foobar
```

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -134,6 +134,14 @@ var copyCommandFlags = []cli.Flag{
 		Name:  "acl",
 		Usage: "set acl for target: defines granted accesses and their types on different accounts/groups",
 	},
+	&cli.StringFlag{
+		Name:  "cache-control",
+		Usage: "set cache control for target: defines cache control header for object",
+	},
+	&cli.StringFlag{
+		Name:  "expires",
+		Usage: "set expires for target (uses RFC3339 format): defines expires header for object",
+	},
 	&cli.BoolFlag{
 		Name:  "force-glacier-transfer",
 		Usage: "force transfer of GLACIER objects whether they are restored or not",
@@ -183,6 +191,8 @@ var copyCommand = &cli.Command{
 			encryptionKeyID:      c.String("sse-kms-key-id"),
 			acl:                  c.String("acl"),
 			forceGlacierTransfer: c.Bool("force-glacier-transfer"),
+			cacheControl:         c.String("cache-control"),
+			expires:              c.String("expires"),
 			// region settings
 			srcRegion: c.String("source-region"),
 			dstRegion: c.String("destination-region"),
@@ -212,6 +222,9 @@ type Copy struct {
 	encryptionKeyID      string
 	acl                  string
 	forceGlacierTransfer bool
+	cacheControl         string
+	expires              string
+
 
 	// region settings
 	srcRegion string
@@ -472,7 +485,9 @@ func (c Copy) doUpload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) er
 		SetStorageClass(string(c.storageClass)).
 		SetSSE(c.encryptionMethod).
 		SetSSEKeyID(c.encryptionKeyID).
-		SetACL(c.acl)
+		SetACL(c.acl).
+		SetCacheControl(c.cacheControl).
+		SetExpires(c.expires)
 
 	err = dstClient.Put(ctx, file, dsturl, metadata, c.concurrency, c.partSize)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


tombokombo:
@aykutfarsak @ilkinulas guys could you please review, thanks.

ilkinulas:
> @aykutfarsak @ilkinulas guys could you please review, thanks.

@tombokombo there has been a lot of development in s5cmd these days, this PR is waiting in the queue, I will take a look at it soon.

command/cp.go
```diff
@@ -134,6 +134,14 @@ var copyCommandFlags = []cli.Flag{
 		Name:  "acl",
 		Usage: "set acl for target: defines granted accesses and their types on different accounts/groups",
 	},
+	&cli.StringFlag{
```
ilkinulas, line 137: Could you please add a sample usage for both `cache-control` and `expires` to the help text?

ilkinulas (changes requested):

ilkinulas (approved):

aykutfarsak:
With invalid expires date value, aws cli returns an error:

```
➜  aws s3 cp file s3://bucket/prefix --expires "2024-13-01T20:30:00Z"
upload failed: ./Makefile to s3://bucket/prefix/file Parameter validation failed:
Invalid type for parameter Expires, value: 2024-13-01T20:30:00Z, type: <class 'str'>, valid types: <class 'datetime.datetime'>, timestamp-string
➜  echo $?
1
```

So I am not sure if we should ignore the invalid expire date error. Maybe we can increase the log level to `error` or return the error immediately.

What do you think @ilkinulas?

ilkinulas:
> With invalid expires date value, aws cli returns an error:
> 
> ```
> ➜  aws s3 cp file s3://bucket/prefix --expires "2024-13-01T20:30:00Z"
> upload failed: ./Makefile to s3://bucket/prefix/file Parameter validation failed:
> Invalid type for parameter Expires, value: 2024-13-01T20:30:00Z, type: <class 'str'>, valid types: <class 'datetime.datetime'>, timestamp-string
> ➜  echo $?
> 1
> ```
> 
> So I am not sure if we should ignore the invalid expire date error. Maybe we can increase the log level to `error` or return the error immediately.
> 
> What do you think @ilkinulas?

Good catch @aykutfarsak 👍  Let's add a validation for `--expires` parameter

tombokombo (commented):
> > With invalid expires date value, aws cli returns an error:
> > ```
> > ➜  aws s3 cp file s3://bucket/prefix --expires "2024-13-01T20:30:00Z"
> > upload failed: ./Makefile to s3://bucket/prefix/file Parameter validation failed:
> > Invalid type for parameter Expires, value: 2024-13-01T20:30:00Z, type: <class 'str'>, valid types: <class 'datetime.datetime'>, timestamp-string
> > ➜  echo $?
> > 1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So I am not sure if we should ignore the invalid expire date error. Maybe we can increase the log level to `error` or return the error immediately.
> > What do you think @ilkinulas?
> 
> Good catch @aykutfarsak  Let's add a validation for `--expires` parameter

Actually, date string validation is already in place, I just choose to print debug message in case of invalid date, so maybe log.Error instead of  log.Debug would by enough?

storage/s3.go
```diff
@@ -345,6 +345,22 @@ func (s *S3) Copy(ctx context.Context, from, to *url.URL, metadata Metadata) err
...
+		input.CacheControl = aws.String(cacheControl)
+	}
+
+	expires := metadata.Expires()
+	if expires != "" {
+		t, err := time.Parse(time.RFC3339, expires)
+		if err == nil {
+			input.Expires = aws.Time(t)
+		} else {
+			msg := log.DebugMessage{Err: err.Error()}
```
ilkinulas, line 359: I was trying invalid time formats to see s3 behavior, here are my findings:
- if we upload a new object s3 ignores invalid `Expires` metadata.
- if we override an existing s3 objects (which has a valid `Expires` metadata) with an invalid expires value, s3 deletes the current object's Expires metadata.

So, it would be better to return the error and do not cp/mv any objects in case of unparsable --expires values.

tombokombo: ok, done

storage/s3.go
```diff
@@ -345,6 +345,21 @@ func (s *S3) Copy(ctx context.Context, from, to *url.URL, metadata Metadata) err
...
+	}
+
+	expires := metadata.Expires()
+	if expires != "" {
+		t, err := time.Parse(time.RFC3339, expires)
+		if err == nil {
+			input.Expires = aws.Time(t)
+		} else {
+			return err
+		}
```
aykutfarsak, line 360:
```suggestion
		if err != nil {
			return err
		}
		input.Expires = aws.Time(t)
```

storage/s3.go
```diff
@@ -492,6 +507,21 @@ func (s *S3) Put(
...
```
aykutfarsak, line 522:
```suggestion
		if err != nil {
			return err
		}
		input.Expires = aws.Time(t)
```

aykutfarsak (approved):

ilkinulas:
@tombokombo  thank you for the contribution.

PR #340, opened 2021-08-12
command: add `sync` command.

### LINKED ISSUE #3: Add "sync" command

ocakhasan:
`Sync` command is implemented. The use case is similar to `aws-cli` and `s3cmd`. 
Usage:
```
s5cmd sync <src> <target>
```
`--delete` flag can be used to delete the object belongs to remote but does not belong to source. 
Sync command works in this way.
1. Get all of the source and destination objects.
2. Determine which ones belongs to only source, which ones only destination and which ones are in common both source and destination in parallel.
3. Only source objects are directly copied. There is no need to check if the object should be overridden.
4. Common objects (both in source and destination) needs to be checked to see if the corresponding destination object should be overridden. 
 `sync` command checks `size` and `modification time` default. If any of those is different, then the destination object should be overridden. However, if `--size-only` flag is set, then `size` becomes the only criteria in override operations. 
5. Only destination objects are deleted if `--delete` flag is set. 

Resolves #3


igungor:
Current plan is to go with another approach:  https://github.com/peak/s5cmd/pull/347.

PR #343, opened 2021-08-17
Support Environment S3_ENDPOINT_URL

### LINKED ISSUE #326: Make `--endpoint-url` also usable as ENV
I don't like always to add this option and also don't like to make a alias.

on linux it would be not the big issue to make a alias, but on windows it sucks... so please let us also use as alternaitve the environment variable.

Dexus:
Make the `S3_ENDPOINT_URL` environment variable usable to omit the --endpoint-url flag.

Closes #326

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,7 +325,16 @@ object storage.
 
     s5cmd --endpoint-url https://storage.googleapis.com ls
 
-will return your GCS buckets.
+or an alternative with environment variable
+
+    S3_ENDPOINT_URL="https://storage.googleapis.com" s5cmd ls
+    
+    # or
+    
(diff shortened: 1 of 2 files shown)


ilkinulas:
Thank you for your contribution.
Could you please update README? There is a [section](https://github.com/peak/s5cmd/blob/master/README.md#google-cloud-storage-support)  where `--endpoint-url` is documented. You can add a sample usage there.

Dexus:
@ilkinulas 
I add the example in the README.

ilkinulas (approved):

Dexus:
fix the "fmt check"

seruman (approved):

joshuarobinson:
Great idea! thanks for this change

PR #347, opened 2021-09-01
command: add `sync` command using `run` method.

### LINKED ISSUE #3: Add "sync" command

ocakhasan:
`Sync` command is implemented. The use case is similar to `aws-cli` and `s3cmd`. 
Usage:
```
s5cmd sync <src> <target>
```
`--delete` flag can be used to delete the object belongs to remote but does not belong to source. 
Sync command works in this way.
1. Get all of the source and destination objects.
2. Determine which ones belongs to only source, which ones only destination and which ones are in common both source and destination in parallel.
3. Only source objects are directly copied. There is no need to check if the object should be overridden.
4. Common objects (both in source and destination) needs to be checked to see if the corresponding destination object should be overridden. 
 `sync` command checks `size` and `modification time` default. If any of those is different, then the destination object should be overridden. However, if `--size-only` flag is set, then `size` becomes the only criteria in override operations. 
5. Only destination objects are deleted if `--delete` flag is set. 

## How does it work

`sync` command generates the commands such as 
```
cp a.txt s3://bucket/a.txt
```
then, it passes this command to run method which accepts a reader. Then `run` command runs generated commands in parallel. 

Updates #3


command/sync.go
```diff
@@ -0,0 +1,495 @@
...
+func (s Sync) PlanRun(ctx context.Context, onlySource, onlyDest []*url.URL, common []*CommonObject,
+	dsturl *url.URL, transferManager transfer.Manager, strategy strategy.Strategy, w *io.PipeWriter) {
+	// only source objects.
+	for _, srcurl := range onlySource {
+		var dsturl_ *url.URL
+		switch {
+		case !srcurl.IsRemote() && dsturl.IsRemote(): // local->remote
+			dsturl_ = transferManager.PrepareRemoteDestination(srcurl, dsturl)
+		case srcurl.IsRemote() && !dsturl.IsRemote(): // remote->local
+			dsturl_, _ = transferManager.PrepareLocalDestination(ctx, srcurl, dsturl)
```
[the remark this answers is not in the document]
ocakhasan: how about we delete  `prepareRemoteDestination` or `prepareLocalDestination` and just pass the `dsturl` which comes from the parameter. The reason for that we pass generated url to `cp` command but `cp` also calculates the destination url and there is an error handling.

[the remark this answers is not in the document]
ocakhasan: I tried that, but I guess we need to handle the remote destination. For the error handling part, I think we can skip url's with errors, what do you think?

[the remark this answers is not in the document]
seruman: I think `transfer` package is not needed. As you pointed above `cp` command should create destination files/directories if destination url passed correctly. It should be enough to calculate destination urls in `sync` command then delegate it to `cp`.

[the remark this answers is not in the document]
sonmezonur: Your suggestion about the transfer package sounds good to me. But I could not understand your second comment. Why not we cannot handle errors for remote URLs? Could you refactor this part as you suggested?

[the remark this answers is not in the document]
ocakhasan: When we did not prepare the destination, `cp` commands generates destination in different manner. It is best explained with an example
Supoose that we sync files from s3 to local  and 
We run the command 
```
s5cmd sync s3://bucket/* folder/
```
one of the files in the s3 is, `s3://bucket/a/file.txt`.
1. if we handle the destination url prepation in sync, then the command becomes
```
cp s3://bucket/a/file.txt folder/a/file.txt
```
2. if we do not handle destination url prepation in sync, (cp handles it), then the command becomes
```
cp s3://bucket/a/file.txt folder/file.txt
```
it does not create the a folder, since specific url path is passed. 

Therefore, some of the tests fail right now, I can fix them, but is this behaviour wanted? What should be the correct implementation? After discussing that, we can handle the destination prepation in `sync` or `cp`. 

For the error handling part, 
`prepareLocalDestination` returns error which indicates the local prepation is successful or not, my suggestion is that, (it only applies we use `transfer` package and handle the destination prepation in `sync` ), if there is an error, we can skip or return that error in the end. What is your suggestion about that?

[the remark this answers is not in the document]
sonmezonur: 👍 Thanks for the clarification.  It makes sense now. I think the first behavior is the expected one. 

```
s5cmd sync s3://bucket/* folder/
```
Should copy the file into ` folder/a/file.txt`
```
cp s3://bucket/a/file.txt folder/a/file.txt
```
Since we expand wildcard(`*`) strings in sync and pass raw URLs to run command, URL preparation should be handled in sync too.  Otherwise, destination/source URLs will not be correct as you pointed out. 


>  For the error handling part,
prepareLocalDestination returns error which indicates the local prepation is successful or not, my suggestion is that, (it only applies we use transfer package and handle the destination prepation in sync ), if there is an error, we can skip or return that error in the end. What is your suggestion about that?

Can we use multierror pattern which is used in other commands? We can log errors and append them to multierror. At the end, we can return error.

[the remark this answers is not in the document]
ocakhasan: Now it appends error at the end.

internal/strategy/customstrategies.go
```diff
@@ -0,0 +1,34 @@
+package strategy
```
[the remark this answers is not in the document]
ocakhasan: moved `strategy/` and `transfer/` to main folder. Removed `internal/` folder.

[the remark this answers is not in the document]
sonmezonur: AFAICS strategy package is only used by `sync.go` and I don't see any other future use case for this package. Moving this codebase (only 34 loc) to sync.go would be a better option. What do you think?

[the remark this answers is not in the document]
ocakhasan: I think current approach is better. Because
1. We can seperate the comparsion to another interface and if we would like to use another comparison function (which includes hashes etc.) we just need a struct which implements `Compare` function.
2. As far as I know, `cp` command has a `shouldOverride` function. We can change it to `strategy` interface, the code would be much more cleaner. So, it can be used in another parts as well. Therefore, I created it as another interface.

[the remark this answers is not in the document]
igungor: It's good practice to move code into its own package if it improves readability and testability. OTOH, I agree with @sonmezonur and @seruman. `transfer` package is not needed. We can move relevant methods into `sync.go`.

For the sync strategy code, it's good it's behind an interface. But the package is not needed. We can again move it under the `command` package with its own file;  may be a `sync_strategy.go`.

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+		},
+		&cli.IntFlag{
+			Name:    "part-size",
+			Aliases: []string{"p"},
+			Value:   defaultPartSize,
+			Usage:   "size of each part transferred between host and remote server, in MiB",
+		},
+		&cli.BoolFlag{
+			Name:  "delete",
+			Usage: "delete objects in destionation but not in source",
```
igungor, line 71:
```suggestion
			Usage: "delete objects in destination but not in source",
```

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+		},
+		Action: func(c *cli.Context) (err error) {
+			defer stat.Collect(c.Command.FullName(), &err)()
+
+			return NewSync(c).Run(c)
+		},
+	}
+}
+
+type CommonObject struct {
```
igungor, line 102: CommonObject holds 2 object. `ObjectPair` would do justice.

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+	}
+
+	sourceObjects, destObjects, err := s.GetSourceAndDestinationObjects(c.Context, srcurl, dsturl)
+	if err != nil {
+		printError(s.fullCommand, s.op, err)
+		return err
+	}
+	onlySource, onlyDest, commonObjects := CompareObjects(sourceObjects, destObjects)
+
+	// clear the arrays.
```
igungor, line 171: code is simple enough that doesn't require comment.

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+		defer close(errDoneCh)
+		for err := range waiter.Err() {
+			if strings.Contains(err.Error(), "too many open files") {
+				fmt.Println(strings.TrimSpace(fdlimitWarning))
+				fmt.Printf("ERROR %v\n", err)
+
+				os.Exit(1)
+			}
+			printError(s.fullCommand, s.op, err)
+			merrorWaiter = multierror.Append(merrorWaiter, err)
```
igungor, line 192: `merrorWaiter` is not used elsewhere. What should we do with it?

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+	}()
+
+	strategy := strategy.New(s.sizeOnly)                                                             // create comparison strategy.
+	transferManager := transfer.NewManager(s.storageOpts, false, isBatch, s.concurrency, s.partSize) // create transfer manager
+	pipeReader, pipeWriter := io.Pipe()                                                              // create a reader, writer pipe to pass commands to run
+
+	var merrorPlan error // for errors in plan
+	// Create commands in background.
+	go func() {
+		s.PlanRun(c.Context, onlySource, onlyDest, commonObjects, dsturl, *transferManager, strategy, pipeWriter, &merrorPlan)
```
igungor, line 203: Why do we pass `merrorPlan` as pointer?

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+		}
+	}()
+
+	strategy := strategy.New(s.sizeOnly)                                                             // create comparison strategy.
+	transferManager := transfer.NewManager(s.storageOpts, false, isBatch, s.concurrency, s.partSize) // create transfer manager
+	pipeReader, pipeWriter := io.Pipe()                                                              // create a reader, writer pipe to pass commands to run
+
+	var merrorPlan error // for errors in plan
+	// Create commands in background.
+	go func() {
```
igungor, line 202: No need to wrap the function with a function. You can call directly as below:

`go s.PlanRun(...)`

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+		s.PlanRun(c.Context, onlySource, onlyDest, commonObjects, dsturl, *transferManager, strategy, pipeWriter, &merrorPlan)
+	}()
+
+	// pass the reader to run
+	return multierror.Append(NewRun(c, pipeReader).Run(c), merrorPlan).ErrorOrNil()
```
igungor, line 207: This line looks a bit compact. Let's remove `NewRun(...).Run()` call outside of this error handling.

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+}
+
+// CompareObjects compares source and destination objects.
+// Returns objects belongs to only source, only destination
+// and common objects.
+// The algorithm is taken from
+// https://github.com/rclone/rclone/blob/HEAD/fs/march/march.go#L304
+func CompareObjects(sourceObjects, destObjects []*storage.Object) (srcOnly, dstOnly []*url.URL, commonObj []*CommonObject) {
```
igungor, line 215: Let's not use named returns.

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+			}
+		}
+
+		command := fmt.Sprintf("cp %v %v\n", curSourceURL, curDestURL)
+		fmt.Fprint(w, command)
+	}
+
+	// for only destination objects.
+	if s.delete { // if delete is set
+		urlpaths := GenerateRemovePath(onlyDest)
```
igungor, line 366: What happens if the `onlyDest` is empty? `rm %v` would be `rm ` and it is gonna fail.

command/sync.go
```diff
@@ -0,0 +1,501 @@
...
+			curDestURL = url
+			if err != nil {
+				printError(s.fullCommand, s.op, err)
+				*errPlan = multierror.Append(*errPlan, err)
+				continue
+			}
+		case srcurl.IsRemote() && dsturl.IsRemote(): // remote->remote
+			curDestURL = transferManager.PrepareRemoteDestination(srcurl, dsturl)
+		default:
+			panic("unexpected src-dst pair")
```
seruman, line 342: I think we should only calculate destination urls in planning phase, `transfer.Manager.PrepareLocalDestination` tries to create local directories hence introduces the error handling. `cp`  would already create files/directories if we pass correct urls to it.

To calculate destination urls, seems like we only need these three bits;
https://github.com/peak/s5cmd/blob/7ba574f1838887a975e7f603c8fdc93b74cfdcb2/transfer/destination.go#L24-L27 https://github.com/peak/s5cmd/blob/7ba574f1838887a975e7f603c8fdc93b74cfdcb2/transfer/destination.go#L17-L19
https://github.com/peak/s5cmd/blob/7ba574f1838887a975e7f603c8fdc93b74cfdcb2/transfer/destination.go#L56-L58



This would also simplify the switch case as we could only check whether destination url is remote or not;
```go
for _, srcurl := range onlySource {
        // ....
    switch {
    case dsturl.IsRemote():
        // calculate remote url
    case !dsturl.IsRemote():
        // calculate local url
    default:
        // ....
    }
}
```

This way `Sync.PlanRun`'s only responsibility would be to create commands for `Run`. One might even add a `dry-run` option by consuming the pipe instead of passing it to `Run` without touching creating any directories.

command/sync.go
ocakhasan: return it at the end

command/sync.go
ocakhasan: deleted it

command/sync.go
ocakhasan: add an if condition, if there are items to delete, then it will call `rm`.

command/sync.go
ocakhasan: Changed as you suggested. Can you please check it from [here](https://github.com/peak/s5cmd/blob/2e00d11b005011d42c3003bdc48c940266fbe25e/command/sync.go#L358-L378)

ocakhasan:
Hi everyone, thank you for the reviews, as you know my school starts tomorrow and I have a part time job now, therefore, I may be late to responding to reviews. I will try to fix them but if I do not have the time, I will let you know it, maybe then you can commit to this PR. 
Have a nice day.

command/sync.go
igungor: Could you write a test for this case please?

command/sync.go
```diff
@@ -0,0 +1,508 @@
...
+		> s5cmd {{.HelpName}} s3://sourcebucket/prefix/* s3://destbucket/
+
+	04. Sync local folder to S3 but delete the files that S3 bucket has but local does not have.
+		> s5cmd {{.HelpName}} --delete folder/ s3://bucket/
+
+	05. Sync S3 bucket to local folder but use size as only comparison criteria.
+		> s5cmd {{.HelpName}} --size-only s3://bucket/* folder/
+		
+	06. Sync S3 bucket to local folder but check the md5 values of files.
+		> s5cmd {{.HelpName}} --checksum s3://bucket/* folder/
```
aykutfarsak, line 49:
```suggestion
```

command/sync.go
igungor: I'm adding the test myself Hasan.

command/sync_strategy.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	srcMod, dstMod := srcObj.ModTime, dstObj.ModTime
+	if !srcMod.After(*dstMod) {
+		stickyErr = errorpkg.ErrObjectIsNewer
+	} else {
+		stickyErr = nil
+	}
+
+	return stickyErr
```
aykutfarsak, line 45:
```suggestion
	if srcObj.Size == dstObj.Size {
		return errorpkg.ErrObjectSizesMatch
	}

	srcMod, dstMod := srcObj.ModTime, dstObj.ModTime
	if !srcMod.After(*dstMod) {
		return errorpkg.ErrObjectIsNewer
	}

	return nil
```

igungor (approved):

seruman (approved):

Issue #354, opened 2021-09-20
custom region doesn't work if bucket is specified


sihara:
custom reason works for bucket listing, but it doesn't work object listing. 
<pre>
$ cat ~/.aws/credentials 
[default]
aws_access_key_id = AKIA991Z6427QRER7O91
aws_secret_access_key = HBwlovGPG2u9bVAcytO+a97GI4m3K7JT1RkrvzNh
region = jp

$ ./s5cmd version 
v1.4.0-rc.1-a0791f8

$ ./s5cmd --endpoint-url http://s3ds-1 ls 
2021/09/20 01:08:01  s3://bucket1

$ ./s5cmd --endpoint-url http://s3ds-1 ls s3://bucket1
ERROR "ls s3://bucket1": AuthorizationHeaderMalformed: The authorization header is malformed; the region 'us-east-1' is wrong; expecting 'jp' status code: 400, request id: 611C446F010F3B13, host id: 611E5582312F22A8

seruman:
Hi @shiara, sorry to hear that you're experiencing this. We're aware of this issue as it's already brought out multiple times; #325, #317, #314. We already have an open issue #320 and a PR #331 to address it. 

Since precedence changes would break usage of auto region detection, we're planning to fix this on next major release v2.0.0, hopefully in the following months.

Closing in favour of #320.

sihara:
@seruman I'm testing s5cmd against DDN S3 Data service (https://www.ddn.com/download/ddn-data-services/) which can be running on DDN's parallel filesystem (EXAScaler/Lustre). S3 Data Service is very scalable and high performance and allows accessing to same name service of EXAScaler/Lustre.
I've done testing of several S3 client utilities (s3cmd, awscli, s3curl, etc) and they have been working well so far. I found s5cmd and it sounds like very fast. So, It's time to test s5cmd and benchmark the performance on DDN S3 Data service. :)
looking forward to getting fix for this region issue in new release sooner!

PR #331, opened 2021-07-29
storage/s3: fix precedence of region detection

seruman:
Previously auto region detection would always override the region defined in SDK config which populated via `AWS_REGION/AWS_DEFAULT_REGION` or region in given profile with `AWS_PROFILE`.

https://github.com/peak/s5cmd/blob/02f63b3798aa5a2424867771308f67cec6fd3071/storage/s3.go#L822-L832

This was causing issues mainly on users of S3 compatible services returning empty response on `HeadBucket` request that's converted into `us-east-1` in `s3manager.GetBucketRegion()`([1](https://github.com/aws/aws-sdk-go/blob/b8fe768e4ce7f8f7c002bd7b27f4f5a8723fb1a5/service/s3/s3manager/bucket_region.go#L146), [2](https://github.com/aws/aws-sdk-go/blob/b8fe768e4ce7f8f7c002bd7b27f4f5a8723fb1a5/service/s3/bucket_location.go#L23-L32)) ; e.g #325, #354.

This PR makes changes to predecense of region detection as below;

1. `--source-region` or `--destination-region` flags of `cp` command.
2. `AWS_REGION` environment variable.
3. Region section of AWS profile.
4. Auto detection from bucket region (via `HeadBucket`).
5. `us-east-1` as default region.

Resolves #325, resolves #320.

Closes #317.

PR #355, opened 2021-09-20
storage: fix region to work properly if bucket is specified

sihara:
This is fix for Issue #354. custom region can be worked, but when if bucket is not specified.
Fix solves problem and user defined region works regardless bucket is specified.


seruman:
Thanks for the PR @sihara. As I've mentioned in https://github.com/peak/s5cmd/issues/354#issuecomment-922952223 there's an open PR https://github.com/peak/s5cmd/pull/331 to address region detection issue, which we both made similar changes apparently 🙂. We're planning to release it in the following months with the new major release.

Closing for #331

PR #361, opened 2021-09-22
.github/workflows: run ci on last the last 3 go versions

arl:
I was just passing by and noticed CI was not running into the latest go version.
So I dropped 1.14 and added 1.17


igungor (approved):
Thanks!

aykutfarsak (approved):

PR #369, opened 2021-09-29
command/sync: update default strategy, update tests

seruman:
Here's a summary of changes;
* Updated default strategy (`SizeAndModificationStrategy`) to match `awscli`'s default strategy ([`SizeAndLastModifiedSync`](https://github.com/aws/aws-cli/blob/ff5961571f534c2e5c5417c3d628d6597d49317c/awscli/customizations/s3/syncstrategy/base.py#L226-L239)).
   * Previously it would only sync if the source file is newer than destination.
* Added tests for existing strategies.
* Updated `sync` tests by mocking time to control remote object's upload time.
* Added option to pass time source to `gofakes3` server.

---
* With Go 1.17, `go fmt` wanted to update build tags;
   *  [Go 1.17 Release Notes - Gofmt](https://golang.org/doc/go1.17#gofmt)


aykutfarsak (changes requested):

command/sync_strategy.go
```diff
@@ -5,42 +5,50 @@ import (
...
+// returns a sentinel error to indicate not to 'sync'. Even if 'srcObj' older than 'dstObj' it returns without error
+// if file sizes would not match.
+//
+// time: src > dst        size: src != dst    should sync: yes
+// time: src > dst        size: src == dst    should sync: yes
+// time: src <= dst       size: src != dst    should sync: yes
+// time: src <= dst       size: src == dst    should sync: no
+type SizeAndModificationStrategy struct{}
+
+func (sm *SizeAndModificationStrategy) Compare(srcObj, dstObj *storage.Object) error {
```
aykutfarsak, line 43: On my first read of the code, I think that if `Compare()` returns `nil`, it means "these files are same, no need to sync". But it works just the opposite. Maybe we could consider renaming it or working the other way around.

seruman: Agree, I've renamed it to `ShouldSync`.

aykutfarsak (approved):
LGTM

README.md
```diff
@@ -269,6 +269,82 @@ mv s3://bucket/2020/03/18/file1.gz s3://bucket/2020/03/18/original/file.gz
 ls # inline comments are OK too
 ```
 
+#### Sync
+With `sync` command, it's possible to synchronise files between an S3 bucket and a local directory;
```
igungor, line 273: I read that as we only support s3->dir sync. Let's elaborate, like "`sync` command synchronizes S3 buckets, prefixes and directories. It also can synchronize given S3 buckets or prefixes.".

README.md
```diff
@@ -269,6 +269,82 @@ mv s3://bucket/2020/03/18/file1.gz s3://bucket/2020/03/18/original/file.gz
...
+
+Suppose we have following files;
```
igungor, line 275: Before diving into how it works, let's make it clear that we never touch anything in the source. We never delete anything from the source. Sync code assumes the source as the truth.

README.md
```diff
@@ -269,6 +269,82 @@ mv s3://bucket/2020/03/18/file1.gz s3://bucket/2020/03/18/original/file.gz
...
+```
+s5cmd sync 's3://bucket/static/*.html' .
+
+cp s3://bucket/prefix/index.html index.html
+cp s3://bucket/prefix/test.html test.html
+```
+
+##### Strategy
+###### Default
+By default `s5cmd` compares files' both size and modification times, treating source files as 'source-of-truth'. Any difference in size or modification time would cause `s5cmd` to copy source object to destination.
```
igungor, line 329:
```suggestion
By default `s5cmd` compares files' both size **and** modification times, treating source files as **source of truth**. Any difference in size or modification time would cause `s5cmd` to copy source object to destination.
```

README.md
```diff
@@ -269,6 +269,82 @@ mv s3://bucket/2020/03/18/file1.gz s3://bucket/2020/03/18/original/file.gz
...
+
+mod time    |  size        |  should sync
+------------|--------------|-------------
+src > dst   |  src != dst  |  ✅
+src > dst   |  src == dst  |  ✅
+src <= dst  |  src != dst  |  ✅
+src <= dst  |  src == dst  |  ❌
+
+###### Size only
+With `--size-only` flag, it's possible to use the strategy that would only compare file sizes. Source treated as 'source-of-truth' and any difference in sizes would cause `s5cmd` to copy source object to destination.
```
igungor, line 339:
```suggestion
With `--size-only` flag, it's possible to use the strategy that would only compare file sizes. Source treated as **source of truth** and any difference in sizes would cause `s5cmd` to copy source object to destination.
```

README.md
seruman: Updated the sentence

README.md
seruman: Tried to make it clear with a little bit of glimpse on how it syncs. How does it sound like?

README.md
```diff
@@ -269,6 +269,88 @@ mv s3://bucket/2020/03/18/file1.gz s3://bucket/2020/03/18/original/file.gz
 ls # inline comments are OK too
 ```
 
+#### Sync
+`sync` command synchronizes S3 buckets, prefixes, directories and files between S3 buckets and prefixes as well.
+It compares files between source and destination, taking source files as **source-of-destination**;
```
igungor, line 274: source-of-truth?

seruman: Updated 🤦

README.md
igungor: Looks better. Thanks.

igungor (approved):

PR #403, opened 2022-01-28
Add --exclude to the sync command

mbrostami:


seruman:
Thanks for the contribution @mbrostami!. 

As #409 adds flags common to both `cp` and `sync` -which includes `--exclude`-. I think we may close this one in favor of #409. Could you please check if it covers your needs?

PR #409, opened 2022-02-21
Add missing sync command flags

### LINKED ISSUE #3: Add "sync" command

### LINKED ISSUE #365: Add missing `sync` command flags
The `cp` command has many options that are missing from the `sync` command.

### LINKED ISSUE #397: sync is not working if file name or directory contain space
Hi,
It seems s5cmd sync can not handle file names with space. sync then generate cp error like such:

ERROR "cp /zfs-pool/00 photo/0829 bla bla/_MG_5332.CR2 s3://aws_bucket/backup/40_photo/0829 bla bla/_MG_5332.CR2": expected source and destination arguments

Would it be possible to contain the name in quotes or something so that the command is not confused with what is the name and path and what are arguments?

thx

### LINKED ISSUE #416: sync does not seem affected by part-size
I have a single 200GB file that I'm trying to upload to my S3 bucket. I'm on a 10Gbps connection so I tried to increase part-size to maximize throughput.

However, sync doesn't seem to react to part-size in the same way as a normal cp. When I set a large enough part-size with cp, I can almost use the entire throughput. The same setting on sync nets me a disappointing 2Gbps. 

This is running on a windows 10 machine with latest from master

sonmezonur:
Resolves #3, resolves #365, fixes #397, fixes #416

Adding missing flags and following changes:

* sync command can use same validation methods with cp. Currently, it does not accept single object sync and other validation logic is exactly the same with cp command. I don't see any reason for not supporting this:

```
sync s3://bucket/source.go .
```

* Updated givenCommand() method to add flags to command string.

* Added e2e tests for `--exclude`, `--raw` and `--no-follow-symlinks` and unit tests for other flags

* Below flags are not tested:

```
sse
sse-kms-key-id
acl
cache-control
expires
force-glacier-transfer
```


e2e/sync_test.go
```diff
@@ -11,8 +12,36 @@ import (
...
+
+	cmd := s5cmd("sync", "-n", srcpath, ".")
+	result := icmd.RunCmd(cmd)
+	result.Assert(t, icmd.Expected{ExitCode: 1})
+
+	// urfave.Cli prints the help text and error message to stdout
+	// if given flags in not present in command options.
+	assertLines(t, result.Stdout(), map[int]compareFunc{
+		0: equals("Incorrect Usage: flag provided but not defined: -n"),
+	}, skipAfterLine(0))
```
igungor, line 40: `strictLineCheck(false)` could be used here.

sonmezonur:
PTAL. All copy flags should work with sync. Added tests

gvreddy579:
👍

igungor (approved):

aykutfarsak (approved):

PR #412, opened 2022-02-23
Dockerfile: use `/aws` as working directory

LadMartin:
Changes relative path to absolute path so a user can easily mount his working dir to the docker
Add empty workdir similar to aws

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,4 +6,5 @@ RUN apk add --no-cache git make && \
 
 FROM alpine:3.13
 COPY --from=build /s5cmd/s5cmd .
-ENTRYPOINT ["./s5cmd"]
+WORKDIR /aws
+ENTRYPOINT ["/s5cmd"]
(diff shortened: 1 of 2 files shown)


LadMartin:
It helps when user wants to upload/download local files from/to his workdir.

Dockerfile
```diff
@@ -6,4 +6,5 @@ RUN apk add --no-cache git make && \
 
 FROM alpine:3.13
 COPY --from=build /s5cmd/s5cmd .
-ENTRYPOINT ["./s5cmd"]
+WORKDIR /aws
```
seruman, line 9: What do think about making it a vendor agnostic name like `src` or something? 

Also it would be good to document the working directory in the Docker section of the README with an example.

LadMartin: My reasoning was to unify the usage since s5cmd also supports some AWS_* env variables. It would be easier to jump from one image to another.

ilkinulas (approved):

seruman (approved):
Thanks for the contribution!

PR #438, opened 2022-04-21
command/run: handle large tokens in scanner

### LINKED ISSUE #435: `sync` fails with a large number of files to remove
`sync` exits with status 1 and no error message when given a situation that results in a large number of deletions:

Run in an empty directory, which will cause around 500k deletions:
```
❯ s5cmd --log=debug --stat sync --delete . s3://my-bucket/lots-of-keys/

Operation	Total	Error	Success
sync		1	1	0

❯ echo $status
1
```

The same behavior also exists when running with the `--dry-run` flag.

Running the same command against an S3 path that results in 1 deletion works normally.

I realize this isn't a lot of information to work with - please let me know if I can provide more help somehow.

Thanks!

### LINKED ISSUE #436: sync: deletes not propagating and comparison different to aws cli
Hi

I'm trying to use this to sync two large areas of buckets, first thing we have noticed that files get transferred but we've not seen any deletes getting propagated when using the --delete flag

when we run the same sync with aws s3 sync we see lots more files being transferred and all the deletes being run 

Do we know what the difference in comparison are?

sonmezonur:
Fixes #435, fixes #436

Sync uses `run` interface to execute commands. Run scans given `io.Reader` and creates a command from each line. If sync finds multiple files to remove, generated command string exceeds the maximum token size of bufio.Scanner and we are getting `bufio.Scanner: token too long` errors. As described in the issues, `sync` fails if it tries to remove a large number of files. 

[Official go documentation](https://pkg.go.dev/bufio) recommends using `bufio.Reader` for handling large tokens: 

```
Programs that need more control over error handling or large tokens, or must run sequential scans 
on a reader, should use bufio.Reader instead.
```

This PR fixes two bugs:

1. Currently, scanner errors are ignored, and commands silently fail if we encounter scanner errors. Always print scanner errors before exiting the command.

2. Use `bufio.Reader` instead of `bufio.Scanner` to handle large tokens.

### DIFF
--- a/command/run.go
+++ b/command/run.go
@@ -181,25 +181,6 @@ func NewScanner(ctx context.Context, r io.Reader) *Scanner {
 	return scanner
 }
 
-func (s *Scanner) readLine() (string, error) {
-	var (
-		isPrefix = true
-		err      error
-		line, ln []byte
-	)
-
-	// If the line was too long for the buffer then isPrefix is set and
-	// the beginning of the line is returned. The rest of the line will be
-	// returned from future calls. isPrefix will be false when returning
-	// the last fragment of the line.
-	for isPrefix && err == nil {
-		line, isPrefix, err = s.Reader.ReadLine()
-		ln = append(ln, line...)
-	}
-
-	return string(ln), err
-}
-
 // scan read the underlying reader.
 func (s *Scanner) scan() {
 	defer close(s.linech)
@@ -210,16 +191,18 @@ func (s *Scanner) scan() {
 			s.err = s.ctx.Err()
 			return
 		default:
-			line, err := s.readLine()
+			// If ReadString encounters an error before finding a delimiter,
+			// it returns the data read before the error and the error itself (often io.EOF).
+			line, err := s.ReadString('\n')
+			if line != "" {
+				s.linech <- strings.TrimSpace(line)
+			}
 			if err != nil {
 				if err == io.EOF {
 					return
 				}
 				s.err = multierror.Append(s.err, err)
-				continue
 			}
-
-			s.linech <- line
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


lozbrown:
please can this be reviewed/merged and a beta 2 created there is a known bug in the current beta...

command/run.go
```diff
@@ -167,15 +172,34 @@ type Scanner struct {
...
+		err      error
+		line, ln []byte
+	)
+
+	// If the line was too long for the buffer then isPrefix is set and
+	// the beginning of the line is returned. The rest of the line will be
+	// returned from future calls. isPrefix will be false when returning
+	// the last fragment of the line.
+	for isPrefix && err == nil {
+		line, isPrefix, err = s.Reader.ReadLine()
```
igungor, line 196: According to the [documentation](https://pkg.go.dev/bufio#Reader.ReadLine);

> ReadLine is a low-level line-reading primitive. Most callers should use ReadBytes('\n') or ReadString('\n') instead or use a Scanner.

Why do you prefer `ReadLine` over the recommended methods?

command/run.go
```diff
@@ -153,60 +153,71 @@ func (r Run) Run(ctx context.Context) error {
...
+func (r *Reader) Read() <-chan string {
+	return r.linech
 }
 
 // Err returns encountered errors, if any.
-func (s *Scanner) Err() error {
-	if s.err != nil {
-		return s.err
+func (r *Reader) Err() error {
+	if errors.Is(r.err, context.Canceled) {
```
igungor, line 216: I think we can return `r.err` directly without checking the type of the error.

command/run.go
igungor: Thanks.

command/run.go
sonmezonur: done

igungor (approved):

aykutfarsak (approved):

sonmezonur:
@lozbrown New release is now available. Let us know if you still encounter the same issue.

PR #453, opened 2022-06-29
storage: add profile flag

### LINKED ISSUE #353: [Feature request] Support for "--profile" global cmdline option
Could you please implement the `--profile` global cmdline option to allow switching between different AWS profiles and which should take precedence over `AWS_PROFILE` env var?
In other words, an option to mimic AWS CLI behavior: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html

This would increase quality of user experience by having same option to switch between profiles in CLI.

Thanks in advance.

kucukaslan:
This commit adds `--profile` flag to allow users to specify a [named profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) rather than using the default.

Fixes #353

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## not released yet
 
+#### Features
+- Added `--profile` flag to allow users to specify a [named profile]. ([#353](https://github.com/peak/s5cmd/issues/353))
+
 
 ## v2.0.0 - 4 Jul 2022
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Features
 - Added `--profile` flag to allow users to specify a [named profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html). ([#353](https://github.com/peak/s5cmd/issues/353))
-- Added `--credential-file` flag to allow users to specify path for the AWS credentials file instead of using the [default location](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-where). 
+- Added `--credentials-file` flag (with `--cf` alias) to allow users to specify path for the AWS credentials file instead of using the [default location](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-where). 
 
 
 ## v2.0.0 - 4 Jul 2022
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Features
 - Added `--profile` flag to allow users to specify a [named profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html). ([#353](https://github.com/peak/s5cmd/issues/353))
-- Added `--credentials-file` flag (with `--cf` alias) to allow users to specify path for the AWS credentials file instead of using the [default location](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-where). 
+- Added `--credentials-file` flag to allow users to specify path for the AWS credentials file instead of using the [default location](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-where). 
 
 #### Improvements
 - Disable AWS SDK logger if log level is not "trace"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 #### Features
 - Added `--profile` flag to allow users to specify a [named profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html). ([#353](https://github.com/peak/s5cmd/issues/353))
+- Added `--credential-file` flag to allow users to specify path for the AWS credentials file instead of using the [default location](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-where). 
 
 
 ## v2.0.0 - 4 Jul 2022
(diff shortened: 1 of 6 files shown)


command/app.go
```diff
@@ -96,6 +100,10 @@ var app = &cli.App{
 			err := fmt.Errorf("retry count cannot be a negative value")
 			printError(commandFromContext(c), c.Command.Name, err)
 			return err
+		} else if c.Bool("no-verify-ssl") && c.String("profile") != "" {
```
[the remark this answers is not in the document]
MuhammedCanKucukaslan: Yes, I meant `no-sign-request`, thanks for warning! I'm fixing it.

ilkinulas (approved):

command/app.go
```diff
@@ -81,6 +81,10 @@ var app = &cli.App{
 			Name:  "request-payer",
 			Usage: "who pays for request (access requester pays buckets)",
 		},
+		&cli.StringFlag{
+			Name:  "profile",
+			Usage: "use the specified `ProfileName` from the credential file",
```
igungor, line 86:
```suggestion
			Usage: "use the specified profile from the credential file",
```

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## not released yet
 
+#### Features
+- Added `--profile` flag to allow users to specify a [named profile]. ([#353](https://github.com/peak/s5cmd/issues/353))
```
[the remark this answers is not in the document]
MuhammedCanKucukaslan: Thanks a lot, nice catch!

storage/s3_test.go
```diff
@@ -128,6 +128,24 @@ func TestNewSessionWithNoSignRequest(t *testing.T) {
 	}
 }
 
+func TestNewSessionWithProfile(t *testing.T) {
+	globalSessionCache.clear()
+	profileName := "default"
```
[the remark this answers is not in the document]
MuhammedCanKucukaslan: Thanks for warning, but to be honest I don't know how can we actually test different profiles.

profiles are read from the actual `.aws/credentials` file of the user. So the tests should not modify it.  
1) We cannot add (and remove)  temporary profiles.
    > We might not have a write access to file.
    >> Even if we have write access and add a temporary profile,  the profile name we've chosen might collide with existing profiles and we may accidentally delete actual profiles.
    >>> To avoid this we can check if there is such a profile with same name before adding temporary profile but "being able to check if there is such a profile (and use if there)" is the very thing this test was. supposed to do.
To me, it seemed as a vicious cycle and meaningless.
2) There is a high probability that  `.aws/credentials` file has a profile named `default`. So trying to obtain "default" profile it from both through our `storage.Options` and aws's `NewSharedCredentials()` seemed to a way to check if our function can actually do the same thing thing with aws.
> ps. If there is no profile with the given profileName (in this. case "default") both our code and the aws's methods give the same error.
 

ps.  by `user` I mean the person who will run this test, not necessarily the end user.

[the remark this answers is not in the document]
MuhammedCanKucukaslan: Just to cover all the cases, AWS SDK allows specifying the credential file. So if we add this feature to `s5cmd` then we can create a [temporary file](https://pkg.go.dev/io/ioutil#TempFile) and use it to write tests that makes more senses, and also cover more cases: e.g. tests for default profile, existing profiles, non-existing profiles.

command/app.go
igungor: 👍

aykutfarsak:
@Kucukaslan could you add `--credential-file` changes to readme and changelog, please?

storage/s3_test.go
```diff
@@ -128,6 +128,97 @@ func TestNewSessionWithNoSignRequest(t *testing.T) {
...
+			expSecretAccessKey: "p2_profile_access_key",
+		},
+		{
+			fileName:           file.Name(),
+			profileName:        "p3",
+			expAccessKeyId:     "p3_profile_key_id",
+			expSecretAccessKey: "p3_profile_access_key",
+		},
+	}
+	for _, tc := range testcases {
```
ilkinulas, line 197: Could you please add a descriptive name for all the test cases and run each test with `t.Run` as in https://github.com/peak/s5cmd/blob/master/storage/s3_test.go#L78 ?

Kucukaslan:
The tests for _macos_ are canceled with `GitHub Actions has encountered an internal error when running your job.`  [error](https://github.com/peak/s5cmd/actions/runs/2691284993). Might be related to [this incident](https://www.githubstatus.com/incidents/437q52fb70hg).

command/app.go
```diff
@@ -81,6 +81,15 @@ var app = &cli.App{
 			Name:  "request-payer",
 			Usage: "who pays for request (access requester pays buckets)",
 		},
+		&cli.StringFlag{
+			Name:  "profile",
+			Usage: "use the specified profile from the credentials file",
+		},
+		&cli.StringFlag{
+			Name:    "credentials-file",
+			Aliases: []string{"cf"},
```
igungor, line 90: I don't think we need an alias for this flag :)

Kucukaslan: Flag name is so long so I think it would be nice to have a short alias. But since it is not supposed to be a common flag, I agree it's a bit unnecessary.

I'll remove. Thanks!

aykutfarsak (approved):

seruman (approved):

igungor (approved):

yermulnik:
👍🏻 Thank you
What's approx ETA on releasing this?

igungor:
@yermulnik `v2.1.0` is released with `--profile` support.

yermulnik:
@igungor Could you please take a look at https://github.com/peak/s5cmd/issues/553#issuecomment-1596938700 + https://github.com/peak/s5cmd/issues/571 ? Thanks

PR #459, opened 2022-07-05
storage/url: Allow usage of adjacent slashes

### LINKED ISSUE #352: s5cmd drops leading slashes in keys
Version: `v1.3.0-bb1775a`

Steps to reproduce:
```sh
$ touch file.txt
$ s5cmd cat s3://bucket/file.txt
ERROR "cat s3://bucket/file.txt": NoSuchKey: The specified key does not exist. status code: 404, request id: Z...BMQB76, host id: Yy5Y9JW0D9Nvx....NMdpe9/dbE=
$ s5cmd cp file.txt s3://bucket//file.txt
cp file.txt s3://bucket/file.txt
$ s5cmd cat s3://bucket/file.txt
$
```

You can compare this behavior to the aws CLI:
```sh
$ aws s3 cp file.txt s3://bucket//file.txt
upload: ./file.txt to s3://bucket//filet.txt
$ aws s3 cp s3://bucket//file.txt .
download: s3://bucket//file.txt to ./file.txt
```

Even more problematic, if there's an existing key with a leading slash, s5cmd won't be able to address it.

### LINKED ISSUE #449: storage: unknown url format
works:
`aws s3 ls s3://foo/bar/s3://baz`
fails:
`s5cmd ls s3://foo/bar/s3://baz`
storage: unknown url format

(keys like these should be legal)

boraberke:
This PR allows url with adjacent slashes to be handled correctly. 

- Disabled Automatic URI cleaning ([search in this link for more information](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/)) as it do not allow adjacent slashes.
- Changed path.Join method to strings.Join when joined remote urls as path.Join also cleans paths which removes additional slashes.
-  Kept local urls to be joined with path.Join.
- Added test cases for copy command for following cases:
    - Remote to local
    - Local to remote
    - Remote to remote
- Removed redundant leading slashes in the keys of s3 objects in ls_test.go and sync_test.go files.

Users should be careful while handling slashes in objectnames. Before this PR,`s5cmd cp /file.txt s3://bucket/` would output
```
cp /file.txt s3://bucket/file.txt
```
But now it will result in double slashes as follows:
```
cp /file.txt s3://bucket//file.txt
```

Fixes: https://github.com/peak/s5cmd/issues/352
Fixes: https://github.com/peak/s5cmd/issues/449


storage/url/url.go
```diff
@@ -185,8 +193,14 @@ func (u *URL) Join(s string) *URL {
 	}
 
 	clone := u.Clone()
-	clone.Path = path.Join(clone.Path, s)
-
+	if !clone.IsRemote() {
+		//destination is not remote, thus clean the path by using path.Join which removes multiple slashes
```
seruman, line 197:
```suggestion
		// URL is local, thus clean the path by using path.Join which
		// removes adjacent slashes.
```

storage/url/url.go
```diff
@@ -185,8 +193,14 @@ func (u *URL) Join(s string) *URL {
...
+		clone.Path = path.Join(clone.Path, s)
```
seruman, line 198: `else` block could be eliminated by returning `clone` here early.

storage/s3.go
```diff
@@ -794,14 +794,15 @@ func (sc *SessionCache) newSession(ctx context.Context, opts Options) (*session.
 	if opts.NoVerifySSL {
 		httpClient = insecureHTTPClient
 	}
-
+	// WithDisableRestProtocolURICleaning is added to allow adjacent slashes to be used in s3 object keys.
```
seruman, line 797: Could be moved right above the `WithDisableRestProtocolURICleaning` call.
```suggestion
	// Disable URI cleaning to allow adjacent slashes to be used in S3 object keys.
```

igungor (approved):

igungor (changes requested):
@boraberke could you update the changelog file to reflect this change please? I'm not sure if this is a bugfix or an improvement :)

seruman:
> @boraberke could you update the changelog file to reflect this change please? I'm not sure if this is a bugfix or an improvement :)

Might even breaking for users those rely on URI cleaning 🙃 .

boraberke:
> @boraberke could you update the changelog file to reflect this change please? I'm not sure if this is a bugfix or an improvement :)

Added three more entries to `CHANGELOG.MD` further explaining the details of this PR :)

seruman (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## not released yet
 
+#### Breaking changes
+- Leading slashes in `objectname` are not trimmed anymore when uploading to remote. Before `s5cmd cp /file.txt s3://bucket/` would output `cp /file.txt s3://bucket/file.txt` but now it will output `cp /file.txt s3://bucket//file.txt`. ([#459](https://github.com/peak/s5cmd/pull/459))
```
igungor, line 6: > Before `s5cmd cp /file.txt s3://bucket/` would output `cp /file.txt s3://bucket/file.txt` but now it will output `cp /file.txt s3://bucket//file.txt`

Hmm, this looks wrong to me. `/file.txt` is a file at `/`. Why do we treat it like it has a `/` in its name?

boraberke: > > Before `s5cmd cp /file.txt s3://bucket/` would output `cp /file.txt s3://bucket/file.txt` but now it will output `cp /file.txt s3://bucket//file.txt`
> 
> Hmm, this looks wrong to me. `/file.txt` is a file at `/`. Why do we treat it like it has a `/` in its name?

Sorry, my mistake, it doesn't treat it like that. I was confused by the test files where we use the `putFile` function. However that function is test purposes only and not related to s5cmd's behavior. 

I will remove this part from changelog.

igungor: Thanks.

igungor (approved):

PR #460, opened 2022-07-05
disable aws sdk logger if log level is not "trace"

kucukaslan:
Currently, we request aws-sdk's logs regardless of s5cmd's log level, then we either use or discard those logs depending on the log level.

This commit disables aws-sdk logs (unless log level is "trace") instead of discarding the logs after they created.

So, it will reduce the memory/CPU usage when other log levels used, since those unneeded logs won't be created at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## not released yet
 
+#### Improvements
+- disable aws sdk logger if log level is not "trace"
 
 ## v2.0.0 - 4 Jul 2022
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## not released yet
 
 #### Improvements
-- disable aws sdk logger if log level is not "trace"
+- Disable AWS SDK logger if log level is not "trace"
 
 ## v2.0.0 - 4 Jul 2022
 
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -801,7 +801,7 @@ func (sc *SessionCache) newSession(ctx context.Context, opts Options) (*session.
 		WithS3UseAccelerate(useAccelerate).
 		WithHTTPClient(httpClient)
 
-	if opts.LogLevel == "trace" {
+	if log.LevelFromString(opts.LogLevel) != log.LevelTrace {
 		awsCfg = awsCfg.WithLogLevel(aws.LogDebug).
 			WithLogger(sdkLogger{})
 	}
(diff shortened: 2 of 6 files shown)


ilkinulas (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## not released yet
 
+#### Improvements
+- disable aws sdk logger if log level is not "trace"
```
igungor, line 6:
```suggestion
- Disable AWS SDK logger if log level is not "trace"
```

storage/s3.go
```diff
@@ -799,9 +799,12 @@ func (sc *SessionCache) newSession(ctx context.Context, opts Options) (*session.
 		WithEndpoint(endpointURL.String()).
 		WithS3ForcePathStyle(!isVirtualHostStyle).
 		WithS3UseAccelerate(useAccelerate).
-		WithHTTPClient(httpClient).
-		WithLogLevel(aws.LogDebug).
-		WithLogger(sdkLogger{})
+		WithHTTPClient(httpClient)
+
+	if opts.LogLevel == "trace" {
```
igungor, line 804: We have log levels in `log` package. It should be safe to export them and use it here.

aykutfarsak (approved):

CHANGELOG.md
MuhammedCanKucukaslan: Thanks, updated changelog!

storage/s3.go
MuhammedCanKucukaslan: I've exported the LogLevel types and changed this line to
`log.LevelFromString(opts.LogLevel) != log.LevelTrace `.

I've 2 questions:
1) Should I add a bullet point saying "LogLevel type is exported..." to Changelog?

2) Should type of storage.Options.LogLevel remain as string or should I change it to LogLevel too.
    > ps. all other options are set with c.String, c.Bool and c.Int methods in  [NewStorageOpts](https://github.com/peak/s5cmd/blob/d6f9d7af625ca1dd8f0941704d09e1a744c136fc/command/app.go#L156-#L167).
If we change its type we this would be set with `log.LevelFromString(c.String("log"))`, different than other option parameters.

igungor: 1. no need as s5cmd doesn't provide an API.
2. it'd better to use the type.

storage/s3.go
```diff
@@ -799,9 +799,12 @@ func (sc *SessionCache) newSession(ctx context.Context, opts Options) (*session.
 		WithEndpoint(endpointURL.String()).
 		WithS3ForcePathStyle(!isVirtualHostStyle).
 		WithS3UseAccelerate(useAccelerate).
-		WithHTTPClient(httpClient).
-		WithLogLevel(aws.LogDebug).
-		WithLogger(sdkLogger{})
+		WithHTTPClient(httpClient)
+
+	if log.LevelFromString(opts.LogLevel) != log.LevelTrace {
```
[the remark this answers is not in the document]
igungor: 👍

igungor (approved):

PR #463, opened 2022-07-07
Change cp error not being descriptive

### LINKED ISSUE #461: error is not descriptive when cp can't find the given object
When copy command is used to upload every file under current directory, it doesn’t print a descriptive error log if it fails to find an object.

### Steps to reproduce:

Create a directory and some files where cp command will not be able to find a file. I used symlink and removed the object that it links to. 

```
mkdir temp
cd temp
touch test1.txt test2.txt test3.txt
ln -s test1.txt symlink
rm test1.txt
s5cmd mb s3://bucket
s5cmd cp . s3://bucket
```

It gives:

```
ERROR "cp . s3://bucket": given object not found
```
This error message does not specify which object is missing. If similar error is created in aws-cli, it results to this warning:

```
Skipping file /Users/username/temp/symlink. File does not exist.
```

boraberke:
With this fix, when "given object not found" error message will also include absolute path of the file.

Fixes https://github.com/peak/s5cmd/issues/461

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -752,8 +752,8 @@ func prepareLocalDestination(
 	obj, err := client.Stat(ctx, dsturl)
 
 	if err != nil {
-		var givenObjNotFoundErr *storage.ErrGivenObjectNotFound
-		if !errors.As(err, &givenObjNotFoundErr) {
+		var objNotFound *storage.ErrGivenObjectNotFound
+		if !errors.As(err, &objNotFound) {
 			return nil, err
 		}
 	}
@@ -765,8 +765,8 @@ func prepareLocalDestination(
 			return nil, err
 		}
 	}
-	var givenObjNotFoundErr *storage.ErrGivenObjectNotFound
-	if errors.As(err, &givenObjNotFoundErr) {
+	var objNotFound *storage.ErrGivenObjectNotFound
+	if errors.As(err, &objNotFound) {
 		err := client.MkdirAll(dsturl.Dir())
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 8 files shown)


sonmezonur (approved):

sonmezonur:
@boraberke Thanks, LGTM! Could you describe this change in CHANGELOG?

storage/storage.go
```diff
@@ -13,13 +13,20 @@ import (
...
 	// ErrNoObjectFound indicates there are no objects found from a given directory.
 	ErrNoObjectFound = fmt.Errorf("no object found")
 )
 
+// ErrGivenObjectNotFound indicates a specified object is not found.
+type ErrGivenObjectNotFound struct {
+	ObjectAbsPath string
+}
+
+func (e *ErrGivenObjectNotFound) Error() string {
```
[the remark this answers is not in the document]
boraberke: Yes, we may or may not need the receiver to be pointer. I used pointer receiver as the implementation of [`errors.New()`](https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/errors/errors.go;l=58) in errors package is similarly with a pointer receiver. We can still change it if you think that's cleaner 👍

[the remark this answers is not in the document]
igungor: 👍

e2e/cp_test.go
```diff
@@ -2802,6 +2802,37 @@ func TestCopyWithFollowSymlink(t *testing.T) {
...
+	t.Parallel()
+
+	s3client, s5cmd, cleanup := setup(t)
+	defer cleanup()
+
+	const bucket = "bucket"
+	createBucket(t, s3client, bucket)
+
+	folderLayout := []fs.PathOp{
+		//we intentionally did not create a/f1.txt to
```
igungor, line 2815: could you add a whitespace after each `//` line please?

command/cp.go
```diff
@@ -749,8 +750,12 @@ func prepareLocalDestination(
 	}
 
 	obj, err := client.Stat(ctx, dsturl)
-	if err != nil && err != storage.ErrGivenObjectNotFound {
-		return nil, err
+
+	if err != nil {
+		var givenObjNotFoundErr *storage.ErrGivenObjectNotFound
```
igungor, line 755: Variable name is too long for me :) Let's rename them as `objNotFound`

boraberke: Fixed 👍

igungor: Thanks.

igungor (approved):

PR #466, opened 2022-07-13
Add content-encoding flag for header

tghaas:

PR #467, opened 2022-07-15
Make it possible to override content-type and content-encoding from CLI arguments

### LINKED ISSUE #264: How to add custom metadata e.g "cache-control"?
Official AWS CLI does support adding metadata such as `--content-type` and `--cache-control`. 

```
$ aws s3 sync ./mydir/ s3://example-bucket/mydir/ \
--content-type 'application/rss+xml; charset=UTF-8' \
--cache-control 'max-age=86400'
```

The reason I need custom metadata is because I use S3 as static website hosting which sits behind Cloudfront. Setting correct origin headers such as `cache-control` and `content-type` for CDN/Proxy is important to my case.

How do I add custom metadata when uploading object using s5cmd? It seems I did not find the options.

nattofriends:
This PR lets users override Content-Type and Content-Encoding headers, to close #264. #318 is a similar PR which added the --cache-control and --expires options.

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -561,7 +561,6 @@ func (c Copy) doUpload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) er
 	}
 
 	metadata := storage.NewMetadata().
-		SetContentType(guessContentType(file)).
 		SetStorageClass(string(c.storageClass)).
 		SetSSE(c.encryptionMethod).
 		SetSSEKeyID(c.encryptionKeyID).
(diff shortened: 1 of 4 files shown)


nattofriends:
Oops, didn't notice #466 exists too

command/cp.go
```diff
@@ -557,6 +569,13 @@ func (c Copy) doUpload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) er
 		SetCacheControl(c.cacheControl).
 		SetExpires(c.expires)
 
+	if c.contentType != "" {
+		metadata.SetContentType(c.contentType)
```
sonmezonur, line 573: If `content-type` is given, there is no need to calculate/guess content-type. Could you please remove line [564](https://github.com/peak/s5cmd/blob/330c26cac176bbaad5690ed5c239b04a17b0f345/command/cp.go#L564) and call `guessContentType` only if content-type flag is empty.

aykutfarsak (approved):

sonmezonur:
@nattofriends Thanks for your contribution. Could you please update CHANGELOG?

sonmezonur:
Also, adding examples to cp help text would be useful.

sonmezonur (approved):

sonmezonur:
🚀

Issue #450, opened 2022-06-27
NoSuchUpload: The specified upload does not exist


eminugurkenar:
Recently we started to see the error below , couldn't find anything that retries on this case.  It basically fails after getting this error.

`ERROR \"cp data s3://data": MultipartUpload: upload multipart failed upload id: IDXXXXXXXXXX caused by: NoSuchUpload: The specified upload does not exist. The upload ID may be invalid, or the upload may have been aborted or completed. status code: 404, request id: PJXXXXX, host id: HOSTIDXX`

PR #469, opened 2022-07-20
retry upload from scratch if fails

eminugurkenar:
solves https://github.com/peak/s5cmd/issues/450 , we can retry from scratch if sdk fails.


ilkinulas:
PR #470 also tries to handle NoSuchUploadError, with a specific hidden flag just for this case.

PR #470, opened 2022-07-21
command: retry upload on s3.NoSuchUpload

### LINKED ISSUE #450: NoSuchUpload: The specified upload does not exist

kucukaslan:
With this change, doUpload method will retry the multipart upload on NoSuchUploadError if the `no-such-upload-retry-count` flag is used with `cp` or `sync` and the value of the special metadata in remote file does not match the one placed in upload request (or does not exists at all) . Otherwise (if it is different), it will assume that upload was successful and ignore the error.

Retry logic is placed into `s3.Put` since it is possible to restart upload operation and write unit test there.

For the retry condition, it will use a user defined metadata field. If the  `no-such-upload-retry-count` given (with positive parameter), then it will put a random string as a [user defined metadata](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata). Then, upon the `s3.NoSuchUpload` error, it will check if that field matches with the one placed or not. If it does not match (or does not even exists), then the upload will be retried.

Fixes #450


Kucukaslan:
A few updates:
I've naively thought that we could, somehow, embed this retry logic into the storage.customRetryer. It would be very nice to able to do it, since it would also enable us to write a unit test*. However after a few (tbh. a lot) attempt, I've realized that even it wouldn't work. The customRetryer retries the failed request. So if we retry it there the same [CompleteMultipartUpload](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html) (?) will be retried with the same nonexistent UploadId. However, what we want is to repeat the MultipartUpload operation for that file, so the solution is to restart the all of MultipartUpload operations.

On the other hand if we leave the retry logic in its current place ((Copy).doUpload) then it would be too high level to manipulate the mockAPI handlers/response to simulate the error for tests.

* Even better simply add another testcase to [storage.TestS3Retry](https://github.com/peak/s5cmd/blob/6b87fb4c7ea9c9cd92d7401ca9fa6740096d59d1/storage/s3_test.go#L390).

Kucukaslan:
Yet another update:
I think it is ready for review.

I've settled to place the retry logic into `s3.Put` which both allows to restart upload operation and to write unit test.
For the retry condition, it will use a special metada field as suggested. If the  `retry-on-no-such-upload` given (with positive parameter), then it will put a random string as a [user defined metada
](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#UserMetadata). Then, upon the `s3.NoSuchUpload` error, it will check if that field matches with the one placed or not. If it does not match (or does not even exists), then the upload will be retried.

ps. I. updated the first comment accordingly.

command/app.go
```diff
@@ -173,16 +173,17 @@ var app = &cli.App{
...
+		Endpoint:                 c.String("endpoint-url"),
+		MaxRetries:               c.Int("retry-count"),
+		NoSignRequest:            c.Bool("no-sign-request"),
+		NoVerifySSL:              c.Bool("no-verify-ssl"),
+		RequestPayer:             c.String("request-payer"),
+		UseListObjectsV1:         c.Bool("use-list-objects-v1"),
+		Profile:                  c.String("profile"),
+		CredentialFile:           c.String("credentials-file"),
+		LogLevel:                 log.LevelFromString(c.String("log")),
+		RetryOnNoSuchUploadError: c.Int("retry-on-no-such-upload"),
```
aykutfarsak, line 186: In my first read of the flag name, it looks like a bool flag to me. Maybe we can change the flag name to `retries-on-no-such-upload`

storage/s3.go
```diff
@@ -554,11 +581,55 @@ func (s *S3) Put(
 		input.ContentEncoding = aws.String(contentEncoding)
 	}
 
+	// add retry code to the object metadata
+	if s.retryOnNoSuchUploadError > 0 {
+		objectMetaData := make(map[string]*string)
+		num, _ := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
+		retryCode = num.String()
+		objectMetaData[retryMetaDataKey] = &retryCode
+		input.Metadata = objectMetaData
```
aykutfarsak, line 590: If we use other metadata in the future, this line might overwrite it. So it would be nice if we set the default value to `input.Metadata` in the beginning and then set the retry code.

command/cp.go
```diff
@@ -177,6 +177,12 @@ func NewSharedFlags() []cli.Flag {
 			Name:  "content-encoding",
 			Usage: "set content encoding for target: defines content encoding header for object, e.g. --content-encoding gzip",
 		},
+		&cli.IntFlag{
+			Name:        "retry-on-no-such-upload",
+			Usage:       "if this flag is given, the multipart upload operation will be retried on NoSuchUpload error. You should not use this, if you're not informally informed to use it.",
```
aykutfarsak, line 182: If we change the flag name, we should change the description also.

storage/s3.go
```diff
@@ -554,11 +581,55 @@ func (s *S3) Put(
...
+	}
+
 	_, err := s.uploader.UploadWithContext(ctx, input, func(u *s3manager.Uploader) {
 		u.PartSize = partSize
 		u.Concurrency = concurrency
 	})
 
+	i := 0
```
aykutfarsak, line 598: We should change the iteration variable name if used after the iteration ends for more readable code. `attempts` maybe?

storage/s3.go
```diff
@@ -554,11 +581,55 @@ func (s *S3) Put(
...
+	for ; i < s.retryOnNoSuchUploadError && err != nil; i++ {
+		// check if the error is NoSuchUpload error
+		if strings.Contains(err.Error(), s3.ErrCodeNoSuchUpload) {
```
aykutfarsak, line 601: Could you change the condition order? 
```
		if !strings.Contains(err.Error(), s3.ErrCodeNoSuchUpload) {
			break
		}
		
		// check if object exists in the destination
		obj, sErr := s.Stat(ctx, to)
		....
```

storage/s3.go
```diff
@@ -554,11 +581,55 @@ func (s *S3) Put(
...
+					err = nil
+					break
+				}
+			}
+			msg := log.ErrorMessage{Err: fmt.Sprintf("Retrying to upload %v upon error: %q", to, err.Error())}
+			log.Error(msg)
+
+			// generate new retry code for this upload attempt
+			num, _ := rand.Int(rand.Reader, big.NewInt(math.MaxInt64))
+			retryCode = num.String()
```
aykutfarsak, line 617: Maybe we can move the retry code generation part to a method.

command/cp.go
aykutfarsak:
```
You should not use this unless you really know what you're doing.
```

aykutfarsak (approved):

storage/s3.go
```diff
@@ -554,14 +581,58 @@ func (s *S3) Put(
...
+			u.Concurrency = concurrency
+		})
+	}
+	if err != nil && strings.Contains(err.Error(), s3.ErrCodeNoSuchUpload) && s.noSuchUploadRetryCount > 0 {
+		err = awserr.New(s3.ErrCodeNoSuchUpload, fmt.Sprintf("RetryOnNoSuchUpload: %v attempts to retry resulted in %v", attempts, s3.ErrCodeNoSuchUpload), err)
+		log.Error(log.ErrorMessage{Err: err.Error()})
+	}
+
 	return err
 }
```
seruman, line 628: Instead of checking then breaking loop depending on error, we might check error and retry count first then return early if these conditions are not met. It would make it more readable IMHO.

storage/s3_test.go
```diff
@@ -565,6 +566,87 @@ func TestS3Retry(t *testing.T) {
...
+			})
+			mockApi.Handlers.Unmarshal.PushBack(func(r *request.Request) {
+				atomic.AddInt32(atomicCounter, 1)
+			})
+
+			f, err := os.CreateTemp("", "")
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer os.Remove(f.Name())
```
seruman, line 636: You can use `gotest.tools/v3/fs` to create files/directories those cleaned up automatically on test exit. It is already used in `e2e` tests you may refere there for examples.

storage/s3_test.go
```diff
@@ -565,6 +566,87 @@ func TestS3Retry(t *testing.T) {
...
+
+			mockS3.Put(ctx, f, url, NewMetadata(), s3manager.DefaultUploadConcurrency, s3manager.DefaultUploadPartSize)
+
+			// +1 is for the original request
+			// *2 is to account for the "Stat" requests that are made to obtain
+			// last modification date( retry decision are made according to mod. date)
+			if *atomicCounter != 1+2*tc.retryCount {
+				t.Errorf("expected retry %d, got %d", tc.retryCount, *atomicCounter)
+			}
```
seruman, line 645: By defining it with the operator precedence it makes it more readable IMHO.
```suggestion
			want := 2 * tc.retryCount + 1
			if *atomicCounter != want {
				t.Errorf("expected retry %d, got %d", tc.retryCount, *atomicCounter)
			}
```

storage/s3_test.go
```diff
@@ -565,6 +566,87 @@ func TestS3Retry(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer os.Remove(f.Name())
+
+			mockS3.Put(ctx, f, url, NewMetadata(), s3manager.DefaultUploadConcurrency, s3manager.DefaultUploadPartSize)
+
+			// +1 is for the original request
+			// *2 is to account for the "Stat" requests that are made to obtain
+			// last modification date( retry decision are made according to mod. date)
```
seruman, line 642: > last modification date( retry decision are made according to mod. date)

Seems outdated, isn't it made depending on existence of custom metadata value?

storage/s3.go
seruman: If it returns early depending on decision to retry, moving retry logic to another function/method would makes this function less crowded.

I was thinking something like;

```go
_, err := s.uploader.UploadWithContext(/* */)

if ! /* conditions to retry /* {
	return err
}

return s.someRetryMethod()


```

command/cp.go
```diff
@@ -177,6 +177,12 @@ func NewSharedFlags() []cli.Flag {
 			Name:  "content-encoding",
 			Usage: "set content encoding for target: defines content encoding header for object, e.g. --content-encoding gzip",
 		},
+		&cli.IntFlag{
+			Name:        "no-such-upload-retry-count",
+			Usage:       "number of times that a request will be retried on NoSuchUpload error. You should not use this unless you really know what you're doing.",
```
seruman, line 182: Looks like we use all lower case in usage texts -except abbrevations-, let's not use upper case to start a new sentence to make it consistent.  `;` might be used to separate sentences as done in `sse-kms-key-id` flag.

storage/s3.go
```diff
@@ -125,6 +135,20 @@ func (s *S3) Stat(ctx context.Context, url *url.URL) (*Object, error) {
 
 	etag := aws.StringValue(output.ETag)
 	mod := aws.TimeValue(output.LastModified)
+
+	if s.noSuchUploadRetryCount > 0 {
+		m := output.Metadata
+		if res, ok := m[META_DATA_S5CMD_RETRY_CODE_KEY]; ok {
```
seruman, line 141: nitpick: I would prefer the name `metadata`for the variable or wouldn't define it as access to it is not that long.

seruman (changes requested):
Also to make sure the default behaviour is not broken, a couple of e2e tests would be good with retry flag set; even though it would not retry.

storage/s3.go
seruman: Could you give context on why retry code is re-generated on each retry? Could not the same one re-used?

storage/s3_test.go
seruman: Error output seems wrong, counter value is expected to be `2 * tc.retryCount + 1` but reported expected value is `tc.retryCount`.

storage/s3.go
```diff
@@ -125,12 +133,19 @@ func (s *S3) Stat(ctx context.Context, url *url.URL) (*Object, error) {
...
 		URL:     url,
 		Etag:    strings.Trim(etag, `"`),
 		ModTime: &mod,
 		Size:    aws.Int64Value(output.ContentLength),
-	}, nil
+	}
+
+	if res, ok := output.Metadata[metadataKeyRetryCode]; ok && s.noSuchUploadRetryCount > 0 {
+		obj.retryCode = *res
+	}
```
seruman, line 146: I think something like below would make it more readable. One may easily spot this only works with `noSuckUploadRetryCode > 0`
```suggestion
	if s.noSuchUploadRetryCount > 0 {
		if retryCode, ok := output.Metadata[metadataKeyRetryCode];ok {
			obj.retryCode = *retryCode
		}
	}
```

storage/s3.go
```diff
@@ -983,3 +1009,41 @@ func errHasCode(err error, code string) bool {
...
+		_, err = s.uploader.UploadWithContext(ctx, input, uploaderOpts...)
+	}
+
+	if errHasCode(err, s3.ErrCodeNoSuchUpload) && s.noSuchUploadRetryCount > 0 {
+		err = awserr.New(s3.ErrCodeNoSuchUpload, fmt.Sprintf(
+			"RetryOnNoSuchUpload: %v attempts to retry resulted in %v", attempts,
+			s3.ErrCodeNoSuchUpload), err)
+	}
+	return err
+}
```
seruman, line 1049: Could you move this function close to the where it used?

seruman (changes requested):

Kucukaslan:
I've tested the read-after-write consistency of [object metadata](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html#ConsistencyModel:~:text=Object%20Tags%2C%20and-,object%20metadata%20(for%20example%2C%20the%20HEAD%20object)%20are%20strongly%20consistent,-.%20For%20more%20information) in [another branch](https://github.com/Kucukaslan/s5cmd/tree/retry-count-metadata-read-after-write-test). The storage.Stat request (after upload) operation for all of over 1000 objects returned the updated metadata. 
Example command execution to experimnet (with binary compiled from that branch):
`s5cmd --stat cp --no-such-upload-retry-count  1  vendor/ s3://BUCKET/KEY_PREFIX/`

seruman (approved):

command/cp.go
```diff
@@ -251,8 +257,8 @@ type Copy struct {
 	acl                   string
 	forceGlacierTransfer  bool
 	ignoreGlacierWarnings bool
-	exclude               []string
 	raw                   bool
+	exclude               []string
```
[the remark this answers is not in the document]
Kucukaslan: Yes and no. In first draft (commit) I was going to add a field to Copy object. While I add `retryOnNoSuchUploadError` field to Copy, I thought it was weird a `string slice` is placed between `bool` fields. So I moved `exclude` field to 1 line below. Then, after I've removed `retryOnNoSuchUploadError` field from Copy object, I didn't reverted the place of `exclude` field since it was minor, and at worst harmless change.

storage/s3.go
```diff
@@ -43,6 +46,9 @@ const (
 
 	// Google Cloud Storage endpoint
 	gcsEndpoint = "storage.googleapis.com"
+
+	// the key of the object metadata which is used to handle retry decision on NoSuchUpload error
+	metadataKeyRetryCode = "S5cmd-Retry-Code"
```
igungor, line 51:
```suggestion
	metadataKeyRetryCode = "S5cmd-Retry-Code"
```

How about `s5cmd-upload-retry-id`?

storage/s3.go
```diff
@@ -125,12 +133,21 @@ func (s *S3) Stat(ctx context.Context, url *url.URL) (*Object, error) {
...
+	obj := &Object{
 		URL:     url,
 		Etag:    strings.Trim(etag, `"`),
 		ModTime: &mod,
 		Size:    aws.Int64Value(output.ContentLength),
-	}, nil
+	}
+
+	if s.noSuchUploadRetryCount > 0 {
+		if retryCode, ok := output.Metadata[metadataKeyRetryCode]; ok {
```
igungor, line 145: How about `retryID`?

storage/s3.go
```diff
@@ -514,6 +531,7 @@ func (s *S3) Put(
 		Key:          aws.String(to.Path),
 		Body:         reader,
 		ContentType:  aws.String(contentType),
+		Metadata:     make(map[string]*string),
```
igungor, line 534:
```suggestion
		Metadata:     make(Metadata),
```

storage/s3.go
```diff
@@ -554,11 +572,53 @@ func (s *S3) Put(
...
+	}
+
+	uploaderOptsFn := func(u *s3manager.Uploader) {
 		u.PartSize = partSize
 		u.Concurrency = concurrency
-	})
+	}
+	_, err := s.uploader.UploadWithContext(ctx, input, uploaderOptsFn)
+
+	if !(errHasCode(err, s3.ErrCodeNoSuchUpload) && s.noSuchUploadRetryCount > 0) {
```
igungor, line 586: Could you switch this logic? It's unreadable to me :)

I'd write this like:

```go
if errHasCode(err, s3.ErrCodeNoSuchUpload) && s.noSuchUploadRetryCount > 0 {
	return s.retryOnNoSuchUpload(ctx, to, input, err, uploaderOptsFn)
}

return err
```

storage/s3.go
Kucukaslan: > moving retry logic to another function/method

I'd some concerns about it. First, I think we should limit the AWS-SDK calls to the storage methods,  so having another method that also uses AWS-SDK (`s.uploader.UploadWithContext`) might be an extra maintenance work. Second, that method needs (in fact [needed](https://github.com/Kucukaslan/s5cmd/blob/9770a80d009bb7ad29588e80b6e81caf374d8b6a/storage/s3.go#L593)) a lot of (5+) parameters. I would prefer not to move a logic that needs so many arguments. 

Nevertheless, I agree that readability is improved by moving it to another method. Especially after considering that that logic was an exceptional/tangential case.

> why retry code is re-generated on each retry
I thought that, at each turn (in the loop) we would check if the _the previous_ upload was successful. However, after discussion, I come to think that if _any of the previous_ uploads were successful than we should not retry the upload. So, I'll use the same retryCode for all of them.

storage/s3.go
Kucukaslan: The [storage.Metadata](https://github.com/peak/s5cmd/blob/123b1c7fc9c614aa214a6795468fa140a38ad05e/storage/storage.go#L202) is `map[string]string`  but the [s3manager.UploadInput.Metadata](https://github.com/aws/aws-sdk-go/blob/3f477ea07655cad1462020fed4c67bfe1c555dc2/service/s3/s3manager/upload_input.go#L176) is `map[string]*string`
So they're, unfortunately of different types.

storage/s3.go
```diff
@@ -554,11 +572,53 @@ func (s *S3) Put(
...
+		return err
+	}
 
+	return s.retryOnNoSuchUpload(ctx, to, input, err, uploaderOptsFn)
+}
+
+func (s *S3) retryOnNoSuchUpload(ctx aws.Context, to *url.URL, input *s3manager.UploadInput,
+	err error, uploaderOpts ...func(*s3manager.Uploader)) error {
+
+	var retryCode string
```
igungor, line 596: Could you rename this variable to `expectedRetryID`?

storage/s3.go
```diff
@@ -983,3 +1043,9 @@ func errHasCode(err error, code string) bool {
 func IsCancelationError(err error) bool {
 	return errHasCode(err, request.CanceledErrorCode)
 }
+
+// generate a retry code for this upload attempt
+func generateRetryCode() *string {
```
igungor, line 1048:
```suggestion
func generateRetryID() *string {
```

storage/s3.go
```diff
@@ -43,6 +46,9 @@ const (
 
 	// Google Cloud Storage endpoint
 	gcsEndpoint = "storage.googleapis.com"
+
+	// the key of the object metadata which is used to handle retry decision on NoSuchUpload error
+	metadataKeyRetryId = "S5cmd--Upload-Retry-Id"
```
igungor, line 51:
```suggestion
	metadataKeyRetryId = "s5cmd-upload-retry-id"
```

igungor: all lowercase.

storage/s3.go
igungor: Id -> ID.

storage/s3_test.go
```diff
@@ -565,6 +566,82 @@ func TestS3Retry(t *testing.T) {
...
+				atomic.AddInt32(atomicCounter, 1)
+			})
+
+			mockS3.Put(ctx, strings.NewReader(""), url, NewMetadata(), s3manager.DefaultUploadConcurrency, s3manager.DefaultUploadPartSize)
+
+			// +1 is for the original request
+			// *2 is to account for the "Stat" requests that are made to obtain
+			// retry code from object metada.
+			want := 2*tc.retryCount + 1
+			if *atomicCounter != want {
```
igungor, line 638: Better use atomic.LoadInt32.

storage/storage.go
```diff
@@ -108,6 +110,7 @@ type Object struct {
 	Size         int64        `json:"size,omitempty"`
 	StorageClass StorageClass `json:"storage_class,omitempty"`
 	Err          error        `json:"error,omitempty"`
+	retryId      string
```
igungor, line 113:
```suggestion
	retryID      string
```

storage/s3.go
Kucukaslan: Thanks. The extra hyphen was a typo but capitalization was not a typo.

Metadata fields are stored in lowercase[^lower]. But when the HeadObjectOutput.Metadata is created "By default ... the first letter and any letter following a hyphen will be capitalized, and the rest as lowercase... Set `aws.Config.LowerCaseHeaderMaps` to `true` to write ... as lowercase" [^meta]. So we should either 
1) change the key to match default behaviour as I did
2) or use call "WithLowerCaseHeaderMaps(true)" when the AWS session configuration is created.

In the AWS-SDK v2 this default behaviour is changed[^v2]. So I will convert it to lower case as you've suggested and add the configuration option (which is to be removed. when we migrate to v2).

[^lower]: User-defined metadata is a set of key-value pairs. [Amazon S3 stores user-defined metadata keys in lowercase ](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#:~:text=User%2Ddefined%20metadata%20is%20a%20set%20of%20key%2Dvalue%20pairs.%20Amazon%20S3%20stores%20user%2Ddefined%20metadata%20keys%20in%20lowercase).
[^meta]: https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#HeadObjectOutput
[^v2]: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#HeadObjectOutput.Metadata

igungor: Good to know, thank you.

Kucukaslan: You're welcome!

igungor (approved):

PR #472, opened 2022-07-22
vendor: upgrade gofakes3 to v0.0.11

### LINKED ISSUE #446: Tests failed on AlpineLinux 32bit architecture
Hi! I'm trying to package `s5cmd` for AlpineLinux. The attempt can be seen [here](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/35426).

Some problems I'm facing:
- `e2e` tests are failing on 32bit architecture builds. [Sample logs on x86](https://gitlab.alpinelinux.org/FollieHiyuki/aports/-/jobs/746921/raw)
- `-race` flag is specified in [e2e/util_test.go line 216](https://github.com/peak/s5cmd/blob/7200415a06500f88cad95ed48e7aa33bc023a915/e2e/util_test.go#L216) which is not compatible with `-buildmode=pie`. Is there a way to configure it during the build process?

I'd love to get some advice. And thanks for `s5cmd`!

kucukaslan:
Upgraded gofakes version to v0.0.11 to solve unaligned bit error in 32 bit architectures.
Allow disabling race flag (race flag is incompatible with `-buildmode=pie`)

Fixes #446

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,17 +11,21 @@ LDFLAGS=-ldflags "-X=github.com/peak/s5cmd/version.Version=$(VERSION) -X=github.
 build:
 	@go build ${GCFLAGS} ${LDFLAGS} -mod=vendor .
 
-RACE_FLAG := -race
+TEST_TYPE:=test_with_race
 ifeq ($(OS),Windows_NT)
-	RACE_FLAG =
+	TEST_TYPE=test_without_race
 endif
 
 .PHONY: test
-test:
-	@S5CMD_TEST_DISABLE_RACE=0 go test -mod=vendor -count=1 ${RACE_FLAG} ./...
+test: $(TEST_TYPE)
 
+.PHONY: test_with_race
+test_with_race:
+	@S5CMD_TEST_DISABLE_RACE=0 go test -mod=vendor -count=1 -race ./...
+
+.PHONY: test_without_race
 test_without_race:
-	@S5CMD_TEST_DISABLE_RACE=1 go test -mod=vendor -count=1 ${RACE_FLAG} ./...
+	@S5CMD_TEST_DISABLE_RACE=1 go test -mod=vendor -count=1 ./...
 
 .PHONY: check
 check: vet staticcheck unparam check-fmt
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,10 @@ endif
 
 .PHONY: test
 test:
-	@go test -mod=vendor -count=1 ${RACE_FLAG} ./...
+	@S5CMD_TEST_DISABLE_RACE=0 go test -mod=vendor -count=1 ${RACE_FLAG} ./...
+
+test_without_race:
+	@S5CMD_TEST_DISABLE_RACE=1 go test -mod=vendor -count=1 ${RACE_FLAG} ./...
 
 .PHONY: check
 check: vet staticcheck unparam check-fmt
--- a/Makefile
+++ b/Makefile
@@ -21,11 +21,11 @@ test: $(TEST_TYPE)
 
 .PHONY: test_with_race
 test_with_race:
-	@S5CMD_TEST_DISABLE_RACE=0 go test -mod=vendor -count=1 -race ./...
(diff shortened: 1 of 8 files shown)


igungor (approved):

ilkinulas (approved):

Kucukaslan:
The commit https://github.com/peak/s5cmd/pull/472/commits/96b27c2e48ae2fbd824a696adf36a3c16f9e3d36 provides a way to solve second problem of #446.

Using
```
go test -mod=vendor -count=1 ${RACE_FLAG} `go list ./... | grep -v e2e`
go test -mod=vendor -count=1 ${RACE_FLAG} ./e2e -args -test.race.disabled
```
instead of `go test -mod=vendor -count=1 ${RACE_FLAG} ./...` would disable race flag in the e2e package tests.

e2e package is treated differently because only it has `test.race.disabled` flag. Alternatively, we can add this flag to all packages (to those have tests). But I prefered not to add unused flags to them just to allow writing `go test -mod=vendor -count=1 ${RACE_FLAG} ./e2e -args -test.race.disabled`.

Kucukaslan:
With https://github.com/peak/s5cmd/commit/96b27c2e48ae2fbd824a696adf36a3c16f9e3d36, e2e package request an _environment variable_ (named _S5CMD_TEST_DISABLE_RACE_) to disable the race flag when compiling the test binary.
To handle both cases we added a new Makefile target named _test_without_race_ to be used in the windows and other distributions that defaults go's buildmode to PIE.

TBH, I'm not sure what reasons do we still have to keep Makefile's _RACE_FLAG_ variable. Or should we (or can be) add another target *test_with_race* and call one of _test_with_race_ and _test_without_race_ from _test_ target (in this case change in ci.yml can be reverted)
- [ ] Add _test_with_race_ target if needed.

Makefile
```diff
@@ -11,14 +11,21 @@ LDFLAGS=-ldflags "-X=github.com/peak/s5cmd/version.Version=$(VERSION) -X=github.
...
 endif
 
 .PHONY: test
-test:
-	@go test -mod=vendor -count=1 ${RACE_FLAG} ./...
+test: $(TEST_TYPE)
+
+.PHONY: test_with_race
+test_with_race:
+	@S5CMD_TEST_DISABLE_RACE=0 go test -mod=vendor -count=1 -race ./...
```
igungor, line 24: > @S5CMD_TEST_DISABLE_RACE=0 go test -mod=vendor -count=1 -race ./...

Environment variable name is misleading. We build the test binary with race disabled, while running the tests in parallel.

Let's change the variable name to `S5CMD_BUILD_BINARY_WITHOUT_RACE_FLAG`.

e2e/main_test.go
```diff
@@ -1,11 +1,16 @@
 package e2e
 
 import (
+	"flag"
 	"os"
 	"testing"
 )
 
 func TestMain(m *testing.M) {
+	if !flag.Parsed() {
```
igungor, line 10: This line shouldn't be necessary.

e2e/util_test.go
```diff
@@ -38,6 +38,12 @@ import (
 	"gotest.tools/v3/icmd"
 )
 
+const (
+	// Don't use "race" flag in the build arguments.
+	TEST_DISABLE_RACE_FLAG_KEY   = "S5CMD_TEST_DISABLE_RACE"
```
igungor, line 43: const names must be camelCase.

e2e/main_test.go
Kucukaslan: I've deleted the line. But I'm not sure do we even need flags at all?
The only flag we defined is [test.log.level](https://github.com/peak/s5cmd/blob/c7dfad601797f1fcd41186cabf1ce1b5ef80fdc4/e2e/util_test.go#L57) and that flag is not used in the Makefile. So it (flagTestLogLevel) takes the default value _err_.  Should I remove them all?

igungor: Hmm, not sure. In any case, its outside the scope of this PR :)

PR #475, opened 2022-07-26
command/storage: add versioning support 

### LINKED ISSUE #218: du: include versioned objects
Hi,

Feature request:
On "du" add an option to include versioned objects in order to obtain a total bucket size.

Thanks.

### LINKED ISSUE #386: How to delete object versions?
I want to delete a non-empty bucket that has versioning enabled.

To do that in the S3 console first I would

1. use the "Empty Bucket" option
2. use the "Delete Bucket" option

I see that to cover step 2 s5cmd has an rb command.

To cover step 1, I tried to use the rm command with syntax like this: `s3://bucketname/*`

It placed a delete marker on all the objects in the bucket, but it didn't permanently delete any verisons.

So S3 still considers the bucket non-empty, and so step 2 fails.

```
$ s5cmd rm s3://bucketname/*
rm s3://bucketname/...
...

$ s5cmd rb s3://bucketname
ERROR "rb s3://bucketname": BucketNotEmpty: The bucket you tried to delete is not empty. You must delete all versions in the bucket. status code: 409, request id: XXXXXXXX, host id: XXXXXXX
```

I don't see an option for permanent deletion in the rm command help.

Am I missing something?

I'm using version v1.4.0-d7a0dda.

### LINKED ISSUE #539: Support `--raw` flag with `cat`
This PR #315 has added `--raw` as a flag for `cp` and `rm` operations. It would be nice to have it for `cat` as well. Currently, we have to use AWS CLI to cat a file on S3 with "*" in the object key; keen on being able to use s5cmd for everything :)

kucukaslan:
This commit adds versioning support to the s5cmd.

- Added `--all-versions` flag to `ls`, `rm`, `du` and `select` subcommands 
 to apply operation on(/over) all versions of the objects. 
- Added `--version-id` flag to `cat`, `cp`/`mv`, `rm`, `du`  and `select` 
 subcommands to apply operation on(/over) a specific versions of the object. 
- Added `bucket-version` command to configure bucket versioning. Bucket name
 alone returns the bucket versioning status of the bucket. Bucket versioning can 
 be configured with `set` flag which only accepts.
- Added `--raw` flag to `cat` and `select` subcommands. It disables the wildcard operations.

Note: Google Cloud Storage uses a different approach for versioning. So with current implementation,  s5cmd cannot use or retrieve[ generation numbers ](https://cloud.google.com/storage/docs/metadata#generation-number). However, `bucket-version` command and `du` command with `all-versions` flag works as expected since they do not use version ids.

Fixes: #218 
Fixes: #386
Fixes: #539


Kucukaslan:
Status as of July 26:
* add all-versions flag to following subcommands:
  - [x] ls
  - [ ] rm 
  - [x] du
* add `version-id` flag to following sub commands:
  - [ ] cp
  - [x] cat
  - [x] rm
  - [x] du

Current problem: `rm` uses expandSource method so It doesn't work when `all-versions` of a particular key was to be deleted, though it. succesfully deletes if wildcard used. expandSource don't have access to values of those flags.
* I want to move versionId/allVersion fields to the URL.
* I want to change the type of src & dst fields of commands to the URL accordingly.

Alternatively, we can add new parameters to expandSource method.

Kucukaslan:
## Update
### Status
### Status as of August 3rd:
* add `all-versions` flag to following subcommands:
  - [x] ls ( including _delete markers_)
  - [x] rm ( including _delete markers_)
  - [x] du
* add `version-id` flag to following subcommands:
  - [x] cp/mv
  - [x] cat
  - [x] rm
  - [x] du
* format outputs
  - [ ] discuss how version id should be placed (append key with `?versionID=` to make it similar to request url)
  - [x] ls
  - [ ] cp
  ... 
* add `set` and `get` flags to `version` subcommand
  - [x] set
  - [x] get
  - [ ] discuss if this is correct way/place to do or should we new flag(s).
* testing
  - [x] prepare versioning setup for gofakes3
  - [ ] ls
  - [ ] cp
  - [ ] ...

### Commentary & Known Issues
* `gofakes3` package that we use in our tests supports versioning only with `in memory backend`, so I've added another method to setup fake server.
* There is a *bug* when I try to delete from gofakes3 server using `s5cmd rm`. Despite using `version-id`/`all-versions` flags, the server does not permanently delete the corresponding objects and just adds _delete marker_ to them. Interestingly:
    * this bug does not happen when I use `aws s3api delete-object` to connect gofakes3 server. 
    * this bug does not happen when I use `s5cmd rm` to connect *real* AWS S3 server.
    * other subcommands of s5cmd works as expected. 
 I'm currently trying to identify root cause of this bug and to fix it.

Kucukaslan:
> add `set` and `get` flags to `version` subcommand

Alternatively we can remove the get flag and use this syntax:
>`$ s5cmd version`
v0.0.0-dev
`$ s5cmd version s3://bucket`
Bucket versioning for "bucket" is "Enabled"
`$ s5cmd version --set Enabled s3://mcks5cmd`
Bucket versioning for "bucket" is set to "Enabled"

ps. At the moment to get bucket versioning we need to write:
`$ s5cmd version --get s3://bucket`

command/version.go
```diff
@@ -1,21 +1,169 @@
...
+	{{.HelpName}} [options] s3://bucketname
+
+Options:
+	{{range .VisibleFlags}}{{.}}
+	{{end}}
+Examples:
+	1. Check the current version of the s5cmd
+	 	 > s5cmd {{.HelpName}}
+	
+	2. Get the bucket versioning status of a bucket
```
sonmezonur, line 29: I think these should not be part of `version` command.  The common practice of `version` subcommand is to display the version of the tool, these added features are related to `S3 object versions`.

Kucukaslan: Added `bucket-version` command to configure bucket versioning. Removed bucket versioning related logic from the `version` command.

Kucukaslan:
### On the Google Cloud Storage
It has generation numbers analogous to S3 version ids.
In its REST API it uses `generation` tag while the AWS S3 uses `VersionId` tag. So the (Un)Marshalers in [AWS SDK](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.44.90/private/protocol) does not handle generation tag. Hence it can neither get generation number from the response nor can send it with the request.

I don't think that intervening into Marshaler's logic via [request handlers](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.44.90/aws/request#Request:~:text=metadata.ClientInfo-,Handlers%20%20%20Handlers,-Retryer%0A%09AttemptTime%20%20time) would be an acceptable/practical solution, even if it were to be possible.

As a last resort I've tried to modify AWS. SDK to add `Generation` fields to relevant types[^fork]. It helped to read generation numbers in List request without breaking any other thing, that is `ls --all-versions` worked with GCS.  

However, it still failed to use those generation numbers in requests,  that is `--version-id` flag and `rm/cp... --all-versions` did not work. I've tried a few other modifications to SDK, but none of them worked with GCS without breaking the AWS S3.  
Even if these attempts were to be succesfull, upstream would not have accepted these changes and we would need to use a custom fork. 

ps. I've used the first version of AWS-SDK-GO but I'm not optimistic that using v2 (or. its [middlewares](https://aws.github.io/aws-sdk-go-v2/docs/middleware/)) would made any difference


### RFC: How should s5cmd act when versioning flags are used with Google Cloud endpoints?
> **Note**
Only `bucket-version` command and `du` command with `--all-versions` flag works accurately with GCS.

- Should it print an error and cancel the operation?
- Should it print a warning and continue to the operation even though the result would not be the one user expected?

[^fork]: The attempt may be seen from [here](https://github.com/Kucukaslan/s5cmd/commit/1c9b23c7feb4bc0bc8cecbe33e48dd594eb68dd7).

command/util.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// checkNumberOfArguments checks if the number of the arguments is valid.
+// if the max is negative then there is no upper limit of arguments.
+func checkNumberOfArguments(ctx *cli.Context, min, max int) error {
+	l := ctx.Args().Len()
+	if min == 1 && max == 1 && l != 1 {
+		return fmt.Errorf("expected only one argument")
+	}
+	if min == 2 && max == 2 && l != 2 {
```
sonmezonur, line 65: as far as I can see this function is only called on [bucket_version.go](https://github.com/Kucukaslan/s5cmd/blob/d48b45b2596233e37411182624487d2cc27034e0/command/bucket_version.go#L52) and its min/max parameters are always 1/1. Do we still need to check other values (for example 2/2 in this line)?

command/util.go
```diff
@@ -0,0 +1,75 @@
+package command
```
sonmezonur, line 1: not sure about the file name since all functions below are used in versioning. maybe it might be a part of `bucket-version.go`.

command/flag.go
```diff
@@ -28,3 +28,7 @@ func (e EnumValue) String() string {
 	}
 	return e.selected
 }
+
+func (e EnumValue) Get() interface{} {
```
sonmezonur, line 32: 👍  thanks for fix

command/bucket_version.go
```diff
@@ -0,0 +1,145 @@
...
+	return &cli.Command{
+		Name:               "bucket-version",
+		CustomHelpTemplate: bucketVersionHelpTemplate,
+		HelpName:           "bucket-version",
+		Usage:              "configure bucket versioning",
+		Flags: []cli.Flag{
+			&cli.GenericFlag{
+				Name: "set",
+				Value: &EnumValue{
+					Enum:    []string{"Suspended", "Enabled"},
```
sonmezonur, line 45: What do you think of making `Suspended` and `Enabled` case-insensitive? It might be more user-friendly.

sonmezonur (commented):
Left some minor comments but LGTM overall. Thank you!

command/util.go
Kucukaslan: No but yes too. I intentionally made it general function. Becuase I hope it can be used by all commands in future. But I didn't change other commands' validations checks, since they would be rather irrelevant to this PR.

command/util.go
Kucukaslan: I will rename it to `validation.go`.

command/flag.go
Kucukaslan: My pleasure.

command/bucket_version.go
Kucukaslan: The API requests is case sensitive.
The `aws s3api` is also case sensitive:
The following fails
```
$ aws s3api put-bucket-versioning --bucket mybucketname --versioning-configuration Status=suspended
````
>An error occurred (MalformedXML) when calling the PutBucketVersioning operation: The XML you provided was not well-formed or did not validate against our published schema

while the following succeds.

```
$ aws s3api put-bucket-versioning --bucket mybucketname --versioning-configuration Status=Suspended
```

mc has a different approach uses subcommands:

```
mc version enable myminio/mydata
mc version info myminio/mydata
mc version suspend myminio/mydata
```

Nevertheless, it could be a nice UX improvement, so I'll give it a try.

File: command/util.go (hunk 2 above)
Comment (sonmezonur) [reply command/util.go:65]: 👍 it makes sense now.

File: command/bucket_version.go (hunk 5 above)
Comment (Kucukaslan) [reply command/bucket_version.go]: I made the changes and committed.

Comment (sonmezonur) [reply command/bucket_version.go]: thank you. it is much better

File: storage/s3.go [hunk 6]
```diff
@@ -801,6 +966,35 @@ func (s *S3) RemoveBucket(ctx context.Context, name string) error {
...
+		VersioningConfiguration: &s3.VersioningConfiguration{
+			Status: aws.String(versioningStatus),
+		},
+	})
+	return err
+}
+
+// GetBucketVersioning returnsversioning property of the bucket
+func (s *S3) GetBucketVersioning(ctx context.Context, bucket string) (string, error) {
+
```
sonmezonur, line 986: please remove empty line

sonmezonur (dismissed):

command/bucket_version.go
Kucukaslan: My pleasure, thanks!

sonmezonur (dismissed):

sonmezonur (dismissed):

sonmezonur (approved):

igungor (approved):

igungor:
🥇

PR #477, opened 2022-07-28
command: fix target file is created despite the download failure

### LINKED ISSUE #348: cp command downloads file from S3 even though the file does not exist on the S3 bucket
When cp command is executed to download a file from the S3 bucket I get the AWS error in console 'NoSuchKey: The specified key does not exist.' but the file with the same name is created on my local storage.
Is there any way to prevent this behaviour?

kucukaslan:
Before downloading a file from s3 a local target file is created. If the download fails the created file should've been deleted. But since the file was not closed, the delete operation fails in Windows.

Fixes #348


command/cp.go
```diff
@@ -522,7 +522,12 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL)
 
 	size, err := srcClient.Get(ctx, srcurl, file, c.concurrency, c.partSize)
 	if err != nil {
-		_ = dstClient.Delete(ctx, dsturl)
+		// file must be closed before deletion
+		file.Close()
```
igungor, line 526: https://github.com/peak/s5cmd/pull/477/files#diff-8e0740ea5af04e2014c49fe30fc0dcb04ccbc849a76160ca4ec80845c248b511R521

will return error because file is first closed here.

Kucukaslan: Yes the "defer file.Close()" will give an error if we close the file here. But that error will be discarded (it won't cause a panic).

As an alternative we can move/write that defer statement to after the [if](https://github.com/Kucukaslan/s5cmd/blob/fix-348-delete-local-file/command/cp.go#L524-L532) statement.

Kucukaslan: Yes the "defer file.Close()" will give an error if we close the file here. But that error will be discarded (it won't cause a panic).

As an alternative we can move/write that defer statement to after the [if](https://github.com/Kucukaslan/s5cmd/blob/fix-348-delete-local-file/command/cp.go#L524-L532) statement.

aykutfarsak: Maybe we can put `defer file.Close()` to else statement to prevent silent error.

Kucukaslan: Yes. I, now, put  it after the if statement. 
The if statements terminates the function, so I didn't enclosed `defer file.Close()` with else.

aykutfarsak (approved):

igungor (approved):

PR #481, opened 2022-08-08
ci: use Go 1.19 for builds, remove 1.16 support

kucukaslan:
* use io and os packages to replace deprecated ioutil functions
* update go.mod go requirement to 1.17


Kucukaslan:
1) Should I also change the go.mod file?
https://github.com/peak/s5cmd/blob/7c90f8a85ffac2fef658541c8c566abd29e5577e/go.mod#L3

ps. as of 1.17 indirect dependencies are brought to go.mod in the root https://go.dev/doc/modules/gomod-ref#go

2) ioutil is deprecated should I change them in this PR?

Kucukaslan:
it seems `go fmt` is already being checked:
https://github.com/peak/s5cmd/blob/7c90f8a85ffac2fef658541c8c566abd29e5577e/.github/workflows/ci.yml#L63
and
https://github.com/peak/s5cmd/blob/7c90f8a85ffac2fef658541c8c566abd29e5577e/Makefile#L46-L50

igungor (approved):

seruman (approved):

PR #490, opened 2022-08-17
Auto completion

### LINKED ISSUE #207: Auto completion for local files
It would be handy, if s5cmd could support auto-completion for local files.

For example
```
> ls -l
total 0
-rw-r--r--  1 ilkinulas  staff  0 Aug  7 23:22 a.txt
-rw-r--r--  1 ilkinulas  staff  0 Aug  7 23:22 b.txt

> s5cmd cp <TAB>
```
should allow me to select `a.txt` or `b.txt`.

### LINKED ISSUE #372: Bucket/key completion not supported?
Hi,

I found some documentation (possibly for an old version?) which says that s5cmd supports bucket and key name completion:
https://reposhub.com/go/miscellaneous/peak-s5cmd.html

While I can use command completion just fine after running `--install-completion` (with s5cmd 1.1), it seems that bucket/key completion is not supported (at least I don't see any reaction when pressing tab while in an S3 URI). I also cannot find the implementation of such a feature in the master branch. Has the feature been removed or is there another branch I should be looking at?

Thanks!

Kucukaslan:
Adds extensive auto-completion support to s5cmd. Including: 
- command completion
- flag completion
- local file/directory completion
- bucket name completion
- s3 key completion

Fixes #207 
Fixes #372


Kucukaslan:
I'm closing this PR in favor of #500.

Among the autocompletion solutions:

### Posener/complete 
it supports zsh, bash and fish
it has two relevant[^rlvt] advantages :
- it is possible to specify the autocompletion suggestions for flag arguments
- the library handles the (un)installation of completion scripts

it has 3 disadvantages
- it cannot handle the colons in bash. the notorious COMP_WORDBREAKS must be changed, unacceptable.
- all of the commands and flags should be also defined for it, at least 'adapted' to it. This problem also avoids the "flag argument suggestion" advantage
- I forgot the 3rd one it[^imp]


### using urfave/cli
it supports zsh, bash and pwsh
#### Changing COMP_WORDBREAKS
straightforward and easy but unacceptable.

#### using well-known `bash-completion`
relatively easy and feels more reliable since it was being used by many applications for years.
But it is not preferable to tell the user "please also import this other library written in [bash](http://harmful.cat-v.org/software/#:~:text=bash)".

#### write your own script and delegate most of the logic into app
- it took a lot of time,
- it required to write shell script, deal with [COMP_WORDBREAKS](https://github.com/scop/bash-completion/blob/40aa4dc3c75791f43f850871466039768b3048ad/completions/_mount.linux#L207) on your own
- ineffective use of time
- slightly better for user experience, not ideal.
- did I mentioned the time spent on it?

### Decision
For obvious reasons, I've prefered using urfave/cli with our own bash completion script.
The user will be given the instructions and the completion script when she called s5cmd with `install-completion` flag. The instructions and script will be written according to value of `SHELL` environment variable. 

[^rlvt]: An example of "irrelavant" advantages is that you can use posener/complete to write completion for any cli application.
[^imp]: https://images.app.goo.gl/pxCKgdQ61oDYtvRa6

PR #491, opened 2022-08-19
Checksum verification

boraberke:


boraberke:
Below are the results of a benchmark test of small files. With checksum verification, it slows down around %50.

|Scenario| Summary |
|:---|:---|
| upload small files | 'PR:478' ran 1.69 ± 0.41 times faster than 'PR:491' |
| download small files | 'PR:478' ran 1.56 ± 0.05 times faster than 'PR:491' |

boraberke:
This PR is still in draft as it is dependent on migrating to aws-sdk-go-v2. Even if migration to v2 happens, this implementation may not be accepted as default as it slows down the performance of `s5cmd` by at least 50%.

sonmezonur:
Closing this PR since we have decided not to migrate to aws-sdk-go-v2. Thanks @boraberke

PR #496, opened 2022-08-24
command/app: force `endpoint-url` to have scheme

boraberke:
Before this PR, if an URL without a scheme was given to `endpoint-url` such as `example.com`, `s5cmd` would add `http` scheme to it. 

After this PR, `s5cmd` will not accept `endpoint-url` without a scheme. User should give either `http` or `https` scheme explicitly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 #### Breaking changes
 - Adjacent slashes in key are no longer removed when uploading to remote. Before `s5cmd cp file.txt s3://bucket/a//b///c/` would copy to `s3://bucket/a/b/c/file.txt` but now to `s3://bucket/a//b///c/file.txt`.([#459](https://github.com/peak/s5cmd/pull/459))
+- `--endpoint-url` will not accept URL's without scheme such as `example.com`. Instead, it will give an error and ask for an url with a scheme; either `http://example.com` or `https://example.com` ([#496](https://github.com/peak/s5cmd/pull/496)).
 
 #### Features
 - Added `--content-type` and `--content-encoding` flags to `cp` command. ([#264](https://github.com/peak/s5cmd/issues/264))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Breaking changes
 - Adjacent slashes in key are no longer removed when uploading to remote. Before `s5cmd cp file.txt s3://bucket/a//b///c/` would copy to `s3://bucket/a/b/c/file.txt` but now to `s3://bucket/a//b///c/file.txt`.([#459](https://github.com/peak/s5cmd/pull/459))
-- `--endpoint-url` will not accept URL's without scheme such as `example.com`. Instead, it will give an error and ask for an url with a scheme; either `http://example.com` or `https://example.com` ([#496](https://github.com/peak/s5cmd/pull/496)).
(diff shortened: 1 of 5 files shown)


e2e/app_test.go
```diff
@@ -227,3 +227,62 @@ func TestInvalidLoglevel(t *testing.T) {
...
+			result.Assert(t, icmd.Expected{ExitCode: tc.expectedExitCode})
+
+			if tc.expectedError == nil {
+				if result.Stderr() != "" {
+					t.Fatalf("expected no error, got: %q", result.Stderr())
+				}
+				return
+			}
+
+			if result.Stderr() == "" {
```
igungor, line 280: could you assert stderr error in a single if block?

command/app.go
```diff
@@ -96,6 +97,7 @@ var app = &cli.App{
 		printJSON := c.Bool("json")
 		logLevel := c.String("log")
 		isStat := c.Bool("stat")
+		endpointUrl := c.String("endpoint-url")
```
igungor, line 100: endpointURL

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 #### Breaking changes
 - Adjacent slashes in key are no longer removed when uploading to remote. Before `s5cmd cp file.txt s3://bucket/a//b///c/` would copy to `s3://bucket/a/b/c/file.txt` but now to `s3://bucket/a//b///c/file.txt`.([#459](https://github.com/peak/s5cmd/pull/459))
+- `--endpoint-url` will not accept URL's without scheme such as `example.com`. Instead, it will give an error and ask for an url with a scheme; either `http://example.com` or `https://example.com` ([#496](https://github.com/peak/s5cmd/pull/496)).
```
igungor, line 7:
```suggestion
- `--endpoint-url` will not accept URLs without scheme such as `example.com`. Instead, it will give an error and ask for an url with a scheme; either `http://example.com` or `https://example.com` ([#496](https://github.com/peak/s5cmd/pull/496)).
```

command/app.go
```diff
@@ -120,6 +122,14 @@ var app = &cli.App{
 			stat.InitStat()
 		}
 
+		if endpointURL != "" {
+			if !strings.HasPrefix(endpointURL, "http") {
```
seruman, line 126: What do you think about using `net/url.Parse` then checking  `Url.Scheme`?

seruman: Oh, I missed that it was moved as is from `s3.go`.
https://github.com/peak/s5cmd/blob/914e701426e31a5819a422a4880bdbe03304f9df/storage/s3.go#L85-L87
Lets keep the existing behaviour then.

seruman (dismissed):

e2e/app_test.go
```diff
@@ -227,3 +227,57 @@ func TestInvalidLoglevel(t *testing.T) {
...
+			defer cleanup()
+
+			cmd := s5cmd("--endpoint-url", tc.endpointUrl)
+			result := icmd.RunCmd(cmd)
+
+			result.Assert(t, icmd.Expected{ExitCode: tc.expectedExitCode})
+
+			if tc.expectedError == nil && result.Stderr() == "" {
+				return
+			} else {
```
igungor, line 275: You can remove the else block.

igungor (approved):

seruman (approved):

PR #500, opened 2022-08-25
command: add extensive autocompletion support

### LINKED ISSUE #207: Auto completion for local files

### LINKED ISSUE #372: Bucket/key completion not supported?

kucukaslan:
Adds extensive auto-completion support to s5cmd for zsh, bash and pwsh. Including: 
- command completion
- flag completion
- local file/directory completion
- bucket name completion
- s3 key completion

The user will be given the instructions and the completion script when she called s5cmd with install-completion flag. The instructions and script will be written according to value of SHELL environment variable

Fixes #207 
Fixes #372


Kucukaslan:
- [x] command completion
- [x] flag completion
- [x] local file/directory completion 
> s5cmd cp <TAB>
- [x] bucket name completion
> s5cmd cp s3://<TAB>
- [x] s3 key completion
> s5cmd cp s3://bucket/<TAB>


In bash
- Completions can handle: `:`, `!`, `*`
- Completions CANNOT handle space characters, 
- Can handle `&`, `@` and `%` if the argument is quoted, that is 
`s5cmd cp "s3://bucket/ke@<TAB>`


In zsh
- Completions can handle: `:`, `?`, `*`
- - Can handle space characters if the argument is quoted, that is 
`s5cmd cp "s3://bucket/sp <TAB>`


ps. `<TAB>` given above always means pressing when the cursor is at that point.

Kucukaslan:
## Autocompletion
### General Notes
We are using the urfave/cli's autocompletion feature.
s5cmd has a special flag (`--generate-bash-completion`) which can only be the last parameter.
When this flag is given, s5cmd executes the appropriate [BashCompleteFunc](https://pkg.go.dev/github.com/urfave/cli/v2#BashCompleteFunc) to prepare autocompletion suggestions.

We provide autocompletion scripts for bash, pwsh, and zsh. The essence of all three script is that:
When a user pressed to TAB
- preserve current arguments, 
- execute s5cmd with `generate-bash-completion ` flag
- process/filter the autocompletion suggestions (and create local file/dir suggestions when appropriate) 
- provide the autocompletion suggestions to shell

### zsh
Its autocompletion behavior uses `:` as a seperator for `suggestion:description` pairs. So when SHELL[^sh] is zsh, s5cmd escapes[^esc] colon charactes.

### bash
#### Colon 
In bash autocompletion behaviour uses `:` as word seperator ("break"), unless the colon is removed from COMP_WORDBREAKS global variable. So the autocompletion suggestions should not include the colon and the text preceding the colon. That when `s3://` is going to be autocompleted to `s3://bucket` the suggestion should be `//bucket` instead of `s3://bucket`. 
If the COMP_WORDBREAKS does not contain colon, then the autocompletion suggestion should remain as is, i.e. as `s3://bucket`.
But s5cmd is unaware of the value of COMP_WORDBREAKS. So when SHELL is bash, s5cmd prepares and gives both styles.
Then our custom autocompletion script filters the appropriate one and forwards it to shell.

#### Using COMP_LINE & COMP_POINT
When an autocompletion request is made BASH sets calls the autocompletion script with some [environment variables](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#:~:text=COMP_CWORD,one%20possible%20completion). COMP_WORDS is one of them. It is an array of strings created from the original command to be completed. Normally it is [used](https://github.com/urfave/cli/blob/main/autocomplete/bash_autocomplete) in urfave/cli's autocompletion. However it splits words according to notorious  COMP_WORDBREAKS and colon characters are considered to be word breaks and seperate words.  It might be still tempting to use it and 'just concat when there is a colon'.  
But it is not possible for s5cmd to discern whether {'a', ':', 'b'} is originally `a:b` or `a :b` or `a: b` or `a : b` by only using COMP_WORDBREAKS. So instead of using the array then combining them, we will directly use the raw COMP_LINE string[^sub] which is the original command string written by user to the terminal.  
This, however, has some side effects. One of them is that it is not possible to discern whether TAB is pressed to complete the last argument or a new argument.  
That is, it cannot discern  `s5cmd cp s3://buck/obj<TAB>` and  `s5cmd cp s3://buck/obj <TAB>`, so it tries to complete  `s3://buck/obj` in both cases when we expect it to suggest a new argument in the second case. To overcome this, we have a simple hack in bash script, if there is a space just before the <TAB> then we pass an empty string argument to s5cmd. But this time it doesn't respect the actual space character in the  `s5cmd cp "s3://buck/obj <TAB>` despite the left quotation mark. I don't have a solution to it, and don't intent to solve that case.

### pwsh
Not doing anything special. It is the Powershell, BTW.

### Manjaro KDE: zsh vs bash
s5cmd reads the SHELL environment variable and prepares the autocompletions according to its value.
Normally the SHELL is same with the default shell of the terminal so it works fine. But, anecdotally, I know that in Manjaro KDE/Plasma the [Konsole](https://konsole.kde.org/) starts with the `zsh`, but the SHELL is still set to bash.
I don't handle this case nor similar cases that shell is different then the SHELL.

### autocompletion suggestions doesn't respect to the number of arguments
Autocompletions currently doesn't check number of arguments. For example `cat` command accepts only a single argument, but autocompletion continues to complete no matter how many args has been provided so far.
Although, it is possible to check number of arguments in s5cmd, I'm afraid that this won't stop the local file/dir suggestions. Since local completion is the fallback behaviour when s5cmd doesn't provide suggestions.

### Number of suggestions provided by s5cmd
I set it to 20 to avoid excessive API calls. But we can increase or remove the limit.

### Bug-like behaviour in bash
if `s3://` can be completed by both some prefix(es) (s3 directory) and key then autocompletion seems to show only the prefix in bash. I know this was too incomprehensible,  so let say there is a bucket named `bucket` having objects with keys `dir/f`, `die2/mc`, `boo` and `dododo`. Then `dododo` and `boo` are filtered out from the suggestions and the only suggestions became `dir/` and `die2/`.  This happens only in root level that is when completion requested for `s3://bucket/`. This behaviour does not occur in other shells. 

ps. This comment may be updated during the process, sorry for inconveniences.

[^sh]: The environment variable SHELL. More precisely the name of the binary that SHELL points to. e.g. when SHELL is set to /bin/bash, I say SHELL is bash. Also beware that, I use shell to mean the shell application used in terminal.
[^esc]: That is to replace every `:`. with `\:`
[^sub]: its substring up to COMP_POINT (cursor) to be more precise

command/auto_complete.go
```diff
@@ -0,0 +1,247 @@
...
+	`
+		cur="${COMP_WORDS[COMP_CWORD]}"
+		cmd="${COMP_LINE:0:$COMP_POINT}"` +
+
+	// if we want to complete the second argurment and we didn't started writing
+	// yet then we should pass an empty string as another argument. Otherwise
+	// the white spaces will be discarded and the program will make suggestions
+	// as if it is completing the first argument.
+	// Beware that the we want to pass empty string so we intentionally write
+	// as it is. Fixes of SC2089 and SC2090 are not what we want.
```
aykutfarsak, line 49:
```suggestion
	// if we want to complete the second argument and we didn't start writing
	// yet then we should pass an empty string as another argument. Otherwise
	// the white spaces will be discarded and the program will make suggestions
	// as if it is completing the first argument.
	// Beware that we want to pass an empty string so we intentionally write
	// as it is. Fixes of SC2089 and SC2090 are not what we want.
```

command/auto_complete.go
```diff
@@ -0,0 +1,247 @@
...
+
+const pwsh = `$fn = $($MyInvocation.MyCommand.Name)
+$name = $fn -replace "(.*)\.ps1$", '$1'
+Register-ArgumentCompleter -Native -CommandName $name -ScriptBlock {
+	param($commandName, $wordToComplete, $cursorPosition)
+	$other = "$wordToComplete --generate-bash-completion"
+		Invoke-Expression $other | ForEach-Object {
+		[System.Management.Automation.CompletionResult]::new($_, $_, 'ParameterValue', $_)
+		}
+}
```
aykutfarsak, line 83:
```suggestion
const pwsh = `$fn = $($MyInvocation.MyCommand.Name)
$name = $fn -replace "(.*)\.ps1$", '$1'
Register-ArgumentCompleter -Native -CommandName $name -ScriptBlock {
	param($commandName, $wordToComplete, $cursorPosition)
	$other = "$wordToComplete --generate-bash-completion"
	Invoke-Expression $other | ForEach-Object {
		[System.Management.Automation.CompletionResult]::new($_, $_, 'ParameterValue', $_)
	}
}
```

aykutfarsak (changes requested):

command/app.go
```diff
@@ -154,13 +154,11 @@ var app = &cli.App{
 	},
 	Action: func(c *cli.Context) error {
 		if c.Bool("install-completion") {
-			if cmpinstall.IsInstalled(appName) {
-				return nil
-			}
+			shell := os.Getenv("SHELL")
```
seruman, line 157: `SHELL` environment read second time down below the autocompletion flow. What do you think creating a new type an storing shell variable in it? We can also keep autocompletion functions as its method. It would also make testing easier IMHO, we might write test just for this type's behaviour.

command/auto_complete.go
```diff
@@ -0,0 +1,247 @@
...
+	// Beware that the we want to pass empty string so we intentionally write
+	// as it is. Fixes of SC2089 and SC2090 are not what we want.
+	// see also https://www.shellcheck.net/wiki/SC2090
+	`
+		[ "${COMP_LINE:COMP_POINT-1:$COMP_POINT}" == " " ] \
+			&& cmd="${cmd} \"\"" ` +
+
+	// execute the command with '--generate-bash-completion' flag to obtain
+	// possible completion values for current word.
+	// ps. SC2090 is not wanted.
```
seruman, line 57: I think it would be more easier to read if we move comments to top of the declaration.

command/auto_complete.go
```diff
@@ -0,0 +1,247 @@
...
+	`
+		opts=$($cmd --generate-bash-completion)` +
+
+	// prepare completion array with possible values and filter those do not start with cur.
+	// if no completion is found then fallback to default completion of shell.
+	`
+
+		while IFS='' read -r line; do COMPREPLY+=("$line"); done < <(compgen -o bashdefault -o default -o nospace -W "${opts}" -- "${cur}")
```
seruman, line 65: Please avoid one liners, Bash is already hard to read -IMHO- this makes it harder.

command/auto_complete.go
```diff
@@ -0,0 +1,247 @@
...
+			return
+		}
+
+		cli.DefaultCompleteWithFlags(cmd)(ctx)
+	}
+}
+
+// it returns a complete function which prints the argument, itself, which is to be completed.
+// If the argument is empty string it uses the defaultCompletions to make suggestions.
+func ineffectiveCompleteFnWithDefault(cmd *cli.Command, defaultCompletions ...string) func(ctx *cli.Context) {
```
seruman, line 122: WDYM by ineffective? Please give it a name that indicates what it does.

command/auto_complete.go
```diff
@@ -0,0 +1,247 @@
...
+		instructions = "# To enable autocompletion you should save the following" +
+			" script to a file named \"s5cmd.ps1\" and execute it.\n# To persist it" +
+			" you should add the path of \"s5cmd.ps1\" file to profile file " +
+			"(which you can locate with $profile) to automatically execute \"s5cmd.ps1\"" +
+			" on every shell start up."
+	} else {
+		instructions = "# We couldn't recognize your SHELL \"" + baseShell + "\".\n" +
+			"# Shell completion is supported only for bash, pwsh and zsh." +
+			"# Make sure that your SHELL environment variable is set accurately."
+	}
```
seruman, line 215: You may use a `switch` statement here.

command/app.go
seruman: I was thinking something like -names are just placeholders, ignore-;
``` go
type autoComplete struct {
	cmd *cli.Command
	shell string
	client storage.Storage
	out io.Writer
}

(a autoComplete) bashCompleteFn() func(ctx *cli.Context) {
	// ...
}
```

With something like this, we might pass fake dependencies to it and write unit tests to its methods, along side with `e2e` tests.

command/auto_complete.go
```diff
@@ -0,0 +1,247 @@
...
+				return
+			}
+			printS3Suggestions(ctx, arg, isOnlyBucket)
+			return
+		}
+
+		cli.DefaultCompleteWithFlags(cmd)(ctx)
+	}
+}
+
```
seruman, line 119: With the comment above making these functions a method of a new type, what do you think writing individual tests for these? They're quite tricky to grasp.

seruman: Maybe separating logic for `bash` from others -in all autocomplete functions- would help as it is the one that requires tinkering

command/auto_complete.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+		if obj.Err != nil {
+			return
+		}
+		fmt.Println(formatSuggestionForShell(obj.URL.Absolute(), argToBeCompleted))
+		i++
+	}
+}
+
+func installCompletionHelp(shell string) {
```
seruman, line 193: It does not really install right? 😆

e2e/util_test.go
```diff
@@ -394,6 +394,16 @@ func withWorkingDir(dir *fs.Dir) func(*icmd.Cmd) {
...
+	return func(cmd *icmd.Cmd) {
+		if i := indexSlice(cmd.Env, key+"=", strings.HasPrefix); i > 0 {
+			cmd.Env[i] = key + "=" + value
+		} else {
+			cmd.Env = append(cmd.Env, key+"="+value)
+		}
+	}
+}
+
 type compareFunc func(string) error
```
seruman, line 407: `gotest.tools/v3/icmd` already defines `WithEnv` option;
https://pkg.go.dev/gotest.tools/v3/icmd#WithEnv

e2e/util_test.go
```diff
@@ -671,3 +681,36 @@ func (l *fixedTimeSource) Advance(by time.Duration) {
...
+
+	if shouldSort {
+		sort.Strings(expected)
+	}
+
+	for i, e := range expected {
+		m[i] = equals(e)
+	}
+	return m
+}
```
seruman, line 707: Instead of this function, you may define `expected` in `TestCompletionFlag` as `map[int]compareFunc` just like the other tests. I think it would make more easier to read the tests as the assertion logic is not that complex to hide it.

e2e/auto_complete_test.go
```diff
@@ -0,0 +1,238 @@
...
+				keysToS3URL("s3://", bucket,
+					"file0.txt",
+					"file1.txt",
+					"filedir/",
+					"dir/"),
+				keysToS3URL("//", bucket,
+					"file0.txt",
+					"file1.txt",
+					"filedir/",
+					"dir/")...),
```
seruman, line 134: You may just repeat yourself. More explicit the tests more easier to read IMHO, you may just focus on what the tested unit does instead decoding how inputs/outputs are prepared.

storage/url/url.go
```diff
@@ -326,6 +326,10 @@ func (u *URL) SetRelative(base *URL) {
 
 // Match reports whether if given key matches with the object.
 func (u *URL) Match(key string) bool {
+	if u.filterRegex == nil {
+		return false
+	}
```
seruman, line 331: Why this was necessary in the scope of this PR?

seruman (changes requested):

command/auto_complete.go
Kucukaslan: I mean it doesn't actually predict anything but merely returns the argument itself (or some default values) to override shell's default autocompletion. May be `nonpredictiveCompleteFnWithDefault` can be a better name (or worse?)

command/auto_complete.go
Kucukaslan: No it doesn't, it only prints the corresponding instructions and shell script 😅
Since the flag name was `install-completion` I named the function as `installCompletionHelp`, not a good naming.
I  renamed it to `printAutocompletionInstructions` now.

storage/url/url.go
Kucukaslan: At some time the s5cmd panicked due to nil pointer.

As far as I remember,  I enabled `raw` option with autocompletion URL's ignoring the value of actual raw `flag`. Since the raw URL does not have `filterRegex`, s3.List method panicked here (at URL.Match) method.
Now, I stopped setting 'raw' option to true.

command/app.go
Kucukaslan: I'm a bit doubtful about the benefits of such a type. You give two reasons for that type:
1) avoid repetitive calls to os.Getenv
I agree it is not neat that `formatSuggestionForShell` calls os.Getenv at each execution. Would using a global(? but private) variable in command package to store shell( and baseShell) values fine for you?

2) simplify writing tests
It seems to be easier (and safer) to manipulate the `SHELL` environment in comparison to current functions (which would require us to change it with os.Setenv which will change the env. var. just for the s5cmd but might be problematic to use when tests concurrently changes it). I'm a bit confused about how to organize these tests and not sure is it really necessary to have a type to write these tests.


 A) Do you suggest `completeFn` to be single  method that would be called for each command and return the appropriate function or do you suggest that the current functions are going to be converted to autoComplete's methods?
If you mean the former I think it would be bad for autoComplete type (need) to be aware of the commands and their expectations from completion.
If you mean the latter, then, each command will call only a method of the autoComplete, so I think it would be better if it just call the function.

command/auto_complete.go
Kucukaslan: I initially thought of these comments as the inline comments of the bash script. So each part if the comment is explaining the line(s) following it. But I converted these comments from `bash comments` to `go comments` to hide them from the end user.

command/auto_complete.go
Kucukaslan: I saw a sample usage of the while loop as a single liner. I did not dare to change it by the fear of breaking it 😅
I've changed it to following
```bash
while IFS='' read -r line;
do
	COMPREPLY+=("$line");
done \
< <(compgen -o bashdefault -o default -o nospace -W "${opts}" -- "${cur}")
```
but not sure whether the last line should've been in the same line with `done` or not

e2e/util_test.go
Kucukaslan: Yes, but it is not equivalent to what this method is doing. The `icmd.WithEnv` replaces the `Env` field with the arguments. But what I want is to change/set an environment variable (`SHELL` in my case) and leave the rest untouched. The `icmd.WithEnv` is not helpful to do it.

command/app.go
seruman: We discussed face to face and decided that a new type is unnecessary and about the unit tests; auto completion requires a remote client and there are already end to end tests with fake S3 server.

command/auto_complete.go
```diff
@@ -0,0 +1,246 @@
...
+			shell := filepath.Base(os.Getenv("SHELL"))
+			printS3Suggestions(c, shell, client, u, arg, isOnlyBucket)
+			return
+		}
+	}
+}
+
+// it returns a complete function which prints the argument, itself, which is to be completed.
+// If the argument is empty string it uses the defaultCompletions to make suggestions.
+func constantCompleteWithDefault(shell, arg string, defaultCompletions ...string) {
```
seruman, line 121:
```suggestion
// constantCompleteWithDefault returns a complete function which prints the argument, itself, which is to be completed.
// If the argument is empty string it uses the defaultCompletions to make suggestions.
func constantCompleteWithDefault(shell, arg string, defaultCompletions ...string) {
```

command/auto_complete.go
```diff
@@ -0,0 +1,246 @@
...
+		return escapeColon(suggestion)
+	default:
+		return suggestion
+	}
+}
+
+// replace every colon : with \:
+func escapeColon(str ...interface{}) string {
+	return strings.ReplaceAll(fmt.Sprint(str...), ":", `\:`)
+}
```
seruman, line 225: What do you think about removing this function and moving `string.ReplaceAll` call to where the function is called as it is just used for `zsh`?

command/auto_complete.go
```diff
@@ -0,0 +1,246 @@
...
+	l := args.Len()
+
+	if l > 0 {
+		arg = args.Get(l - 1)
+	}
+
+	// argument may start with a quotation mark, in this case we want to trim that before
+	// checking if it has prefix s3://
+	// Beware that we only want to trim the first char, not all of the leading
+	// quotation marks, because those quotation marks may be actual charactes.
```
seruman, line 239: nitpick/typo: let's limit length of the comments to ~80 chars per lone.
```suggestion
	// argument may start with a quotation mark, in this case we want to trim
	// that before checking if it has prefix 's3://'.
	// Beware that we only want to trim the first char, not all of the leading
	// quotation marks, because those quotation marks may be actual characters.
```

seruman (approved):
🚀  Thank you!

aykutfarsak (approved):

Kucukaslan:
> 🚀 Thank you!
My pleasure. Thanks a lot for help and suggestions!

command/auto_complete.go
Kucukaslan: For the record with https://github.com/Kucukaslan/s5cmd/commit/578f64c79aa9d76206c5fe90e8435a6afad91fdb this function is split to two parts and the auto-completion function is named `constantCompleteWithDefault`.

PR #505, opened 2022-09-02
storage/url: allow wildcards to match new line characters

kucukaslan:
This commit adds `s` flag [^1] to URL's filter Regexp to "let `.` match `\n`".

[^1]: https://github.com/google/re2/wiki/Syntax#:~:text=text%20(default%20false)-,s,-let%20.%20match


command/exclude.go
```diff
@@ -20,7 +20,7 @@ func createExcludesFromWildcard(inputExcludes []string) ([]*regexp.Regexp, error
 	for _, input := range inputExcludes {
 		if input != "" {
 			regexVersion := wildCardToRegexp(input)
-			regexpCompiled, err := regexp.Compile(regexVersion)
```
ilkinulas, line 23: @Kucukaslan thank you for debugging this and creating the PR. I have two cosmetic comments: 
1. Regex pattern is built in `wildCardToRegexp` function. Prefixing the `patternRegex` in that function with `(?s)` will be more convenient. We can also drop a comment about why we are using this flag.

2. `regexp.Compile("(?s)", pattern)` function calls can be encapsulated in its own method where we prepend `(?s)` with a an explanation. In this way the regex flag `(?s)`  will not be visible in multiple places in the code. 

What do you think?

Kucukaslan: Thanks for review.

As a first note, I assume these two options are alternative options. So my further comments will be about which one is preferable and why.

### On option 1
Although the `regexp` that `wildCardToRegexp` prepared does not work accurately with new line character, I'm rather hesitated to add `(?s)` in that method. Because the `(?s)` is, at least technically, a flag and my concern is that would specifying flags be an expected behaviour of `wildCardToRegexp` method (or adding flags is the responsibility of the caller of that method)?

### On option 2
I'm not sure what kind of function do you mean.
a) If you mean a function that only takes pattern and returns the new string with `(?s)` it's doable and I'm fine with it.
b) But if you mean to replace whole `regexp.Compile("(?s)", pattern)` then I've a few concerns:
1. In the tests we use its other variant `regexp.MustCompile` which panics instead of returning error. Then we should either write seperate method for it too[^update], or same method should take another parameter "isMustCompile".
2. If we use the same wrapper method for both Compile and MustCompile the surrounding method[^sur] still needs to receive the error so it should return both the regex end the error even tough MustCompile won't return an error.

My preference would be Option 2.a if we want to handle both Compile and  MustCompile together.

[^sur]: https://github.com/peak/s5cmd/blob/acb97e9fa4d186ee47307f011fbe03f90b3e2a57/storage/url/url.go#L202-L226
[^update]: Or do you think changing wildCardToRegexp method will already fix this.

Kucukaslan: I've extracted the regular expression preperation to three util functions: WildCardToRegexp, MatchFromStartToEnd and AddNewLineFlag.

Since the exclude.go was in `command` package but the `setPrefixAndFilter` was in `storage/url` they must be exported.
Neither of the packages seemed a correct place, so I placed them to `strutil` package. It might be better to rename `strutil` to, more generic name, `util`.


'setPrefixAndFilter' preferred `*?` over `*` which was preferred by `exclude.go`. As far as I understood, they won't affect the results since we always use full text match. But it may have some effect in the speed.

strutil/strutil.go
```diff
@@ -40,3 +42,23 @@ func JSON(v interface{}) string {
...
+func AddNewLineFlag(pattern string) string {
+	return "(?s)" + pattern
+}
+
+// WildCardToRegexp converts a wildcarded expresiion to equivalent regular expression
+func WildCardToRegexp(pattern string) string {
+	patternRegex := regexp.QuoteMeta(pattern)
+	patternRegex = strings.Replace(patternRegex, "\\?", ".", -1)
+	patternRegex = strings.Replace(patternRegex, "\\*", ".*", -1)
+	//patternRegex = "^" + patternRegex + "$"
```
ilkinulas, line 57:
```suggestion
```

ilkinulas (dismissed):

storage/url/url_test.go
```diff
@@ -70,7 +71,7 @@ func TestNew(t *testing.T) {
 				Prefix:    "key",
 				Delimiter: "/",
 			},
-			wantFilterRe: regexp.MustCompile(`^key.*$`).String(),
+			wantFilterRe: regexp.MustCompile(strutil.AddNewLineFlag(`^key.*$`)).String(),
```
sonmezonur, line 74: IMHO instead of using `strutil.AddNewLineFlag`, writing all regex will be much more clear (`(?s)^key.*$`). If I change the logic of strutil.AddNewLineFlag, tests will not fail. Not sure if this is something we want.

sonmezonur (approved):

ilkinulas (approved):

PR #547, opened 2023-02-17
Add a -cpuprofile flag

nelhage:
I'm not sure if this is something you want to merge, but I found it useful just now so I figured I'd propose it.


igungor:
@nelhage Thank you for the PR. I agree that it is useful from time to time, but it's not something we'd like to add to the current flag set. I hope it's OK for you.

PR #558, opened 2023-04-04
Update README.md

tweep:
Added example on how to use the new --profile option in a cp operation. I found it a bit misleading and it took me some time to figure out the correct syntax. hope this helps peeps to use the new option. Thx for adding it , great tool.


szeitlin (approved):
+1 to update the README, I only knew this syntax because I used s3 CLI a lot, and this is the same.

igungor:
@tweep @szeitlin Hey, I just pushed a PR to clarify more about why profile selection and credentials work. How does that look?

igungor:
#570

PR #565, opened 2023-05-07
Fix for ISSUE-564

monobaila:
* Ensure "fail fast" during sync if any AWS Errors come back, if we don't have accurate list ouputs from source/dest it's safer to bail over treating them as empty.
* Add E2E test, currently for failing fast if target bucket doesn't exist. I'm not able to add a test for original failure case in bug report as mock framework doesn't seem to support this?

### DIFF
--- a/command/sync.go
+++ b/command/sync.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/hashicorp/go-multierror"
 	"github.com/urfave/cli/v2"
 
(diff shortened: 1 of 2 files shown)


ahmethakanbesel:
Hi monabaila,

Thank you for your interest in `s5cmd` and your contribution to resolving this issue. The issue has been fixed in version 2.2.0. We have decided to take a different approach to the problem, rather than checking for errors at the end of the listing process.

You are correct that the mock s3 server used in the tests does not support bucket policies. As a result, we can only test for the `NoSuchBucket` error.

Please let me know if you have any other questions.

igungor:
Thank you for the PR @monobaila. As @ahmethakanbesel explained above, the issue is fixed with https://github.com/peak/s5cmd/pull/597. I'm closing your PR.

PR #567, opened 2023-05-15
Try Getting Time from Server for List Calls

BrentSouza:
We have noticed that `s5cmd ls` sometimes fails to list objects immediately after they are written. This seems to be related to clock skew since `s5cmd` gets the current time from the client and compares it against each object's lastModified date. This is detailed a bit in issue #531

In our case, S3 servers seem to be a couple seconds ahead of our servers (which are synced with NTP servers so not sure where the issue lies). While troubleshooting, we did notice that the Date header from S3's http responses lined up with the future times we saw in the lastModified metadata.

This patch will request the Date header from the response and use it for the timestamp. It will fallback to the client's current time if the header is not present or in the event that the date cannot be parsed.

I've modified the tests to check for both scenarios and locally all unit test are working. Please let me know if there are any additional checks you'd like me to run.


mwkaufman:
I built this locally and it did now successfully return the future-dated (rounded up to the next whole second) object in our environment.

storage/s3.go
```diff
@@ -330,10 +330,20 @@ func (s *S3) listObjectsV2(ctx context.Context, url *url.URL) <-chan *Object {
 
 				objectFound = true
 			}
+
 			// track the instant object iteration began,
 			// so it can be used to bypass objects created after this instant
 			if now.IsZero() {
-				now = time.Now().UTC()
+				if serverDate != "" {
+					n, err := http.ParseTime(serverDate)
```
ilkinulas, line 338: `http.ParseTime` function is not able to accurately represent the last modified time of an S3 object, because the latter has a much higher precision. This is causing the tests to fail. 

For example:
- Parsed time in header : 2023-07-01 19:43:58 +0000 UTC
- Last Modified time of S3 object: 2023-07-01 19:43:58.058 +0000 UTC

I will let you know if I am able to find a solution to the time [comparison](https://github.com/peak/s5cmd/pull/567/files#diff-22a91901acea81cf09b58954babe267e1a5b503dbce7c37ab0df11ebfc86b0b1R355) issue.

PR #574, opened 2023-06-20
readme: update build from source instruction

### LINKED ISSUE #568: Update README re: installing on linux; update to use `go install` and fix version tagging?
`sudo yum install golang` works

but `go get github.com/peak/s5cmd` doesn't work because `go get` has been deprecated in favor of `go install`, which did not work for me on my first try. 

Currently trying `go install github.com/peak/s5cmd@latest` but it's slow and I can't tell if it's working or not. I volunteer to make this update if/when I can get it to work.

igungor:
Fixes #568

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ ps.  Quoted from [s5cmd feedstock](https://github.com/conda-forge/s5cmd-feedstoc
 You can build `s5cmd` from source if you have [Go](https://golang.org/dl/) 1.17+
 installed.
 
-    go install github.com/peak/s5cmd/v2
+    go install github.com/peak/s5cmd/v2@master
 
 ⚠️ Please note that building from `master` is not guaranteed to be stable since
 development happens on `master` branch.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,7 +94,7 @@ ps.  Quoted from [s5cmd feedstock](https://github.com/conda-forge/s5cmd-feedstoc
 You can build `s5cmd` from source if you have [Go](https://golang.org/dl/) 1.17+
 installed.
 
-    go get github.com/peak/s5cmd
+    go install github.com/peak/s5cmd/v2
```
seruman, line 97: I think the path needs to be something like `github.com/peak/s5cmd/v2@master` or `github.com/peak/s5cmd@master` ;
```bash
; go version
go version go1.20.5 darwin/amd64

; go install github.com/peak/s5cmd/v2
go: 'go install' requires a version when current directory is not in a module
        Try 'go install github.com/peak/s5cmd/v2@latest' to install the latest version
```

Omitting version or using `@latest` would try to build latest versioned release AFAIK;
> https://go.dev/ref/mod#version-queries
> The string latest, which selects the highest available release version.

igungor: PTAL

seruman (approved):

ilkinulas (approved):

PR #581, opened 2023-07-03
command: update misleading help descriptions

### LINKED ISSUE #555: `s5cmd` help is misleading
`s5cmd` help provides Examples:

  
```
Examples:
...
  06. Upload matching files to S3 bucket
     > s5cmd cp dir/*.gz s3://bucket/
...
```

But the above example does not work unless the text including the wildcard character (*) is enclosed with quotation marks.

This works instead:

```
Examples:
  06. Upload matching files to S3 bucket
     > s5cmd cp "dir/*.gz" s3://bucket/
```

denizsurmeli:
This PR updates the help messages for the commands that takes arguments containing wildcards. Since the expansion of wildcards are evaluated by the shell and not by the `s5cmd` beforehand, there is no way for app to prevent the expansion. We urge users to consider escaping from globbing by wrapping the related argument to quotes. Obviously, the escaping method can be different for every shell implementation. 

Fixes #555

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -45,7 +45,7 @@ Examples:
 		 > s5cmd {{.HelpName}} s3://bucket/prefix/object.gz myobject.gz
 
 	03. Download all S3 objects to a directory
-		 > s5cmd {{.HelpName}} s3://bucket/* target-directory/
+		 > s5cmd {{.HelpName}} "s3://bucket/*" target-directory/
 
 	04. Download an S3 object from a public bucket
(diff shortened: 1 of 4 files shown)


sonmezonur (approved):

igungor:
@denizsurmeli congrats on your first PR!

We try to use `Fixes #ISSUE_NO` in a separate line, usually after the PR/commit description. Please check out https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword.

seruman (approved):

Issue #541, opened 2023-01-12
Sync doesn't allow whitspaces in --cache-control


7rulnik:
I guess this is a problem with parsing and passing `--cache-control` option.

This example doesn't work:

```sh
s5cmd --dry-run sync --cache-control 'public, max-age=31536000, immutable' ./dist/selene-static/ s3://selene-static/selene-static-s5cmd/
```

But this one does:

```sh
s5cmd --dry-run sync --cache-control 'public,max-age=31536000,immutable' ./dist/selene-static/ s3://selene-static/selene-static-s5cmd/
```

And for `cp` command both variants work

PR #582, opened 2023-07-03
command: fix file overriding when download fails (#479)

### LINKED ISSUE #479: Local file is lost even if the download fails
### Issue & steps to reproduce
```
$ echo "Example text." > tmp
$ cat tmp
Example text.
$ s5cmd cp s3://nonexistentbucket/key tmp
ERROR "cp s3://nonexistentbucket/key tmp": InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records. status code: 403, request id: PAYWDH7HHVPAHC8G, host id: cnI/cBhYZKKJ159f8mMs4KHOn1+jYtvfpYnXgMuKMRd9pZ10FBdi/cGuVxyr+iwrKbk2kP7Opx4=
$ cat tmp
cat: tmp: No such file or directory
```
We would expect the local file to remain untouched since the download operation did not even started.

### Proposed Solution
Instead of creating the file before the download request a custom [WriterAt](https://pkg.go.dev/io#WriterAt), which will create the file once the first write request came, can be used. Following snippets are going to change:
https://github.com/peak/s5cmd/blob/123b1c7fc9c614aa214a6795468fa140a38ad05e/command/cp.go#L511-L517
https://github.com/peak/s5cmd/blob/0431f503d99953e1809bf0e86d73750c0c1f561e/storage/fs.go#L208-L214

ahmethakanbesel:
Resolves #479 
Local files used to be overwritten even if downloads failed.
Solved it by creating a temporary file and renaming it with the original filename after completing the download successfully.

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -545,12 +545,12 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) 
 		}
 		return err
 	}
-	defer file.Close()
 
 	if c.deleteSource {
 		_ = srcClient.Delete(ctx, srcurl)
 	}
 
+	file.Close()
 	os.Rename(tempurl.Absolute(), dsturl.Absolute())
 
 	msg := log.InfoMessage{
--- a/command/cp.go
+++ b/command/cp.go
@@ -549,7 +549,10 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) 
 		_ = srcClient.Delete(ctx, srcurl)
 	}
 
-	dstClient.Rename(file, dsturl.Absolute())
+	err = dstClient.Rename(file, dsturl.Absolute())
+	if err != nil {
+		return err
+	}
 
 	msg := log.InfoMessage{
 		Operation:   c.op,
--- a/command/cp.go
+++ b/command/cp.go
@@ -528,7 +528,9 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) 
 		return err
 	}
 
-	file, err := dstClient.Create(dsturl.Absolute())
+	var tempurl = *dsturl
+	tempurl.Path = tempurl.Path + ".tmp"
+	file, err := dstClient.Create(tempurl.Absolute())
 	if err != nil {
 		return err
 	}
@@ -537,9 +539,9 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) 
 	if err != nil {
 		// file must be closed before deletion
 		file.Close()
-		dErr := dstClient.Delete(ctx, dsturl)
+		dErr := dstClient.Delete(ctx, &tempurl)
 		if dErr != nil {
-			printDebug(c.op, dErr, srcurl, dsturl)
+			printDebug(c.op, dErr, srcurl, &tempurl)
 		}
 		return err
 	}
@@ -548,6 +550,8 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL) 
 	if c.deleteSource {
 		_ = srcClient.Delete(ctx, srcurl)
 	}
+
+	os.Rename(tempurl.Absolute(), dsturl.Absolute())
 
 	msg := log.InfoMessage{
 		Operation:   c.op,
--- a/command/cp.go
+++ b/command/cp.go
@@ -5,12 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math/rand"
 	"mime"
 	"net/http"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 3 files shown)


sonmezonur (dismissed):

sonmezonur:
@ahmethakanbesel Thanks for your PR. Could you please link your PR to the issue? (when PR merged it will automatically close the referenced issue) ([docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

command/cp.go
```diff
@@ -528,7 +528,9 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL)
 		return err
 	}
 
-	file, err := dstClient.Create(dsturl.Absolute())
+	var tempurl = *dsturl
+	tempurl.Path = tempurl.Path + ".tmp"
+	file, err := dstClient.Create(tempurl.Absolute())
```
sonmezonur, line 533: if the file (`$file.tmp`) exists (somehow, probably won't happen), the command will fail. Could you please try to use `ioutil.Tempfile()` which also handles file name collisions?

seruman: +1, this would also keep users's cwd clean.

command/cp.go
```diff
@@ -528,7 +528,9 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL)
 		return err
 	}
 
-	file, err := dstClient.Create(dsturl.Absolute())
+	dstPath := filepath.Dir(dsturl.Absolute())
+	dstFile := filepath.Base(dsturl.Absolute())
+	file, err := dstClient.CreateTemp(dstPath, dstFile)
```
sonmezonur, line 533:
```suggestion
	file, err := dstClient.CreateTemp("", dstFile)
```

lets do not give any path to keep cwd clean as @seruman suggested

command/cp.go
```diff
@@ -537,18 +539,20 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL)
...
 		return err
 	}
-	defer file.Close()
 
 	if c.deleteSource {
 		_ = srcClient.Delete(ctx, srcurl)
 	}
 
+	file.Close()
+	dstClient.Rename(file, dsturl.Absolute())
```
sonmezonur, line 554: Please handle error case

command/cp.go
```diff
@@ -537,18 +539,20 @@ func (c Copy) doDownload(ctx context.Context, srcurl *url.URL, dsturl *url.URL)
...
 		}
 		return err
 	}
-	defer file.Close()
 
 	if c.deleteSource {
 		_ = srcClient.Delete(ctx, srcurl)
 	}
 
+	file.Close()
```
sonmezonur, line 553: we are closing the file for both error and success cases. can we move this block after this line?https://github.com/peak/s5cmd/blob/15fdef5620567a1ec379ce024d0cb10985d72d33/command/cp.go#L538

thus we can remove line #541

```go
       size, err := srcClient.Get(ctx, srcurl, file, c.concurrency, c.partSize)
       // close after read
       file.Close()

       if err != nil {
		// file must be closed before deletion
		dErr := dstClient.Delete(ctx, &url.URL{Path: file.Name(), Type: dsturl.Type})
		if dErr != nil {
			printDebug(c.op, dErr, srcurl, dsturl)
		}
		return err
	}
```

command/cp.go
ahmethakanbesel: Creating the temporary file in cwd can improve performance when the temp folder and target path are on different drives or partitions. Otherwise, it will need to make a copy of the temp file, which can take a long time for large files.

command/cp.go
ahmethakanbesel: Do we really need to handle file closing errors? The rest of the code doesn't handle errors for `file.Close()`. If I do, it will look like this:

```code
size, err := srcClient.Get(ctx, srcurl, file, c.concurrency, c.partSize)
err2 := file.Close()
if err2 != nil {
	return err2
}
if err != nil {
	dErr := dstClient.Delete(ctx, &url.URL{Path: file.Name(), Type: dsturl.Type})
	if dErr != nil {
		printDebug(c.op, dErr, srcurl, dsturl)
	}
	return err
}
```

command/cp.go
ahmethakanbesel: Yes, we can call `file.Close()` only once. I will move it to below of that line.

command/cp.go
sonmezonur: Not need to handle closing errors better to handle `Rename()`.

sonmezonur (dismissed):

e2e/cp_test.go
```diff
@@ -4140,3 +4140,32 @@ func TestDeleteFileWhenDownloadFailed(t *testing.T) {
...
+	defer workdir.Remove()
+
+	// It is going try downloading a nonexistent file from the s3 so it will fail.
+	// In this case we don't expect to have a local file will be overwritten.
+	cmd := s5cmd("cp", "s3://"+bucket+"/"+filename, filename)
+	result := icmd.RunCmd(cmd, withWorkingDir(workdir))
+
+	result.Assert(t, icmd.Expected{ExitCode: 1})
+
+	// assert local filesystem does not have the inital file
```
ilkinulas, line 4168:
```suggestion
	// assert initial file is untouched
```

ilkinulas (approved):

sonmezonur (approved):

Issue #584, opened 2023-07-03
Latest goreleaser deprecates archives.replacements


JeremyGHutchins:
I needed to build the code for ppc64le, but it seems that the latest goreleaser does not support the `archives.replacements` as it is removed in goreleaser 1.19.  I get the following error when running `goreleaser --debug build`...

```
   ⨯ build failed after 0s                    error=yaml: unmarshal errors:
  line 30: field replacements not found in type config.Archive
```
See [here](https://goreleaser.com/deprecations/#archivesreplacements) for details.  Should be a relatively straight forward fix, but not sure if it affects anything else.

PR #585, opened 2023-07-03
Update .goreleaser.yaml to correct archives.replacements deprecation

JeremyGHutchins:
Adds the ppc64le architecture to the .goreleaser.yaml and also fixes the deprecation issue with archives.replacements. See issue: #584

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -16,18 +16,22 @@ builds:
       - amd64
       - arm
       - arm64
+      - ppc64le
     ignore:
       - goos: darwin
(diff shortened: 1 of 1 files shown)


terryrankine:
For anyone hitting this with a current goreleaser: the working v2 config (archives.name_template, `version: 2`, `skip_upload` for forks) is in terryrankine/s5cmd at .goreleaser.yml; releases build from it. `brews` itself is now deprecated in favour of `homebrew_casks` — that's a change to the tap's formula type and is upstream's call.

Issue #522, opened 2022-10-10
Read local object from STDIN 


snebel29:
Hi,
This topic has been discussed before, at least in https://github.com/peak/s5cmd/issues/182.

I am just weighting (once more) the apetite for s5cm to provide this feature, to get a contribution and if the answer is positive, what the ideal approach could be.

Cheers.

igungor:
Hey,

There's an open PR https://github.com/peak/s5cmd/pull/587 to resolve this issue.

PR #587, opened 2023-07-06
command: add pipe command (#182)

### LINKED ISSUE #182: Add option to stdin to cp

ahmethakanbesel:
Resolves #182 and #522
Added `pipe` command to upload `stdin` output to a bucket.

Example usage:
`echo "content" | s5cmd pipe s3://bucket/file.txt`


storage/stdin.go
```diff
@@ -0,0 +1,26 @@
+package storage
+
+import (
+	"io"
+	"os"
+)
+
+type StdinReader interface {
```
seruman, line 8: Do we really need `io.ReadCloser` wrapper and `storage.StdinReader` abstraction for `os.Stdin`? It is used as `io.Reader` in pipe command anyways as `storage.Storage.Put` interface accepts an `io.Reader`. Par with `cat` command, I think we can use `os.Stdin` as is;
https://github.com/peak/s5cmd/blob/74d21bb4f022edb0bfb69f412c4baa69153ba8c2/command/cat.go#L106

command/pipe.go
```diff
@@ -0,0 +1,360 @@
...
+	deleteSource bool
+
+	// flags
+	noClobber             bool
+	storageClass          storage.StorageClass
+	encryptionMethod      string
+	encryptionKeyID       string
+	acl                   string
+	forceGlacierTransfer  bool
+	ignoreGlacierWarnings bool
```
seruman, line 153: In `cp`, Glacier options are only valid for source objects;
https://github.com/peak/s5cmd/blob/74d21bb4f022edb0bfb69f412c4baa69153ba8c2/command/cp.go#L410-L417
https://github.com/peak/s5cmd/blob/74d21bb4f022edb0bfb69f412c4baa69153ba8c2/command/cp.go#LL358-L364


I think those can be omitted here as there aren't any objects, just stdin.

command/pipe.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+
+	msg := log.InfoMessage{
+		Operation:   c.op,
+		Source:      nil,
+		Destination: dsturl,
+		Object: &storage.Object{
+			StorageClass: c.storageClass,
+		},
+	}
```
seruman, line 291:
```bas
; echo 'stuff' | ./s5cmd pipe s3://bucket/selman/stuff-test.txt
pipe <nil> s3://bucket/selman/stuff-test.txt
```

Looks like `log.InfoMessage.String()` does not handle this case.

seruman (changes requested):

hsource:
I tested this in production for uploading a 500 GB file that was the output of tar and it seemed to work well! Just wanted to add my 2 cents here.

I did this by

1. Forking the branch and running `go install github.com/wanderlog/s5cmd/v2@v2.1.0-wanderlog.1` to install it
2. Running this command:

```sh
tar -cf - "$INPUT" | zstd -1 --stdout | \
  ~/go/bin/s5cmd --endpoint-url "$S3_ENDPOINT" --credentials-file "$S5CMD_CREDENTIALS_FILE" \
    pipe --part-size "$PART_SIZE_MB" "$S3_DESTINATION"
```

Verifying the command completed successfully

storage/stdin.go
ahmethakanbesel: Yes, you're right. I simplified it but still we can't use `os.stdin` directly due to limitations in AWS SDK (it throws illegal seek error).
https://github.com/peak/s5cmd/pull/587/commits/19d2675fe8f27d9f92ebfb5fda5aeb1e16cc3d91

command/pipe.go
ahmethakanbesel: Yes, we don't need them. I have removed.
https://github.com/peak/s5cmd/pull/587/commits/3b6d0a8dfb260b215a6601a6ee4f1269b6c35c54

command/pipe.go
ahmethakanbesel: Done. https://github.com/peak/s5cmd/pull/587/commits/3c5c712588edada2b2813559beb9cb05c317868c

command/pipe.go
```diff
@@ -0,0 +1,354 @@
...
+	contentType := mime.TypeByExtension(filepath.Ext(dsturl.Absolute()))
+	if contentType == "" {
+		return "application/octet-stream"
+	}
+	return contentType
+}
+
+// if we pass os.Stdin directly to Put function AWS SDK throws an error
+// ReadRequestBody: unable to initialize upload caused by: seek /dev/stdin: illegal seek
+// due to that stdin struct has only file without seek
```
seruman, line 347: It would be good to follow the convention; 
https://tip.golang.org/doc/comment#type

Instead of dumping error how about something more abstract like;
```
// stin is an io.Reader adapter for os.File, enabling it to function solely as
// an io.Reader. The AWS SDK, which accepts an io.Reader for multipart uploads,
// will attempt to use io.Seek if the reader supports it. However, os.Stdin is
// a specific type of file that can not seekable.

```

ahmethakanbesel: Thanks for the tip. I will update the comment.

seruman:
I think with refactoring `cp` command, we might able to use some of the code for stuff like common flags, content type guessing, override logic but not exactly sure if it worths, what do you think? @ahmethakanbesel @igungor

command/pipe.go
```diff
@@ -0,0 +1,355 @@
...
+	return nil
+}
+
+func guessStdinContentType(dsturl *url.URL) string {
+	contentType := mime.TypeByExtension(filepath.Ext(dsturl.Absolute()))
+	if contentType == "" {
+		return "application/octet-stream"
+	}
+	return contentType
+}
```
seruman, line 343: Not really guessing from stdin, maybe it could be renamed to something like `guessContentTypeByExtension`.

ahmethakanbesel: It used to try to predict file type from stdin. It didn't work as I expected then I removed that logic. I should rename the function. Thanks for the pointing that out.

ahmethakanbesel:
> I think with refactoring `cp` command, we might able to use some of the code for stuff like common flags, content type guessing, override logic but not exactly sure if it worths, what do you think? @ahmethakanbesel @igungor

`cp` has the `NewSharedFlags` function, which defines 18 different common flags for `cp` and `sync`. `pipe` uses 13 of them. We can reduce the number of shared flags to use them all in `cp`, `sync` and `pipe`, but it can affect the user experience.

seruman (commented):
@ahmethakanbesel Can you update the changelog please?

ahmethakanbesel:
> @ahmethakanbesel Can you update the changelog please?

I updated the changelog.

seruman (dismissed):

command/app.go
```diff
@@ -197,6 +197,7 @@ func Commands() []*cli.Command {
 	return []*cli.Command{
 		NewListCommand(),
 		NewCopyCommand(),
+		NewPipeCommand(),
```
igungor, line 200: Could you move this line under NewCatCommand()? It makes sense to show `cat` and `pipe` together.

command/pipe.go
```diff
@@ -0,0 +1,355 @@
...
+
+Usage:
+	{{.HelpName}} [options] destination
+
+Options:
+	{{range .VisibleFlags}}{{.}}
+	{{end}}
+Examples:
+	01. Upload stdin output to a bucket
+		 > echo "content" | s5cmd {{.HelpName}} s3://bucket/prefix/object.gz
```
igungor, line 30: let's make example better by doing exactly what it says:

`echo "object content" | gzip | ./s5cmd pipe s3://bucket/prefix/object.gz`

command/pipe.go
```diff
@@ -0,0 +1,355 @@
...
+		},
+		&cli.BoolFlag{
+			Name:  "raw",
+			Usage: "disable the wildcard operations, useful with filenames that contains glob characters",
+		},
+		&cli.StringFlag{
+			Name:  "content-type",
+			Usage: "set content type for target: defines content type header for object, e.g. --content-type text/plain",
+		},
+		&cli.StringFlag{
```
igungor, line 83: let's add `content-disposition` flag as well. https://github.com/peak/s5cmd/pull/589

command/pipe.go
```diff
@@ -0,0 +1,355 @@
...
+	partSize    int64
+	storageOpts storage.Options
+}
+
+// NewPipe creates Pipe from cli.Context.
+func NewPipe(c *cli.Context, deleteSource bool) (*Pipe, error) {
+	fullCommand := commandFromContext(c)
+
+	dst, err := url.New(c.Args().Get(0),
+		url.WithRaw(c.Bool("raw")))
```
igungor, line 164: there's only 2 parameters for url.New. Could you convert it to 1 line please?

command/pipe.go
```diff
@@ -0,0 +1,355 @@
...
+	}, nil
+}
+
+// Run starts copying stdin output to destination.
+func (c Pipe) Run(ctx context.Context) error {
+	var task = c.prepareUploadTask(ctx, c.dst)
+	var err = task()
+
+	return err
+}
```
igungor, line 200: This function is essentially 1 line if you take a close look:

`return c.prepareUploadTask(ctx, c.dst)()`

`prepareUploadTask` and `doUpload` helps `cp` command readable but in here, they don't add much. Let's `prepareUploadTask` and `doUpload` code into `Run` method.

command/pipe.go
```diff
@@ -0,0 +1,355 @@
...
+	} else {
+		metadata.SetContentType(guessContentTypeByExtension(dsturl))
+	}
+
+	if c.contentEncoding != "" {
+		metadata.SetContentEncoding(c.contentEncoding)
+	}
+
+	err = dstClient.Put(ctx, &stdin{file: os.Stdin}, dsturl, metadata, c.concurrency, c.partSize)
+
```
igungor, line 262: remove empty line

command/pipe.go
```diff
@@ -0,0 +1,355 @@
...
+	if err != nil {
+		return err
+	}
+
+	// if destination not exists, no conditions apply.
+	if dstObj == nil {
+		return nil
+	}
+
+	var stickyErr error
```
igungor, line 304: `stickyErr` is not sticky. Please remove the variable and return the error directly in 2 lines below.

e2e/pipe_test.go
```diff
@@ -0,0 +1,470 @@
...
+	"bytes"
+	"fmt"
+	"runtime"
+	"testing"
+
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/icmd"
+)
+
+// pipe s3://bucket/object
```
igungor, line 13:
```suggestion
// pipe s3://bucket/object.zip
```

e2e/pipe_test.go
```diff
@@ -0,0 +1,470 @@
...
+	assertLines(t, result.Stdout(), map[int]compareFunc{
+		0: equals(`pipe %v`, dstpath),
+	})
+
+	// assert s3 object
+	assert.Assert(t, ensureS3Object(s3client, bucket, fmt.Sprintf("s5cmdtest/%v", filename), content))
+}
+
+// pipe s3://bucket/object
+func TestUploadStdinToS3WithCustomName(t *testing.T) {
```
igungor, line 193: `pipe` doesn't work without a remote object name. This test looks duplicate.

e2e/pipe_test.go
```diff
@@ -0,0 +1,470 @@
...
+	assertLines(t, result.Stdout(), map[int]compareFunc{
+		0: equals(`pipe %v`, dstpath),
+	})
+
+	// assert s3 object
+	assert.Assert(t, ensureS3Object(s3client, bucket, filename, content))
+}
+
+// pipe -n s3://bucket/object
+func TestUploadStdinToS3WithNoClobber(t *testing.T) {
```
igungor, line 224: How does this test work? I missed the `-n` assertions.

command/app.go
ahmethakanbesel: Resolved. https://github.com/peak/s5cmd/pull/587/commits/c563a044e1c8e075e79c5c71b9317fd9fac7beb9

command/pipe.go
ahmethakanbesel: Resolved. https://github.com/peak/s5cmd/pull/587/commits/2f448b21806d42ba85a7a15f9a0a0578c16f7f4b

command/pipe.go
ahmethakanbesel: Resolved. https://github.com/peak/s5cmd/pull/587/commits/b75cde65e62a8a549f7b618cc6b99f99d7f4d345

e2e/pipe_test.go
ahmethakanbesel: Resolved. https://github.com/peak/s5cmd/pull/587/commits/c869dd1b9f071bf1d110ff5ee5078fdc9890ac49

igungor: 👍

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+	{{.HelpName}} - {{.Usage}}
+
+Usage:
+	{{.HelpName}} [options] destination
+
+Options:
+	{{range .VisibleFlags}}{{.}}
+	{{end}}
+Examples:
+	01. Upload stdin output to a bucket
```
igungor, line 29:
```suggestion
	01. Stream stdin to an object
```

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+		partSize:           c.Int64("part-size") * megabytes,
+		encryptionMethod:   c.String("sse"),
+		encryptionKeyID:    c.String("sse-kms-key-id"),
+		acl:                c.String("acl"),
+		cacheControl:       c.String("cache-control"),
+		expires:            c.String("expires"),
+		contentType:        c.String("content-type"),
+		contentEncoding:    c.String("content-encoding"),
+		contentDisposition: c.String("content-disposition"),
+		dstRegion:          c.String("destination-region"),
```
igungor, line 192: `destination-region` flag should not be necessary because there's only 1 client. `AWS_REGION` environment variable changes the client region if explicitly given.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+		return err
+	}
+	// override destination region if set
+	if c.dstRegion != "" {
+		c.storageOpts.SetRegion(c.dstRegion)
+	}
+	dstClient, err := storage.NewRemoteClient(ctx, c.dst, c.storageOpts)
+	if err != nil {
+		return err
+	}
```
igungor, line 219: please add newline.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+		if errorpkg.IsWarning(err) {
+			printDebug(c.op, err, nil, c.dst)
+			return nil
+		}
+		return err
+	}
+	// override destination region if set
+	if c.dstRegion != "" {
+		c.storageOpts.SetRegion(c.dstRegion)
+	}
```
igungor, line 215: please add newline.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+		return fmt.Errorf("target %q must be an object", c.dst)
+	}
+	err := c.shouldOverride(ctx, c.dst)
+	if err != nil {
+		if errorpkg.IsWarning(err) {
+			printDebug(c.op, err, nil, c.dst)
+			return nil
+		}
+		return err
+	}
```
igungor, line 211: please add newline.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+		// s3 options
+		storageOpts: NewStorageOpts(c),
+	}, nil
+}
+
+// Run starts copying stdin output to destination.
+func (c Pipe) Run(ctx context.Context) error {
+	if c.dst.IsBucket() || c.dst.IsPrefix() {
+		return fmt.Errorf("target %q must be an object", c.dst)
+	}
```
igungor, line 203: please add newline.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+		SetSSEKeyID(c.encryptionKeyID).
+		SetACL(c.acl).
+		SetCacheControl(c.cacheControl).
+		SetExpires(c.expires)
+
+	if c.contentType != "" {
+		metadata.SetContentType(c.contentType)
+	} else {
+		metadata.SetContentType(guessContentTypeByExtension(c.dst))
+	}
```
igungor, line 232: please add newline.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+	if c.contentEncoding != "" {
+		metadata.SetContentEncoding(c.contentEncoding)
+	}
+	if c.contentDisposition != "" {
+		metadata.SetContentDisposition(c.contentDisposition)
+	}
+	err = dstClient.Put(ctx, &stdin{file: os.Stdin}, c.dst, metadata, c.concurrency, c.partSize)
```
igungor, line 239: please add newline.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+	if err != nil {
+		return err
+	}
+	msg := log.InfoMessage{
```
igungor, line 243: please add newline.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+		Operation:   c.op,
+		Source:      nil,
+		Destination: c.dst,
+		Object: &storage.Object{
+			StorageClass: c.storageClass,
+		},
+	}
+	log.Info(msg)
```
igungor, line 251: please add newline.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+	if err != nil {
+		return &errorpkg.Error{
+			Op:  c.op,
+			Src: nil,
+			Dst: c.dst,
+			Err: err,
+		}
+	}
```
igungor, line 259: please add newline.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+	msg := log.InfoMessage{
+		Operation:   c.op,
+		Source:      nil,
+		Destination: c.dst,
+		Object: &storage.Object{
+			StorageClass: c.storageClass,
+		},
+	}
+	log.Info(msg)
+	if err != nil {
```
igungor, line 252: this block looks out of place.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+// shouldOverride function checks if the destination should be overridden if
+// the destination object and given pipe flags conform to the
+// override criteria.
+func (c Pipe) shouldOverride(ctx context.Context, dsturl *url.URL) error {
+	// if not asked to override, ignore.
+	if !c.noClobber {
+		return nil
+	}
+
+	dstClient, err := storage.NewClient(ctx, dsturl, c.storageOpts)
```
igungor, line 272: there's only 1 client. you can use `client` as var name.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+	if err != nil {
+		return err
+	}
+
+	dstObj, err := getObject(ctx, dsturl, dstClient)
```
igungor, line 277: `obj`.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+		return err
+	}
+
+	// if destination not exists, no conditions apply.
+	if dstObj == nil {
+		return nil
+	}
+
+	if c.noClobber {
+		err = errorpkg.ErrObjectExists
```
igungor, line 288:
```suggestion
		return errorpkg.ErrObjectExists
```

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	return err
```
igungor, line 291:
```suggestion
	return nil
```

command/pipe.go
```diff
@@ -0,0 +1,355 @@
...
+	dst := c.Args().Get(0)
+
+	dsturl, err := url.New(dst, url.WithRaw(c.Bool("raw")))
+	if err != nil {
+		return err
+	}
+
+	// destination must be a file
+	if dsturl.IsBucket() || dsturl.IsPrefix() {
+		return fmt.Errorf("target %q must be a file", dsturl)
```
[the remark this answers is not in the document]
igungor: this change is not applied.

command/pipe.go
```diff
@@ -0,0 +1,337 @@
...
+		return fmt.Errorf("expected destination argument")
+	}
+
+	dst := c.Args().Get(0)
+
+	dsturl, err := url.New(dst, url.WithRaw(c.Bool("raw")))
+	if err != nil {
+		return err
+	}
+
```
igungor, line 305: we should also check if dsturl is of remote type.

```
echo foo | ./s5cmd pipe heyyo
```

e2e/pipe_test.go
```diff
@@ -0,0 +1,481 @@
...
+			assertLines(t, result.Stdout(), map[int]compareFunc{
+				0: suffix(`pipe %v`, dstpath),
+			})
+
+			// assert S3
+			assert.Assert(t, ensureS3Object(s3client, bucket, objkey, content))
+		})
+
+	}
+
```
igungor, line 361: remove empty line

command/pipe.go
```diff
@@ -0,0 +1,328 @@
...
+		&cli.StringFlag{
+			Name:  "content-disposition",
+			Usage: "set content disposition for target: defines content disposition header for object, e.g. --content-disposition 'attachment; filename=\"filename.jpg\"'",
+		},
+		&cli.IntFlag{
+			Name:        "no-such-upload-retry-count",
+			Usage:       "number of times that a request will be retried on NoSuchUpload error; you should not use this unless you really know what you're doing",
+			DefaultText: "0",
+			Hidden:      true,
+		},
```
seruman, line 92: We should also remove this flag. 

As standart input is not seek-able, on such error it might be fully or partially consumed. To handle it we need to wait for standart input to be closed but doing makes the command behaves exactly like `cp` -with all contents in memory- that invalidates the reason we have this command in the first place.

> PS: We've seem to overlooked the case where reader would already be consumed when this option first introduced, it looks buggy anyways. I'll try to test and open a PR if necessary.

seruman (changes requested):

command/pipe.go
ahmethakanbesel: Removed. https://github.com/peak/s5cmd/pull/587/commits/bb450a131c93afa69403f9a8196ec989774dee61

seruman:
@ahmethakanbesel There's a conflict on CHANGELOG.

seruman (dismissed):

command/pipe.go
igungor: still relevant. please ping my comments if you've any relevant change.

command/pipe.go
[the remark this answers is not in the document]
igungor: @denizsurmeli have you applied?

command/pipe.go
ahmethakanbesel: Sorry, resolved now. https://github.com/peak/s5cmd/pull/587/commits/e128849a061b777e6e9eb233efabace52f53138b

command/pipe.go
ahmethakanbesel: Resolved. https://github.com/peak/s5cmd/pull/587/commits/8f4967ded03fa8ab1a1e81563624101aa0d5726c

command/pipe.go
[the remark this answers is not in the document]
ahmethakanbesel: Yes, https://github.com/peak/s5cmd/pull/587/commits/a572980d332d25f5adb245b5f93702e607ef6703. @igungor

igungor (approved):

igungor:
🎖️

seruman (approved):

PR #478, opened 2022-08-01
storage: migrating to aws-sdk-go-v2

boraberke:
This PR migrates `s5cmd` from `aws-sdk-for-go` to `aws-sdk-for-go-v2`.

#### Useful links for future reference:

1. [Github: aws-sdk-for-go-v2](https://github.com/aws/aws-sdk-go-v2/)
2. [Migration Guide](https://aws.github.io/aws-sdk-go-v2/docs/migrating/)
3. [Developer Guide](https://aws.github.io/aws-sdk-go-v2/docs/)

##### Major changes:
1.  `s3.go`
4. `s3_test.go`
5. `util_test.go`
6. `mock_s3.go`

##### Minor fixes:

1. `cat_test.go`
2. `cp_test.go`
3. `mb_test.go`
4. `rb_test.go`
5. `run_test.go`
7. `log.go`

#### Important changes:

1. [Sessions](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/sessions.html) have been removed in [aws-sdk-v2](https://aws.github.io/aws-sdk-go-v2/). Because of this, `s5cmd` will not have [sessionCache](https://github.com/peak/s5cmd/blob/master/storage/s3.go#L49) anymore. Instead of that, it will have `clientCache` which in the essence is the same.
8. The environment variable `AWS_SDK_LOAD_CONFIG` is not used anymore by the sdk. However, `s5cmd` will still not load from config files if this variable is set to 0 explicitly. Otherwise, it will load from default config files.
9. There is no `s3iface.S3API` in v2. Instead of [s3iface.S3API](https://github.com/aws/aws-sdk-go/blob/main/service/s3/s3iface/interface.go), s5cmd will have its own s3Client interface.
10. As `session` structure has been changed, unit tests also require changes. Instead of [unit.session](https://github.com/peak/s5cmd/blob/master/storage/s3_test.go#L219), mockgen will be used to mock `s3Client` interface and middleware will be used for some other tests such as `TestS3Retry`.
11. There is no `CredentialsChainVerboseErrors` setting in new sdk. There is an issue related to this [here](https://github.com/aws/aws-sdk-go-v2/issues/98).

#### Changes worth to mention:

1. There is no `WithDisableRestProtocolURICleaning` setting anymore as v2 does not [do any cleaning or url joining.](https://github.com/aws/aws-sdk-go-v2/issues/120)
[...]

PR #591, opened 2023-07-14
storage/s3: `endpoint_url` from AWS config file

jsproede:
Hi 👋

We've used this tool a lot since we've discovered `s5cmd` for our S3 buckets. As we're not using AWS, we need to set the `S3_ENDPOINT_URL` each time we open a new Terminal.

As the AWS-CLI [officially supports `endpoint_url`](https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html#ss-endpoints-config) in the AWS config file for a profile, I thought it would be great if `s5cmd` would support `endpoint_url` in AWS config profiles as well. Unfortunately [the AWS-SDK does not support `endpoint_url`](https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html#ss-endpoints-sdk-compat), therefore we're using the [go-ini module](https://ini.unknwon.io/docs) to read the AWS config file.

Therefore we extended the `newSession` function to use either the configured AWS config file (via `AWS_CONFIG_FILE` env-variable) or the default AWS config file (`$HOME/.aws/config`) and lookup/use the `endpoint_url` for the specified profile (only if configured and not specified by `$S3_ENDPOINT_URL` or the `--endpoint-url`-flag).

Additionally we created a test which covers the config file including a `endpoint_url` and a profile which has not set the `endpoint_url`.

What do you think about this?

If there is anything missing or we need to change something in this pull request, feel free to let us know 🙂

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/lanrat/extsort v1.0.0
 	github.com/termie/go-shutil v0.0.0-20140729215957-bcacb06fecae
 	github.com/urfave/cli/v2 v2.11.2
+	gopkg.in/ini.v1 v1.67.0
 	gotest.tools/v3 v3.0.2
 )
 
--- a/go.sum
+++ b/go.sum
@@ -109,6 +109,8 @@ gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8
 gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 h1:qIbj1fsPNlZgppZ+VLlY7N33q108Sa+fhmuc+sWQYwY=
 gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
 gopkg.in/errgo.v2 v2.1.0/go.mod h1:hNsd1EY+bozCKY1Ytp96fpM3vjJbqLJn88ws8XvfDNI=
+gopkg.in/ini.v1 v1.67.0 h1:Dgnx+6+nfE+IfzjUEISNeydPJh9AXNNsWbGP9KzCsOA=
+gopkg.in/ini.v1 v1.67.0/go.mod h1:pNLf8WUiyNEtQjuu5G5vTm06TEv9tsIgeAvK8hOrP4k=
 gopkg.in/mgo.v2 v2.0.0-20180705113604-9856a29383ce h1:xcEWjVhvbDy+nHP67nPDDpbYrY+ILlfndk4bRioVHaU=
 gopkg.in/mgo.v2 v2.0.0-20180705113604-9856a29383ce/go.mod h1:yeKp02qBN3iKW1OzL3MGk2IdtZzaj7SFntXj72NppTA=
 gopkg.in/yaml.v2 v2.2.2/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -1073,8 +1073,9 @@ func (sc *SessionCache) newSession(ctx context.Context, opts Options) (*session.
 				return nil, err
 			}
 
-			if awsProfileConfig.HasSection(opts.Profile) && awsProfileConfig.Section(opts.Profile).HasKey(endpointURLKey) {
-				opts.Endpoint = awsProfileConfig.Section(opts.Profile).Key(endpointURLKey).String()
+			awsProfileName := "profile " + opts.Profile
+			if awsProfileConfig.HasSection(awsProfileName) && awsProfileConfig.Section(awsProfileName).HasKey(endpointURLKey) {
+				opts.Endpoint = awsProfileConfig.Section(awsProfileName).Key(endpointURLKey).String()
 			}
 		}
 	}
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -22,6 +22,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -29,6 +30,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager/s3manageriface"
+	"gopkg.in/ini.v1"
 
 	"github.com/peak/s5cmd/v2/log"
 	"github.com/peak/s5cmd/v2/storage/url"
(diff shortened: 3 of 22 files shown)


jsproede:
I forgot, that AWS config profiles must be prefixed with `profile`. After pushing my changes the "ci / test (windows/go-1.17.x) (pull_request)" is now failing 🤔 I'll mark my PR as ready again. Can someone look into it (maybe restart the failing job)?

tilimil:
What about the "region" and "ca_bundle" in the config file as well?  It appears these aren't being consumed by s5cmd either

yakimant:
Thanks for the PR, looking forward for it being merged.
Thanks to the reviewers in advance!

jsproede:
> What about the "region" and "ca_bundle" in the config file as well? It appears these aren't being consumed by s5cmd either. Is there a more complete way to bring in all the setting in the config file and pass those to the SDK?

@tilimil Great suggestion. We've mainly focused on the endpoint-URL. I'm not sure, if the SDK could properly handle `region` and `ca_bundle` itself and therefore `go-ini` would not be necessary for these two configuration properties. I need to take a deeper look into the code of s5cmd to properly answer that. Maybe one of the maintainers can help us with that? :)

MrWaip:
Hi! It's nice feature to have

k0ste:
Needs rebase

jsproede:
@k0ste Since the PR has now been open since July 2023, I wanted to save myself the time of working on a regular rebase as long as there is no feedback, because I don't even know yet whether the PR would be accepted at all.

Muhamob:
Any updates on this? Looking forward for this feature

OS-M:
Any updates on this?

valerytschopp:
Any progress on this? AWS CLI `~/.aws/config` file with the `[profile USER]` section containing the `endpoint_url =`, or `region =` are very common...

jsproede:
I did not receive any response yet 😕

OS-M:
Finally I made a quick fork of s5cmd to fix this because it's so annoying to use --endpoint-url flag for our team (we're building yet another S3-compatible storage :/ ): https://github.com/OS-M/s5cmd.

You can install it with `go install github.com/OS-M/s5cmd/v2@v2.0.4`. However I give no guarantee of this working fine for everyone because my fix is too easy and straightforward.

OS-M:
Concerning this PR: when implementing my own fix I ran in a behaviour of s3-clients creation on **every** query because I was modifying it in the `newSession` method so the config was unmodified outside this method and the sessions cache won't work. 
Then I moved the modification to the `NewStorageOpts` method so make the cache work fine.
Here is the commit with the fix: https://github.com/OS-M/s5cmd/commit/178bece8e02255042a82649360e50abf0028a366.

I seems like this cache issue affects this PR too.

frittentheke:
@ilkinulas @seruman @denizsurmeli would you consider accepting this PR by any chance then?

seruman:
Hi all, @frittentheke, @jsproede ,

Sorry for late reply. *Personally* I'm all against reading the shared config file. I think s5cmd should leave all configuration to AWS SDK code and provide necessary flags to configure it, like `--endpoint-url` , `--source-region` etc. 

With reading config file by ourselves, we kinda mess with SDK's discovery mechanism, like;

```shell
; cat ~/.aws/credentials

[some-profile]
endpoint_url = http://localhost:4566
```

```shell
; AWS_PROFILE=some-profile ./s5cmd -r 0 ls
ERROR "ls": NoCredentialProviders: no valid providers in chain. Deprecated. For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```
but with `--profile` flag;

```shell
; ./s5cmd -r 0 --profile some-profile ls
ERROR "ls": SharedCredsAccessKey: shared credentials some-profile in /Users/selman/.aws/credentials did not contain aws_access_key_id
```

Looks OK, let's pass credentials via environment variables;

```shell
; AWS_ACCESS_KEY_ID="dummy" AWS_SECRET_ACCESS_KEY="dummy" ./s5cmd -r 0 --profile some-profile ls

ERROR "ls": SharedCredsAccessKey: shared credentials some-profile in /Users/selman/.aws/credentials did not contain aws_access_key_id
```

As if, you were to define endpoint-url in config file you won't be able to use other configuration mechanisms of the SDK.

I think we should aim for SDK upgrade, seems like SDK for Go v2 supports it;
https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html#ss-endpoints-sdk-compat

It's been long since we last tried but it was no avail due to breaking of GCS support; https://github.com/peak/s5cmd/pull/478#issuecomment-1249311139

seruman:
Hi @frittentheke,

Sorry for late reply. *Personally* I'm all against reading the shared config file. I think s5cmd should leave all configuration to AWS SDK code and provide necessary flags to configure it, like `--endpoint-url` , `--source-region` etc. 

With reading config file by ourselves, we kinda mess with SDK's discovery mechanism -IMHO-, like;

```shell
; cat ~/.aws/credentials

[some-profile]
aws_access_key_id =  dummy
aws_secret_access_key = dummy
region = us-east-1

; cat ~/.aws/config

[profile some-profile]
endpoint_url = http://localhost:4566

```

```shell
; AWS_PROFILE=some-profile ./s5cmd -r 0 ls
ERROR "ls": InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records. status code: 403, request id: T53PBF24JEMYVBQP, host id: WSjAFfmmffVRXTqh0Gfu9mdjeuqBJAh2XNhFxzJApMO4UpH4lAn5Xkt+xlq4jIOm0Lw7k2RjK0A=
```
This hits the real AWS, `AWS_PROFILE` environment variable seems to be unused. With `--profile` flag it works.

```shell
; ./s5cmd --profile some-profile ls ; echo $?
0

```

Instead of rolling our own solution, I think we should aim for SDK upgrade; seems like SDK for Go v2 supports it;
https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html#ss-endpoints-sdk-compat

It's been long since we last tried but it was no avail due to breaking of GCS support; https://github.com/peak/s5cmd/pull/478#issuecomment-1249311139

jsproede:
This PR has been open for over a year now, to be honest until this week without any response from the maintainers, although the desire to merge this PR has been expressed several times by others. When I implemented the changes with `go-ini`, `endpoint_url` was not supported by the AWS SDK for Go in AWS profiles. At that time, we worked relatively little with AWS, but with AWS-compatible services from third-party providers and it was annoying to have to set the `endpoint_url` again and again. We now mainly work with AWS services, so we are no longer really affected by the problem.

If this is now supported by V2, I would also prefer an SDK upgrade, but that is not part of this PR. After all this time, the message from one of the maintainers (which is not even addressed to the me, the PR creator) seems like this PR will not be approved anyway, so I will close this PR now.

frittentheke:
@ilkinulas @seruman @denizsurmeli I honestly understand @jsproede's frustration and am sad to see this PR go to waste.

s5cmd is really an awesome tool. But as an open-source tool it needs to evolve, be maintained and in that be open to feedback to bugs and lack of features from the community, including patches to address those.

Would you potentially pick up on the SDK upgrade approach yourself then? The issue of some half-hearted, pick an choose use of the AWS-like config methods still stands. The otherwise always appearing unexpected side effects such as `--cmd-line-parameters` vs. `ENV`-vars vs. `~/.aws/config` files really render ANY other solution compared to using AWS' own config library pointless.

seruman:
> After all this time, the message from one of the
maintainers (which is not even addressed to the me, the PR creator) seems like
this PR will not be approved anyway, so I will close this PR now.

@jsproede, I'm really sorry about this. I should have either given a proper review on time or mentioned you in the message above. I didn't mean to be disrespectful. 

You didn't have to close the PR. I didn't mean to imply that this PR would not be approved. I was merely stating my opinion on the matter. As almost all maintainers subscribed alongside the reviewers, they might have different opinions on the matter.

zalsader:
Any updates on this?

PR #593, opened 2023-07-14
cat: make it concurrent 

### LINKED ISSUE #245: 's5cmd cat' sub command is not using concurrent connections
's5cmd cat' sub command is not using concurrent connections, like 's5cmd cp' does.

My use-case is downloading a 427 GB tarball from S3 and extracting it on the fly:
`time s5cmd cat s3://bucket/file.tar.zst - | pzstd -d | tar -xv -C /`

Example EC2 instance type: c5d.9xlarge with 36 CPU cores, 72 GB RAM, 900 GB local SSD

When just comparing the download part with aws cli:
```
# time aws s3 cp s3://bucket/file.tar.zst - | cat >/dev/null
real    37m56.415s
user    22m50.195s
sys     19m8.677s
(around 192 MB/s)
```

With 's5cmd cat':
```
# time s5cmd cat s3://bucket/file.tar.zst >/dev/null
Still running. Only around 85 MB/s on a single S3 connection, according to netstat.
```

With 's5cmd cp' and writing to disk (without decompression):
```
time s5cmd cp s3://bucket/file.tar.zst /file.tar.zst
real    23m58.230s
user    7m56.734s
sys     22m40.482s
(around 304 MB/s)
```

With higher concurrency and larger parts:
```
# time s5cmd cp -c 36 -p 600 s3://bucket/file.tar.zst /file.tar.zst
real    10m3.064s
user    6m53.378s
sys     41m30.392s
(around 729 MB/s)

### PR DESCRIPTION
**Pull request by denizsurmeli**
This PR adds a new `io.WriterAt` adapter for non-seekable writers. It uses an internal linked list to order the incoming chunks. The implementation is independent from the [download manager](https://github.com/aws/aws-sdk-go/blob/main/service/s3/s3manager/download.go) of aws-sdk-go, and because of that currently it **can not bound** the memory usage. In order to limit the memory usage, we would have had to write a custom manager other than the aws-sdk-go's implementation, which seemed unfeasible. 

The new implementation is about %25 percent faster than the older implementation for a 9.4 GB file with `partSize=50MB` and `concurrency=20` parameters, with significantly higher memory usage, on average it uses 0.9 `GB` of memory and at most 2.1 `GB` is observed. Obviously, the memory usage and performance is dependent on the `partSize`-`concurrency` configuration and the link.

Resolves #245.

### CODE REVIEW

**Comment by denizsurmeli** [comment]
I have reduced the test size in here.
https://github.com/peak/s5cmd/blob/5b550850d640ab9f17b358e5d515a3581388e1b7/e2e/cat_test.go#L24
The reason for this is that since we have more test cases, we should aim to show that the feature/fix works and also maintain reasonable times during so. 
https://github.com/peak/s5cmd/blob/5b550850d640ab9f17b358e5d515a3581388e1b7/e2e/cat_test.go#L50-L75
Lowering the file size and adjusting the `partSize` and `concurrency` on the tests both show the feature works correctly while maintaining a feasible duration for the test.

File: buffer/llbuffer.go [hunk 1]
```diff
@@ -0,0 +1,107 @@
...
+	totalExpected int64
+	written       int64
+	list          *list.List
+	mu            *sync.Mutex
+	Queued        int
+}
+
+func NewOrderedBuffer(expectedBytes int64, w io.Writer) *OrderedBuffer {
+	return &OrderedBuffer{
+		totalExpected: expectedBytes,
```
igungor, line 25: `totalExpected` is unused, hence `expectedBytes` parameter is unused, hence extra `headobject` call in https://github.com/peak/s5cmd/pull/593/files#diff-015e2d270a318042cca60407874486ab8b45b1c3f60b81e9e1b3563ae5143b95R115 is unnecessary.

buffer/llbuffer.go
```diff
@@ -0,0 +1,107 @@
+package buffer
+
+import (
+	"container/list"
+	"io"
+	"sync"
+)
+
+type FileChunk struct {
```
igungor, line 9: no need to export this struct as it's not unused outside of this package.

buffer/llbuffer_test.go
```diff
@@ -0,0 +1,401 @@
+package buffer_test
```
igungor, line 1: We use same package name as the original package. Although this pattern is useful for public API testing, we don't do that on s5cmd.

buffer/llbuffer_test.go
```diff
@@ -0,0 +1,401 @@
...
+				i += chunkSize
+			}
+			// shuffle the chunks
+			for i := range chunks {
+				j := rand.Intn(i + 1)
+				chunks[i], chunks[j] = chunks[j], chunks[i]
+			}
+
+			// currently the max buffer size is disabled
+			rb := buffer.NewOrderedBuffer(expectedFileSize, &result)
```
igungor, line 111: what does `rb` stands for :)

buffer/llbuffer_test.go
```diff
@@ -0,0 +1,401 @@
...
+
+			for _, task := range chunks {
+				rb.WriteAt(task.content, int64(task.start))
+
```
igungor, line 115: remove empty line

buffer/llbuffer_test.go
```diff
@@ -0,0 +1,401 @@
...
+			}
+
+			// Ensure all chunks have been written correctly
+			if !bytes.Equal(result.Bytes(), expected) {
+				t.Errorf("Length comparison: Got: %d bytes, expected: %d bytes\n", len(result.Bytes()), len(expected))
+				t.Errorf("Got: %v, expected: %v\n", result.Bytes(), expected)
+			}
```
igungor, line 122: You could use `cmp.Diff` for prettier diff.

igungor (changes requested):

buffer/llbuffer.go
```diff
@@ -0,0 +1,107 @@
...
+	Content []byte
+}
+
+type OrderedBuffer struct {
+	w             io.Writer
+	totalExpected int64
+	written       int64
+	list          *list.List
+	mu            *sync.Mutex
+	Queued        int
```
seruman, line 20: Does `Queued` need to be exported?

buffer/llbuffer.go
denizsurmeli: You are right that `totalExpected` and `expectedBytes` are unnecessary, but the `HeadObject` is useful for correct error messages. When an object is not found, the head returns the humanly formatted error message while if we handle during the `Get` call, there are status code, host ID, etc. Here are the comparison for both: 
`HeadObject` error:
```
{"operation":"cat","command":"cat s3://test-cat-s-3-object-failcat-non-existent-r-ah5j9t5/prefix/file.txt","error":"given object s3://test-cat-s-3-object-failcat-non-existent-r-ah5j9t5/prefix/file.txt not found"}
```
`Get` error:
```
{"operation":"cat","command":"cat s3://test-cat-s-3-object-failcat-non-existent-r-rvvlx5y/prefix/file.txt","error":"NoSuchKey:  status code: 404, request id: 0000000000000003, host id: MDAwMDAwMDAwMDAwMDAwMzAwMDAwMDAwMDAwMDAwMDMwMDAwMDAwMDAwMDAwMDAzMDAwMDAwMDAwMDAwMDAwMw=="}
```

buffer/llbuffer.go
```diff
@@ -0,0 +1,107 @@
...
+	"io"
+	"sync"
+)
+
+type FileChunk struct {
+	Start   int64
+	Content []byte
+}
+
+type OrderedBuffer struct {
```
seruman, line 14: I'm not keen on the name, from API perspective this is an `io.WriterAt` adapter for an `io.Writer` which does buffering in order to satisfy the write order.

buffer/llbuffer.go
```diff
@@ -0,0 +1,107 @@
...
+		ob.Queued++
+		return len(p), nil
+	}
+	inserted := false
+	for e := ob.list.Front(); e != nil; e = e.Next() {
+		v, ok := e.Value.(*FileChunk)
+		if !ok {
+			// NOTE: should we handle this ?
+			continue
+		}
```
seruman, line 63: I think we may omit type assertion checks as list is internal to the `OrderedBuffer`, panicking here is not a bad thing IMHO, it would be better than silently ignoring.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 #### Features
 -- Added `--show-fullpath` flag to `ls`. (#[596](https://github.com/peak/s5cmd/issues/596))
 #### Improvements
+--- Implemented concurrent multipart download support for `cat`. ([#593](https://github.com/peak/s5cmd/pull/593))
```
seruman, line 9: Please use a single dash for bullet points;
image
https://github.github.com/gfm/#bullet-list-marker

buffer/llbuffer.go
denizsurmeli: It was a statistic, I think removing it completely would be better. The tests use it as an assertion that no items are left in the queue, but the assertion that checks the output would fail first if any items were left in the buffer. Just checking the result is sufficient, hence I will remove the field completely.

CHANGELOG.md
denizsurmeli: The line above has the same problem, would you like me to fix it in this PR or make another one ?

buffer/llbuffer_test.go
denizsurmeli: I have tried to implement a ring buffer first for this issue, however it didn't work. Unit tests were already written, hence `rb` from **R**ing**B**uffer :) Refactored.

buffer/llbuffer.go
denizsurmeli: Would `WriterAtAdapter` be good ?

CHANGELOG.md
igungor: you may change it in this pr since it's a trivial change.

buffer/llbuffer.go
seruman: How about `OrderedWriterAtAdapter` or `OrderedWriterAt` ? I'm not sure tbh not the best when it comes to naming 😄

denizsurmeli: The latter seems better. :) Refactoring to that.

buffer/buffer.go
```diff
@@ -0,0 +1,95 @@
...
+func NewOrderedWriterAt(w io.Writer) *OrderedWriterAt {
+	return &OrderedWriterAt{
+		written: 0,
+		list:    list.New(),
+		mu:      &sync.Mutex{},
+		w:       w,
+	}
+}
+
+func (ob *OrderedWriterAt) WriteAt(p []byte, offset int64) (int, error) {
```
igungor, line 33: OrderedWriterAt` is the only type with a receiver, so we can just swap `ob` with `w`.

buffer/buffer.go
```diff
@@ -0,0 +1,95 @@
...
+	ob.mu.Lock()
+	defer ob.mu.Unlock()
+
+	if ob.list.Front() == nil {
+		// if the queue is empty and the chunk is writeable, push it without queueing
+		if ob.written == offset {
+			n, err := ob.w.Write(p)
+			if n != len(p) || err != nil {
```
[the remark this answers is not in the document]
denizsurmeli: After reading the documentation of `Write`,  it's not necessary.

orderedwriter/orderedwriter.go
```diff
@@ -0,0 +1,96 @@
+// Package buffer implements an unbounded buffer for ordering concurrent writes for
+// non-seekable writers. It keeps an internal linked list that keeps the chunks in order
+// and flushes buffered chunks when the smallest expected offset is available.
```
seruman, line 3: Package is now called `orderedwriter`.

seruman (changes requested):

orderedwriter/orderedwriter_test.go
```diff
@@ -0,0 +1,324 @@
...
+			t.Parallel()
+			var result bytes.Buffer
+
+			// we know the filesize ahead so make it buffered
+			expected := make([]byte, fileSize)
+
+			for i := 0; i < fileSize; i++ {
+				ch := rand.Intn(255)
+				expected[i] = uint8(ch)
+			}
```
igungor, line 39: we've `randomString` helper function in `e2e` package, you may want to use it.

orderedwriter/orderedwriter_test.go
```diff
@@ -15,6 +15,16 @@ const (
 	testRuns = 64
 )
 
+func randomBytes(n int) []byte {
```
igungor, line 18: i meant, you can use `randomString` and convert it to byte slice :)

denizsurmeli: Since the function you have mentioned is under the package `e2e` and this test is under `orderedwriter`, I thought rather than importing it, copying would be better. Also avoiding unnecessary casting seemed better too :)

igungor: it's a test helper, so string->[]byte conversion cost is negligible. it might be better to create a new `testutil` package to store test helpers but it's not necessary for this PR.

igungor (approved):

seruman (approved):

PR #599, opened 2023-07-18
ls: add `--show-fullpath` flag

### LINKED ISSUE #596: Flatten list?
I see you have `--flatten` with `cp` but it seems like it's not supported for `ls`? 

I might be able to do a pull request to add this, it would be really useful for deeply nested directories (not my idea to store the data this way, I just get paid to clean it up).

denizsurmeli:
Resolves #596 .

### DIFF
--- a/command/ls.go
+++ b/command/ls.go
@@ -55,6 +55,9 @@ Examples:
 	10. List all versions of all objects in the bucket
 		 > s5cmd {{.HelpName}} --all-versions "s3://bucket/*"
 
+	11. List all files only with their fullpaths 
+		 > s5cmd {{.HelpName}} --show-fullpath "s3://bucket/*"
+
 `
 
 func NewListCommand() *cli.Command {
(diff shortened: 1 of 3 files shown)


e2e/ls_test.go
```diff
@@ -291,6 +291,44 @@ func TestListS3ObjectsAndFolders(t *testing.T) {
...
+	putFile(t, s3client, bucket, "f/txt/testfile10.txt", "content")
+
+	cmd := s5cmd("ls", "--show-fullpath", "s3://"+bucket+"/*")
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+
+	// assert lexical order
+	assertLines(t, result.Stdout(), map[int]compareFunc{
+		0: equals(fmt.Sprintf("s3://%v/a/testfile2.txt", bucket)),
```
sonmezonur, line 319: `equals` already accepts formatted strings. You can use like:

```go
0: equals("s3://%v/a/testfile2.txt", bucket),
```

sonmezonur (approved):

ilkinulas (approved):

command/ls.go
```diff
@@ -239,6 +250,9 @@ const (
 
 // String returns the string representation of ListMessage.
 func (l ListMessage) String() string {
+	if l.showFullPath {
+		return l.Object.URL.String()
+	}
```
igungor, line 255: I'm not sure that's what we want. I expected `ls` command output as-is, but are absolute instead of relative. With this flag, we only print full path without date and etag.

PR #606, opened 2023-07-21
e2e/run: update racy test case

### LINKED ISSUE #602: os.Rename() returns an error on Windows systems.
`os.Rename()` in  `Rename()` which is inside in `storage/fs.go` , returns an `Access Denied` error randomly on Windows systems.

Started to happen after the PR https://github.com/peak/s5cmd/pull/582

ahmethakanbesel:
Resolves #602

The `TestRunWildcardCountGreaterEqualThanWorkerCount` test used to download the same file to the same directory concurrently, and it used to cause `Access Denied` errors on Windows systems due to trying to rename a file while downloading a file with the same name at the same time.

Changes are made:
- Changed the download paths in the `TestRunWildcardCountGreaterEqualThanWorkerCount` test.

### DIFF
--- a/e2e/run_test.go
+++ b/e2e/run_test.go
@@ -220,9 +220,9 @@ func TestRunWildcardCountGreaterEqualThanWorkerCount(t *testing.T) {
 	putFile(t, s3client, bucket, "file.txt", "content")
 
 	content := []string{
-		"cp s3://" + bucket + "/f*.txt .",
-		"cp s3://" + bucket + "/f*.txt .",
-		"cp s3://" + bucket + "/f*.txt .",
+		"cp s3://" + bucket + "/f*.txt folder1",
+		"cp s3://" + bucket + "/f*.txt folder2",
+		"cp s3://" + bucket + "/f*.txt folder3",
 	}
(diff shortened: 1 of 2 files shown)


igungor (dismissed):

seruman (approved):

e2e/run_test.go
```diff
@@ -235,10 +235,12 @@ func TestRunWildcardCountGreaterEqualThanWorkerCount(t *testing.T) {
 	result := icmd.RunCmd(cmd)
 	result.Assert(t, icmd.Success)
 
+	fmt.Println(result.Stdout())
```
seruman, line 238: Left over debug print

seruman (changes requested):

e2e/run_test.go
ahmethakanbesel: Resolved. https://github.com/peak/s5cmd/pull/606/commits/427c2bf036e31711b9ba44fe804a8b6c162f326b

igungor (approved):

seruman (approved):

PR #607, opened 2023-07-21
command/run: set reader err if ctx is cancelled

ahmethakanbesel:
`r.err` should be set manually when `EOF` is  received  before `context.Canceled`.
It is important to stop the execution of Goroutines which use one common context.
Necessary for PR #597 

Any other solutions are welcome.


command/run.go
```diff
@@ -197,6 +198,10 @@ func (r *Reader) read() {
 			}
 			if err != nil {
 				if err == io.EOF {
+					if errors.Is(r.ctx.Err(), context.Canceled) {
+						r.err = r.ctx.Err()
+						return
```
sonmezonur, line 203: does not need early-return here. the following line already returns with an error.

PR #613, opened 2023-07-25
e2e/cp: update TestCopyLocalFileToS3WithNoClobber

ahmethakanbesel:
It used to not put an object in a bucket, so it was impossible to test whether the remote file was overwritten or not.


sonmezonur:
I think there is already an existing test for this scenario:

https://github.com/peak/s5cmd/blob/faaff9e615be224b934d263f2cf4e9c25492900c/e2e/cp_test.go#L2481

The purpose of the modified test was to check the behavior of `--no-clobber` if there is no object on target. Am I missing smth?

PR #583, opened 2023-07-03
github/workflows: Drop EOL versions and added latest 1.20

jpds:
- Dropped end-of-life 1.17 and 1.18.
- Added 1.20.

PR #622, opened 2023-07-27
all: min Go version is 1.19, build with 1.20

igungor:
- Custom `atomic` package is replaced with `sync/atomic.Bool` struct, which is added in Go 1.19.


igungor:
Changes a few more places in addition to #583

seruman (approved):

sonmezonur (approved):

PR #636, opened 2023-08-05
context: put quotes around flagvalue incase of whitespaced values

### LINKED ISSUE #541: Sync doesn't allow whitspaces in --cache-control

ataberkgrl:
Resolves https://github.com/peak/s5cmd/issues/541

While running the sync command it starts to generate cp commands in the generateCommand function and writes it to a pipe where it is read by the Run function of the Run command. In the Run function, there is a line 

`fields, err := shellquote.Split(line)`

to split the generated cp command into its fields. 

  For a command  
  
```  
s5cmd sync --cache-control ‘public, max-age=31536000, immutable’ /Users/ataberk/desktop/test s3://s5cmd-test2
```

We would have expected the generateCommand function to write into the pipe a command  

```
cp --cache-control='public, max-age=31536000, immutable' --raw='true' \"/Users/ataberk/desktop/test/hello.txt\" \"s3://s5cmd-test2/test/hello.txt\"
```

But instead, it writes the command  

```
cp --cache-control=public, max-age=31536000, immutable --raw=true \"/Users/ataberk/desktop/test/hello.txt\" \"s3://s5cmd-test2/test/hello.txt\"
```

Screenshot 2023-08-05 at 13 45 52

This causes the `shellquote.Split(line)` to split fields incorrectly. This causes `flagset.Parse(fields)` to populate flagset incorrectly and as a result we end up with an error.

Screenshot 2023-08-05 at 05 53 43

The main problem occurs in the generateCommand function while appending flags without quotes. If we add quotes around the flagValue while generating a command. The `shellquote.Split(line)` acts as intended and successfully sync with the given cache-control metadata in S3

Screenshot 2023-08-05 at 13 58 28

Screenshot 2023-08-05 at 13 59 31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Upgraded minimum required Go version to 1.19. ([#583](https://github.com/peak/s5cmd/pull/583))
 
 #### Bugfixes
+- Fixed a bug that causes `sync` command with whitespaced flag value to fail. ([#541](https://github.com/peak/s5cmd/issues/541))
 - Fixed a bug introduced with `external sort` support in `sync` command which prevents `sync` to an empty destination with `--delete` option. ([#576](https://github.com/peak/s5cmd/issues/576))
 - Fixed a bug in `sync` command, which previously caused the command to continue running even if an error was received from the destination bucket. ([#564](https://github.com/peak/s5cmd/issues/564))
 - Fixed a bug that causes local files to be lost if downloads fail. ([#479](https://github.com/peak/s5cmd/issues/479))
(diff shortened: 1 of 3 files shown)


sonmezonur (commented):
Thanks for the PR! Could you please add a test for your use case? (example command you shared)

sonmezonur (approved):

ilkinulas (approved):

7rulnik:
Thanks, guys ❤️

PR #661, opened 2023-09-14
dockerfile: use Alpine 3.18 as base for release image

embik:
Alpine 3.15 will be EOL on 2023-11-01 ([source](https://alpinelinux.org/releases/)). Releases happen quite frequently, but I wanted to propose updating the base image to Alpine 3.18 (which will be supported until 2025-05-09) so the EOL of 3.15 is anticipated for future releases.


sonmezonur (approved):

igungor (approved):

igungor:
@embik thanks!

PR #663, opened 2023-09-16
add `--log-stderr`: write log output to stderr

bsilverthorn:
Especially if you're using `s5cmd cat`, this flag lets you see log output even when redirecting stdout.

### DIFF
--- a/command/app.go
+++ b/command/app.go
@@ -59,6 +59,10 @@ var app = &cli.App{
 			Usage: "log level: (trace, debug, info, error)",
 		},
 		&cli.BoolFlag{
+			Name:  "log-stderr",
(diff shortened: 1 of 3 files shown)


denizsurmeli:
Hello, thanks for the PR. It's a needed feature I believe, but do you think rather than a bool flag, would it be better to make a flag that takes a filepath and write logs to the described file. We can have `STDOUT`, `STDERR` as enums, so we can handle those cases too. Something like this seemed feasible:
```
s5cmd --log-output=STDERR cat "s3://bucket/object"
```

```
s5cmd --log-output="debug.log" cat "s3://bucket/object"
```

Would like your comment on this approach. Obviously, we can direct `STDERR` and `STDOUT` to any file with redirection with this approach.

PR #675, opened 2023-10-24
add loong64 support

yzewei:
Provides the Loongarch architecture to support the S5CMD project.
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

One of the most important thing is there is a [docker image](https://github.com/merore/archlinux-loongarch64) where s5cmd tests can be done. You can use it on x86 machine with docker run -it --rm merore/archlinux-loongarch64.
About this image:
- Boot an Loongarch64-based Arch Linux with qemu-system-loongarch64
- The default user/password is root/loongarch64
- All dependcies required by criu build and test are pre-installed
- You must configure your own dependence on the environment, Golang's installation package is located at http://ftp.loongnix.cn/toolchain/golang/go-1.20/abi2.0/go1.20.2.linux-loong64.tar.gz


yzewei:
@igungor @nelhage Please~

PR #683, opened 2023-12-05
fix: renew the session token when the token expires

### LINKED ISSUE #678: ExpiredToken error when `cp` large dataset
Some context:
- large dataset: > 4TB, million of files
- s5cmd running in AWS EKS with assumeRole

I'm facing intermittent problem to upload or download with the following error:

```
MultipartUpload: upload multipart failed upload id: xxx-- caused by: ExpiredToken: The provided token has expired. status code: 400
```

After looking at the code, there is an [explicit deny for retrying](https://github.com/peak/s5cmd/blob/master/storage/s3.go#L1346) this kind of error.

The [PR](https://github.com/peak/s5cmd/pull/337) introducing this code seems to tackle another kind of issue. So I'm wondering if that would make sense to revisit the s3 session management.

MqllR:
This PR fix https://github.com/peak/s5cmd/issues/678

In a Kubernetes context, when running a very long process, we can easily fall into the `ExpiredToken` error. In order to not introduce any performance degradation, I propose to keep to global variable caching the session token but clear it when catching this specific error. The session will be renewed by reading again the projected volume `/var/run/secrets/eks.amazonaws.com/serviceaccount/token`.

### DIFF
--- a/storage/s3.go
+++ b/storage/s3.go
@@ -1339,11 +1339,6 @@ func newCustomRetryer(sc *SessionCache, maxRetries int) *customRetryer {
 // ShouldRetry overrides SDK's built in DefaultRetryer, adding custom retry
 // logics that are not included in the SDK.
 func (c *customRetryer) ShouldRetry(req *request.Request) bool {
-	log.Error(log.ErrorMessage{
-		Command: "retrier",
-		Err:     req.Error.Error(),
-	})
-
 	shouldRetry := errHasCode(req.Error, "InternalError") || errHasCode(req.Error, "RequestTimeTooSkewed") || errHasCode(req.Error, "SlowDown") || strings.Contains(req.Error.Error(), "connection reset") || strings.Contains(req.Error.Error(), "connection timed out") || errHasCode(req.Error, "ExpiredToken") || errHasCode(req.Error, "ExpiredTokenException")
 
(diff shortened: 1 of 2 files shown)


jasonyic:
this also an issue on out side and could speed up the review & release process, thank you.

storage/s3.go
```diff
@@ -1235,9 +1235,9 @@ func (sc *SessionCache) newSession(ctx context.Context, opts Options) (*session.
 			WithLogger(sdkLogger{})
 	}
 
-	awsCfg.Retryer = newCustomRetryer(opts.MaxRetries)
+	awsCfg.Retryer = newCustomRetryer(sc, opts.MaxRetries)
 
-	useSharedConfig := session.SharedConfigEnable
+	useSharedConfig := session.SharedConfigDisable
```
cynix, line 1240: Is this an accidental change?

l1n:
+1, this would be great to have!

tooptoop4:
@MqllR does the error only happen if s5cmd runs for more than hour? or it happens when trying to run short s5cmd on pod that is older than an hour?

terryrankine:
Same idea landed in terryrankine/s5cmd (91f45bd): ExpiredToken/ExpiredTokenException are retried; note the SDK's `AfterRetryHandler` calls `Credentials.Expire()` on those codes, which is what actually forces the refresh for SSO/assume-role providers — clearing a session cache on top is harmless but not the mechanism. Running in v2.4.1 of that fork.

PR #732, opened 2024-07-04
fix: ls differentiate exit codes for empty and non-existent buckets

### LINKED ISSUE #722: Error on empty bucket listing
`s5cmd` finishes with error (exit code = 1) when listing empty bucket:
```
s5cmd ls s3://empty-bucket; echo exit_code=$?
ERROR "ls s3://empty-bucket": no object found
exit_code=1
```
The same exit code returns for non-existent buckets:
```
s5cmd ls s3://non-existent-bucket; echo exit_code=$?
ERROR "ls s3://non-existent-bucket": NotFound: Not Found status code: 404, request id: tx000000000000000000014-0066647b03-8d9195fc-default, host id:
exit_code=1
```
There is no way to differentiate empty and non-existent buckets.
I think it is better not to display any error and return `0` exit code.
For example: `s3cmd` works as expected:
```
s3cmd ls s3://empty-bucket; echo exit_code=$?
exit_code=0

s3cmd ls s3://non-existent-bucket; echo exit_code=$?
ERROR: Bucket 'non-existent-bucket' does not exist
ERROR: S3 error: 404 (NoSuchBucket)
exit_code=12
```

s5cmd version: 2.2.2

tarikozyurtt:
Previously, s5cmd ls returned an exit code of 1 for both empty buckets and non-existent buckets, making it difficult to differentiate between the two cases. This change updates the behavior to:

- Return exit code 0 for empty buckets, similar to the behavior of s3cmd. 

- Return exit code 1 for non-existent buckets, providing a clear distinction. 

Example behavior after the change:

`s5cmd ls s3://empty-bucket` returns exit code 0. It does not print anything. `s5cmd ls s3://non-existent-bucket` returns exit code 1 with an appropriate error message. Also, select command with `--all-versions true` flag no longer prints `ERROR "s3://bucket/": no object found` message for empty buckets. Similarly, du command used to print
```
ERROR "du s3://empty-bucket": no object found
0 bytes in 0 objects: s3://empty-bucket
```
Now it will omit the part with error, it will yield:
```
0 bytes in 0 objects: s3://empty-bucket
```

They both exit with 0 without any change.

Resolves #722


denizsurmeli:
Hi, thanks for your contribution. It would be nice to extend test cases for demonstrating the behavior you have mentioned in the PR description:
>  Also, select command with --all-versions true flag no longer prints "ERROR " s3://bucket/": no object found" message for empty buckets. Similarly, du command used to print
"ERROR "du s3://empty-bucket": no object found
0 bytes in 0 objects: s3://empty-bucket"
Now it will omit "ERROR "du s3://bucket": no object found" part. They both exit with 0 without any change.

Adding these assertions as tests and when we run those tests in this [revision](https://github.com/peak/s5cmd/commit/1a5e5465b146d1e5c28a8809001e6c99fed6a656) they should fail.

e2e/ls_test.go
```diff
@@ -786,3 +786,20 @@ func TestListNestedLocalFolders(t *testing.T) {
...
+
+	bucket := s3BucketFromTestName(t)
+	createBucket(t, s3client, bucket)
+
+	cmd := s5cmd("ls", "s3://"+bucket)
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+
+	assertLines(t, result.Stdout(), map[int]compareFunc{}, alignment(true))
```
denizsurmeli, line 804:
```suggestion
	assertLines(t, result.Stdout(), nil)
```
Since we don't expect any output, the alignment option is not needed, we can also pass a `nil` map rather than an empty one. This one is a bit of nitpick :)

e2e/du_test.go
```diff
@@ -300,3 +300,22 @@ func TestDiskUsageByVersionIDAndAllVersions(t *testing.T) {
...
+
+	s3client, s5cmd := setup(t)
+
+	bucket := s3BucketFromTestName(t)
+	createBucket(t, s3client, bucket)
+
+	cmd := s5cmd("du", "s3://"+bucket)
+	result := icmd.RunCmd(cmd)
+
+	// Expecting success exit code with no error message
```
denizsurmeli, line 315:
```suggestion
```
The assertion is clear,  we can remove it.

e2e/select_test.go
```diff
@@ -218,6 +218,18 @@ func TestSelectCommand(t *testing.T) {
...
+					"--query", query,
+				},
+				informat:      "json",
+				structure:     "lines",
+				outformat:     "json",
+				expectedValue: "",
+			},
 		},
 		"csv": {
 			{
```
denizsurmeli, line 235: We can add tests for `csv` and `parquet` too.

tarikozyurtt: > We can add tests for `csv` and `parquet` too.

I added the test cases for CSV and fixed the errors I noticed in the test script. However, it seems that the latest Bitnami MinIO version still doesn't support Parquet files. Therefore, I couldn't run the tests through the `select_test.go` script. Additionally, since the `select` command requires an S3 premium membership, I wasn't able to run the tests myself. Nevertheless, there aren't any edge cases or changes specifically affecting Parquet functionality in my commit. Given that I couldn't run the tests and the chances of my changes affecting Parquet cases are almost negligible, I skipped test function for parquet files.

e2e/select_test.go
```diff
@@ -177,7 +177,7 @@ func TestSelectCommand(t *testing.T) {
 				expectedValue: "{\"id\":\"id0\"}\n",
 			},
 			{
-				name: "input:json-document,output:json,compression:gzip,select:with-document-access",
+				name: "input:json-document,output:json-lines,compression:gzip,select:with-document-access",
```
denizsurmeli, line 180: What is the reason for this naming change ?

e2e/select_test.go
```diff
@@ -402,9 +450,21 @@ func TestSelectCommand(t *testing.T) {
...
+				name: "input:csv,output:csv,delimiter:comma,all-versions:true,empty-bucket:true",
+				cmd: []string{
+					"select", "csv",
+					"--all-versions",
+					"--query", query,
+				},
+				informat:      "csv",
+				structure:     ",",
+				outformat:     "csv",
+				expectedValue: "-",
```
denizsurmeli, line 466: Please refer `"-"` with a const, defined on the top of the test file and refactor other references to `"-"` be referred with the const you have defined. As a name suggestion:
```suggestion
				expectedValue: EMPTY_BUCKET_VALUE,
```
Emphasis on suggestion, a better naming is always preferred :) 

But rather than overloading `expectedValue` with a special value that implicitly skips the file uploading to the bucket, a new field to testcase such as `emptyBucket` and running the test cases accordingly might be better. This review encapsulates the change made in the test's body. Namely, this change:
```go
	if tc.expectedValue == "-" { // test empty bucket case
		src = fmt.Sprintf("s3://%s/", bucket)
		expected = ""
	} else {
		putFile(t, s3client, bucket, filename, contents)
        }
```

denizsurmeli (changes requested):

e2e/select_test.go
sonmezonur, line 466: I couldn't understand the reason for its use. Here, we write expectedValue as `-` and then convert the `-` string to an empty string ('') in the function below. Wouldn't it be the same if we used an empty string as the expected value instead? It is confusing

e2e/select_test.go
```diff
@@ -15,6 +15,16 @@ import (
...
+type testcase struct {
+	name          string
+	cmd           []string
+	informat      string
+	structure     string
+	compression   bool
+	outformat     string
+	expectedValue string
+}
+
```
denizsurmeli, line 27: Each test leverages it's own testcase type. Moving this type definition in order to reuse in the `TestSelectWithEmptyBucket` is not feasible, since `TestSelectWithEmptyBucket` does not require `informat`, `structure`, `compression`, `outformat` or `expectedValue`, it only asserts that the stdout should be empty and process exists with 0. Thus, moving this struct back to where it belongs and creating a new struct in the `TestSelectWithEmptyBucket` that only has the required fields is better, also it's easier to read the tests later on.

There is a [good article](https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html) if you want to look at on reusing components on tests 😄

e2e/select_test.go
```diff
@@ -486,6 +523,72 @@ func TestSelectCommand(t *testing.T) {
...
+	endpoint := os.Getenv(s5cmdTestEndpointEnv)
+	if endpoint == "" {
+		t.Skipf("skipping the test because %v environment variable is empty", s5cmdTestEndpointEnv)
+	}
+
+	type testcase struct {
+		name          string
+		cmd           []string
+		expectedValue string
```
denizsurmeli, line 545:
```suggestion
		expectError bool
```
__Current cases__ require no `expectedValue`. When there is a case that will require `expectedValue`, the author can refactor accordingly. The assertions of these cases are there are no errors, refactoring the runner accordingly would be better.

e2e/select_test.go
```diff
@@ -486,6 +523,71 @@ func TestSelectCommand(t *testing.T) {
...
+			createBucket(t, s3client, bucket)
+
+			src := fmt.Sprintf("s3://%s/", bucket)
+			tc.cmd = append(tc.cmd, src)
+			cmd := s5cmd(tc.cmd...)
+
+			result := icmd.RunCmd(cmd, withEnv("AWS_ACCESS_KEY_ID", accessKeyID), withEnv("AWS_SECRET_ACCESS_KEY", secretKey))
+
+			result.Assert(t, icmd.Success)
+			assert.DeepEqual(t, "", result.Stdout())
```
denizsurmeli, line 586:
```suggestion
                  if !tc.expectError {
                  	result.Assert(t, icmd.Success)
			assert.DeepEqual(t, "", result.Stdout())
                  }
```
Body of the test should utilize the `expectError`, please use it for correct testcase execution. Considering that `expectError` can be `true`, body should behave accordingly.  If you think that it's not required, `expectError` can be removed completely.

e2e/select_test.go
denizsurmeli: Please let the reviewer decide whether the conversation is resolved or not, it's hard to track for the reviewer, every resolved conversation must be uncollapsed from the reviewers' side to see what was the response/comment.

e2e/select_test.go
```diff
@@ -165,7 +165,7 @@ func TestSelectCommand(t *testing.T) {
 				expectedValue: "id0\n",
 			},
 			{
-				name: "input:json-document,output:json,select:with-document-access",
+				name: "input:json-document,output:json-lines,select:with-document-access",
```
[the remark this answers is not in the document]
tarikozyurtt: > It is for standardizing output file format in "json" group. all the other elements in map's json part used `json-lines` for JSON output format cases.

Sorry, I didn't realized my comment was in pending state. This is the reason for the changes, if `document` and `lines` are descriptors of input data and not output, maybe I should change "json-lines" output format with "json" in all test cases?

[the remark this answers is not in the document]
denizsurmeli: `json-lines` does not imply something wrong, let's do it in another PR, it's not related to this one.

denizsurmeli (dismissed):
FYI: If your comments are in the _Pending_ state, the reviewers do not see it. Thus, some of the reviews might come a bit off, thinking that you have responded/changed already :)

sonmezonur (dismissed):

denizsurmeli (approved):

e2e/select_test.go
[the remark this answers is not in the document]
tarikozyurtt: Reverted it.

sonmezonur (approved):

PR #734, opened 2024-07-04
fix: sync glacier objects overwrite

### LINKED ISSUE #712: sync command copies a new object everytime when the s3 version is on GLACIER
I'm running version `v2.2.2-48f7e59` and when I add the `--storage-class 'GLACIER'` option to my command line, it does properly put the object in Glacier, but then subsequent sync commands fail to detect that version as the same as the one being on disk and subsequent runs just copy a new version over and over again.

This really defeats the purpose for me... I'm trying to upload backups of a large photo library to use as a cloud backup from my NAS. So there is rarely if ever a need to download everything, but I do... at times... update a few files here and there, mostly just adding files to an already existing directory... so this makes it impossible for me to write a script that executes via cron once a night to upload my backups.

4o4x:
This pull request addresses a bug (#712)  in the sync command where objects in Glacier storage at the destination were being overwritten during synchronization.  closes #712

## Bug Details:

During the sync process, if an object exists in both the source and destination, and the destination object is in Glacier storage, it was incorrectly being overwritten by the source object.

## Fixes:

- Modified the function that checks which objects to compare. 

Previously, if either the source or destination object was in Glacier storage, it was skipped. In the new version, only the source object is skipped, but the destination object is not. If the destination object is skipped, it will be considered as only in the source during the comparison phase, leading to it being overwritten by the source object.

- Added tests for this specific case to ensure correctness.


command/sync.go
```diff
@@ -547,7 +547,7 @@ func generateDestinationURL(srcurl, dsturl *url.URL, isBatch bool) *url.URL {
 }
 
 // shouldSkipObject checks is object should be skipped.
-func (s Sync) shouldSkipObject(object *storage.Object, verbose bool) bool {
+func (s Sync) shouldSkipObject(object *storage.Object, isSrc, verbose bool) bool {
```
ilkinulas, line 550: lets try to separate this into two different functions:
- shouldSkipSourceObject
- shouldSkipDestObject

we also need to check the existence of `--delete` flag when deciding whether to skip or not.

test-dir/myfile.txt
```diff
@@ -0,0 +1 @@
+Hello, LocalStack
```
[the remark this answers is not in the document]
4o4x: Apologies, I forgot to discard this directory before committing.

4o4x:
I wrote a script to observe how the sync command behaves with Glacier objects in AWS CLI. I used these flags in every possible combination:

- --delete
- --ignore-glacier-warnings
- --force-glacier-transfer
- --storage-class=STANDARD | GLACIER


The most important result obtained:

If the object in the source bucket is a Glacier object, no matter which flags are used, we cannot do anything to the object.
Here are the test results:

[delete.md](https://github.com/user-attachments/files/16131030/delete.md)
[sync.md](https://github.com/user-attachments/files/16131031/sync.md)

e2e/sync_test.go
```diff
@@ -852,6 +856,235 @@ func TestSyncS3BucketToS3BucketSameSizesSourceOlder(t *testing.T) {
...
+	t.Parallel()
+	s3client, s5cmd := setup(t)
+	bucket := s3BucketFromTestName(t)
+	dstbucket := s3BucketFromTestNameWithPrefix(t, "dst")
+
+	createBucket(t, s3client, bucket)
+	createBucket(t, s3client, dstbucket)
+
+	// put objects in source bucket
+	S3Content := map[string]string{
```
sonmezonur, line 870:
```suggestion
	s3Content := map[string]string{
```

4o4x: It has been taken from older tests. Should I change those too ?

ilkinulas:
> I wrote a script to observe how the sync command behaves with Glacier objects in AWS CLI. I used these flags in every possible combination:
> 
> * --delete
> * --ignore-glacier-warnings
> * --force-glacier-transfer
> * --storage-class=STANDARD | GLACIER
> 
> The most important result obtained:
> 
> If the object in the source bucket is a Glacier object, no matter which flags are used, we cannot do anything to the object. Here are the test results:
> 
> [delete.md](https://github.com/user-attachments/files/16131030/delete.md) [sync.md](https://github.com/user-attachments/files/16131031/sync.md)

Thank you for your efforts  on this.  What happens if the src or dst object's storage class is `glacier deep archive`? Does `GLACIER` also covers `DEEP ARCHIVE`?

command/sync_strategy_test.go
```diff
@@ -137,7 +137,32 @@ func TestSizeOnlyStrategy_ShouldSync(t *testing.T) {
 			dst:      &storage.Object{ModTime: timePtr(ft), Size: 10},
 			expected: errorpkg.ErrObjectSizesMatch,
 		},
+
+		{
```
ilkinulas, line 141: These test cases appear to be unnecessary. The sync strategy remains unchanged, and they only verify the size and modification time. Therefore, adding a test case for the storage class is not needed.

e2e/run_test.go
```diff
@@ -244,35 +244,35 @@ func TestRunWildcardCountGreaterEqualThanWorkerCount(t *testing.T) {
 	assertLines(t, result.Stderr(), map[int]compareFunc{})
 }
 
-func TestRunSpecialCharactersInPrefix(t *testing.T) {
-	t.Parallel()
+// func TestRunSpecialCharactersInPrefix(t *testing.T) {
```
ilkinulas, line 247: Please uncomment the test.

ilkinulas (changes requested):

e2e/sync_test.go
sonmezonur: I think we should. Variables should be camelcase

4o4x: done.

igungor:
@4o4x could you take a look at the PR description if it's still correct after the review changes please? thanks.

igungor:
@sonmezonur @ilkinulas PTAL

4o4x:
@ilkinulas @sonmezonur PTAL

sonmezonur (approved):

ilkinulas (approved):

4o4x:
Merge ?

PR #736, opened 2024-07-08
feat: add wildcard and prefix support to cat

### LINKED ISSUE #716: Support cat with glob
Hi there,

For many services like CloudTrail and S3 access logs, dominating factor regarding key count has more to do with architectural constraints of the writer than anything useful for downstream processing.

In the case of S3 access logs, ability to assemble millions of tiny files into a larger access log is pretty much a requirement before any meaningful local processing (e.g. DuckDB, clickhouse-local, ..) can occur.

Would therefore love if something like the following command could work:

`s5cmd cat s3://s3-logs-bucket/someprefix/2024-03-27-* > 2024-03-27.txt`

Similarly,

`s5cmd cat s3://..../* | clickhouse-local`

It's already easy to do this in 3 steps (cp, find | xargs cat > ...txt), but that at least requires having local storage available, and does not allow starting otherwise streamy processing (such as clickhouse-local reading from stdin) until all downloads are complete

tarikozyurtt:
This pull request enhances the cat command in s5cmd to support wildcard and prefix operations, improving its flexibility and usability for users. It includes comprehensive test cases.

- Wildcard Support:
**Example:** `s5cmd cat s3://bucket/foo*` concatenates content of files with paths starting with `s3://bucket/foo`. Use quotation marks according to your terminal type (e.g., bash, zsh).

- Prefix Support:
Enables cat to concatenate all files within a specified directory. Ignores nested folders. 
**Example:** `s5cmd cat s3://bucket/dir/` concatenates content of all files within the "dir" directory. But would not match with files inside `s3://bucket/dir/another_folder/`. Wildcard characters can be used for this case.

Resolves: #716


e2e/select_test.go
```diff
@@ -218,6 +218,18 @@ func TestSelectCommand(t *testing.T) {
...
+				cmd: []string{
+					"select", "json",
+					"--all-versions",
+					"--query", query,
+				},
+				informat:      "json",
+				structure:     "lines",
+				outformat:     "json",
+				expectedValue: "",
+			},
```
denizsurmeli, line 232: This testcase is unrelated to this PR.

e2e/ls_test.go
```diff
@@ -786,3 +786,19 @@ func TestListNestedLocalFolders(t *testing.T) {
...
+	bucket := s3BucketFromTestName(t)
+	createBucket(t, s3client, bucket)
+
+	cmd := s5cmd("ls", "s3://"+bucket)
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+
+	assertLines(t, result.Stdout(), nil)
+}
```
denizsurmeli, line 804: This testcase is unrelated to this PR.

e2e/du_test.go
```diff
@@ -300,3 +300,21 @@ func TestDiskUsageByVersionIDAndAllVersions(t *testing.T) {
...
+
+	cmd := s5cmd("du", "s3://"+bucket)
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+
+	assertLines(t, result.Stdout(), map[int]compareFunc{
+		0: suffix(`0 bytes in 0 objects: s3://%v`, bucket),
+	})
+}
```
denizsurmeli, line 320: This testcase is unrelated to this PR.

denizsurmeli (changes requested):

e2e/du_test.go
tarikozyurtt: [737](https://github.com/peak/s5cmd/pull/732) This PR contains changes from this other PR. I adjusted test cases according to the scenario where both PRs merged to the master branch. The other PR's changes must be present for the CI to pass in this branch. 
S, this PR can be merged after [737](https://github.com/peak/s5cmd/pull/732).

e2e/ls_test.go
tarikozyurtt: This PR contains changes from [737](https://github.com/peak/s5cmd/pull/732). I adjusted test cases according to the scenario where both PRs merged to the master branch. Other PRs changes must be present for the CI to pass in this branch. 
Therefore, this PR can be merged after [737](https://github.com/peak/s5cmd/pull/732).

e2e/du_test.go
tarikozyurtt: https://github.com/peak/s5cmd/pull/732 merged. This PR just contains aimed changes to solve related issue now. It can be reviewed.

e2e/cat_test.go
```diff
@@ -305,4 +316,150 @@ func TestCatByVersionID(t *testing.T) {
...
+
+	for _, tc := range testCases {
+		cmd := s5cmd("--json", "cat", fmt.Sprintf("s3://%v/%v", bucket, tc.prefix))
+		result := icmd.RunCmd(cmd)
+
+		result.Assert(t, icmd.Expected{ExitCode: 1})
+		assertLines(t, result.Stderr(), map[int]compareFunc{
+			0: equals(`{"operation":"cat","command":"cat s3://%v/%v","error":"no object found"}`, bucket, tc.prefix),
+		}, strictLineCheck(false))
+	}
```
seruman, line 453: you can combine both tests in single loop and define two subtests, something like ;
```go
for _, tc := testcases {
    t.Run(tc.name, func(t * testing.T){
        t.Run("default", func(t *testing.T){
            cmd := /* ... */
            // ...
        })


        t.Run("json", /* ... */)
    })
}
```

e2e/cat_test.go
```diff
@@ -305,4 +316,150 @@ func TestCatByVersionID(t *testing.T) {
...
+}
+
+func verifyCatCommand(t *testing.T, s5cmd func(...string) icmd.Cmd, bucket, expectedContent, prefix string) {
+	cmd := s5cmd("cat", fmt.Sprintf("s3://%v/%v", bucket, prefix))
+	result := icmd.RunCmd(cmd)
+
+	result.Assert(t, icmd.Success)
+	assertLines(t, result.Stdout(), map[int]compareFunc{
+		0: equals(expectedContent),
+	}, alignment(true))
```
seruman, line 464: as this is the whole test body, i think it should be in the tests themselves and not in another function. imho copying is not a bad thing when it comes to tests, keeping assertions as much as close to test cases makes it easier to follow.

if you really want to reuse this test body; in table test structs, you can define cmd, exitcode, stdout assertions, stderr assertions as fields and combine both TestCatWildcard and TestCatPrefix in a single test function, which *i think is **not necessary** in these tests*.

e2e/cat_test.go
```diff
@@ -305,4 +316,150 @@ func TestCatByVersionID(t *testing.T) {
...
+	s3client, s5cmd := setup(t)
+	bucket := s3BucketFromTestName(t)
+	createBucket(t, s3client, bucket)
+
+	testCases := []struct {
+		files    []string
+		prefix   string
+		expected string
+	}{
+		{files: []string{"file1.txt", "file2.txt"}, prefix: "", expected: "content0content1"},
```
seruman, line 360: when cases contains more than 2 or 3 fields with fields typed slices/maps, it is hard to read them at a first glance, it is better to spread out;

```go
{
            files: []string{"file1.txt", "file2.txt"},
            prefix: "",
            expected: "content0content1",
},
```

you can also omit zero values;

```go
{
            files: []string{"file1.txt", "file2.txt"},
            expected: "content0content1",
},
```

e2e/cat_test.go
```diff
@@ -305,4 +316,150 @@ func TestCatByVersionID(t *testing.T) {
...
+		verifyCatCommand(t, s5cmd, bucket, tc.expected, tc.prefix)
+	}
+}
+
+func TestCatWildcard(t *testing.T) {
+	t.Parallel()
+
+	s3client, s5cmd := setup(t)
+	bucket := s3BucketFromTestName(t)
+	createBucket(t, s3client, bucket)
```
seruman, line 387: no need to reuse these, we use gofake3 which is an inmemory s3 implementation in go. it is cheap to create fake s3 server and buckets. you can re-create those in test body to isolate cases.

e2e/cat_test.go
```diff
@@ -305,4 +316,150 @@ func TestCatByVersionID(t *testing.T) {
...
+	for _, tc := range testCases {
+		if tc.files != nil {
+			var concatenatedContent strings.Builder
+			for idx, file := range tc.files {
+				content := fmt.Sprintf("content%d", idx+offset)
+				putFile(t, s3client, bucket, file, content)
+				concatenatedContent.WriteString(content)
+			}
+			offset += len(tc.files)
+		}
```
seruman, line 377: this is so hard to understand, we have offset which is kept between cases, we create a concatted version of content but newer use it again?

i think it is better to define files and their contents explicitly in the test cases.

command/cat.go
```diff
@@ -111,16 +114,43 @@ func (c Cat) Run(ctx context.Context) error {
...
+	} else {
+		_, err = client.Stat(ctx, c.src)
+		if err != nil {
+			printError(c.fullCommand, c.op, err)
+			return err
+		}
+		singleObjChan := make(chan *storage.Object, 1)
+		singleObjChan <- &storage.Object{URL: c.src}
+		close(singleObjChan)
+		objectChan = singleObjChan
```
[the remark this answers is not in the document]
tarikozyurtt: Replaced the logic with this structure. Used `processSingleObject` and `processObjects` functions for different cases.

command/cat.go
```diff
@@ -111,16 +114,43 @@ func (c Cat) Run(ctx context.Context) error {
...
+		if obj.Type.IsDir() {
+			continue
+		}
+		buf := orderedwriter.New(os.Stdout)
+
+		_, err := client.Get(ctx, obj.URL, buf, c.concurrency, c.partSize)
+		if err != nil {
+			printError(c.fullCommand, c.op, err)
+			return err
+		}
```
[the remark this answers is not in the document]
tarikozyurtt: extracted that part to a function named `processSingleObject`. It takes `*storage.Object` which I pass `&storage.Object{URL: c.src}` for single object scenario.

e2e/cat_test.go
```diff
@@ -229,6 +205,41 @@ func TestCatLocalFileFail(t *testing.T) {
...
 
+func TestCatInEmptyBucket(t *testing.T) {
+	t.Parallel()
+
+	s3client, s5cmd := setup(t)
+
+	bucket := s3BucketFromTestName(t)
+	createBucket(t, s3client, bucket)
+
+	t.Run("EmptyBucket", func(t *testing.T) {
```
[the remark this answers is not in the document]
tarikozyurtt: replaced them with explanatory names: "without prefix", "with prefix" and "with wildcard".

e2e/cat_test.go
```diff
@@ -305,4 +316,150 @@ func TestCatByVersionID(t *testing.T) {
...
+		prefix   string
+		expected string
+	}{
+		{"foo*", "content0content1content3"},
+		{"log-file-2024-*", "content4content5"},
+		{"log-file-*", "content7content6content4content5"},
+	}
+
+	for _, tc := range testCases {
+		verifyCatCommand(t, s5cmd, bucket, tc.expected, tc.prefix)
```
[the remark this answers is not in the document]
tarikozyurtt: Applied to everywhere it should be applied.

e2e/cat_test.go
tarikozyurtt: Done.

e2e/cat_test.go
tarikozyurtt: Removed `verifyCatCommand` function and used it's content in both places it has been used.

e2e/cat_test.go
tarikozyurtt: Done.

e2e/cat_test.go
tarikozyurtt: Removed reuses except `TestCatInEmptyBucket` and `TestCatWildcard` functions where I run test commands in the same folder with the same content (in same empty bucket and same bucket with same content tree, respectively).

e2e/cat_test.go
tarikozyurtt: Changed logic. Added files to put, content they have, prefix to use in command and expected content for every separate test case. `strings.Builder` gone completely.

e2e/cat_test.go
```diff
@@ -305,4 +322,215 @@ func TestCatByVersionID(t *testing.T) {
...
+			files:    []string{"file1.txt", "file2.txt", "dir/file3.txt", "dir/file4.txt", "dir/nesteddir/file5.txt"},
+			contents: []string{"content0", "content1", "content2", "content3", "content4"},
+			prefix:   "dir/nesteddir/",
+			expected: "content4",
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.expected, func(t *testing.T) {
```
seruman, line 394: instead of expected, please give a name to your subtests, example;
https://github.com/peak/s5cmd/blob/master/storage/url/url_test.go#L16-L21

if not, just use iteration index of the loop.

e2e/cat_test.go
```diff
@@ -305,4 +322,215 @@ func TestCatByVersionID(t *testing.T) {
...
+		0: equals(`ERROR "cat --version-id=%v s3://%v/*": wildcard/prefix operations are disabled with --version-id flag`, version, bucket),
+	}, strictLineCheck(false))
+}
+
+func TestCatPrefix(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		files    []string
+		contents []string
```
seruman, line 359: why two slices instead of `map[string]string` ?

e2e/cat_test.go
```diff
@@ -305,4 +322,215 @@ func TestCatByVersionID(t *testing.T) {
...
+func TestCatWildcard(t *testing.T) {
+	t.Parallel()
+
+	s3client, s5cmd := setup(t)
+	bucket := s3BucketFromTestName(t)
+	createBucket(t, s3client, bucket)
+
+	files := []struct {
+		key     string
+		content string
```
seruman, line 427: could not we use `map[string]string` here?

e2e/cat_test.go
tarikozyurtt: Sure, named them case-[idx].

e2e/cat_test.go
tarikozyurtt: Used `map[string]string`.

e2e/cat_test.go
tarikozyurtt: Used `map[string]string` without any struct type.

seruman (approved):
🎉

igungor (approved):

denizsurmeli (approved):

PR #740, opened 2024-07-10
fix: fix single object cases for sync

### LINKED ISSUE #676: `sync` command copies object everytime
Native `aws s3 sync` command copies the file from S3 bucket to local only once (as expected):

`aws s3 sync s3://data/testfile .`

But `s5cmd sync` copies the file everytime and overwrites:

`s5cmd sync s3://data/testfile .`

tarikozyurtt:
The problem is mainly caused by the `compareObjects` function inside `command/sync.go`, where `s5cmd` compares source and destination paths and extracts files that are present only in the source or destination path (while also counting nested folders or rather name with its prefixes) along with common objects. If they both are non-objects, like wildcard expression, prefix, or bucket, getting relative paths of files with `src.URL.Relative()` results in compatible and comparable paths. In this case, no problem is present, at least within the scope of this issue.

However, when an object is selected as the source, it is not assigned a relative path using `func (u *url.URL) SetRelative(base *url.URL)`, so the `src.URL.Relative()` function returns its absolute path.

Let's say the source file has an absolute path of `folder/foo.txt`. The algorithm compares `folder/foo.txt` with `s3://bucket/remoteFolder/` and looks for the item `s3://bucket/remoteFolder/folder/foo.txt`. If it does not match, except for the edge case where there is a duplicate item inside the searched path, the files never match.

While copying files, `s5cmd` does not use relative paths, so `foo.txt` is written to the intended path in the remote. However, this happens during every sync operation, as they do not match.

Problem solved by taking path of source object as its name. This made algorithm to simply look for matches in destination, a file named `foo.txt` as intended. 

This PR adds new test cases to the sync command. Previously, tests failed to capture sync command cases where the source is an object in a prefix, not an object directly in a bucket, or not multiple objects like a wildcard or prefix expression. 

If an object is in the `s3://bucket/` path, its relative path is the same as its absolute path, so they match during comparison. This prevented copying the file every time. The new test cases cover all scenarios.
Resolves: https://github.com/peak/s5cmd/issues/676.

### DIFF
--- a/command/sync.go
+++ b/command/sync.go
@@ -261,9 +261,8 @@ func compareObjects(sourceObjects, destObjects chan *storage.Object, isSrcBatch 
 
 		for {
 			if srcOk {
-				if isSrcBatch {
-					srcName = filepath.ToSlash(src.URL.Relative())
-				} else {
+				srcName = filepath.ToSlash(src.URL.Relative())
+				if !isSrcBatch {
 					srcName = src.URL.Base()
 				}
 			}
--- a/command/sync.go
+++ b/command/sync.go
@@ -203,7 +203,7 @@ func (s Sync) Run(c *cli.Context) error {
 		isBatch = obj != nil && obj.Type.IsDir()
 	}
 
-	onlySource, onlyDest, commonObjects := compareObjects(sourceObjects, destObjects)
+	onlySource, onlyDest, commonObjects := compareObjects(sourceObjects, destObjects, isBatch)
 
 	sourceObjects = nil
 	destObjects = nil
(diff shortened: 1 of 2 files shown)


command/sync.go
```diff
@@ -261,7 +261,11 @@ func compareObjects(sourceObjects, destObjects chan *storage.Object) (chan *url.
 
 		for {
 			if srcOk {
-				srcName = filepath.ToSlash(src.URL.Relative())
+				if isSrcBatch {
+					srcName = filepath.ToSlash(src.URL.Relative())
+				} else {
+					srcName = src.URL.Base()
+				}
```
seruman, line 268:
```suggestion
				srcName = src.URL.Base()

				if isSrcBatch {
					srcName = filepath.ToSlash(src.URL.Relative())
				}
```

or

```suggestion
				srcName = filepath.ToSlash(src.URL.Relative())

				if !isSrcBatch {
					srcName = src.URL.Base()
				}
```


could be better.

seruman (commented):
thanks for the fix.

all my reviews are on the style side.

tarikozyurtt:
Fixed styles and naming.

seruman (approved):

ilkinulas (approved):

PR #793, opened 2025-02-12
fix: skip stat call when progress bar is disabled

### LINKED ISSUE #792: [Bug] Unnecessary `headObject` calls on `cp` command when progress bar is disabled
**Description:**  
When using the `cp` command to copy or download objects, the progress bar logic is making an unnecessary `headObject` call (via `client.stat`) to retrieve the size of the object, even when the progress bar is explicitly disabled. This results in:  

1. **Increased Latency:** Each unnecessary `headObject` call adds overhead, especially when downloading or copying a large number of small files.  
2. **Unnecessary Costs:** Since `headObject` calls are billed requests, this behavior incurs additional costs for users, particularly in use cases involving bulk operations.  

**Expected Behavior:**  
If the progress bar is disabled (default), the `cp` command should skip the `headObject` call entirely, as the object size is not needed for progress tracking.  

**Steps to Reproduce:**  
1. Run a `cp` command with the progress bar disabled (default).  
2. Observe network traffic logs (`--log=trace`) to confirm that a `headObject` call is still being made.  

**Impact:**  
- **Performance:** Slower operations due to unnecessary API calls.  
- **Cost:** Increased billing for `headObject` requests, especially in bulk operations.

ItielOlenick:
This PR fixes unnecessary headObject calls in the cp command when the progress bar is disabled. It ensures `Client.stat` (headObject when copying remote->local) is only called when the progress bar is explicitly enabled, reducing latency and costs for bulk operations.

Changes:

Added a conditional check to skip `Client.stat` unless the progress bar is active.

Only requests object size when necessary, improving performance and cost efficiency for cp operations.

Closes #792.

### DIFF
--- a/command/cp.go
+++ b/command/cp.go
@@ -521,7 +521,7 @@ func (c Copy) Run(ctx context.Context) error {
 		srcurl := object.URL
 		var task parallel.Task
 
-		if object.Size == 0 && !(srcurl.Type == c.dst.Type) {
+		if _, ok := c.progressbar.(*progressbar.CommandProgressBar); ok && object.Size == 0 && srcurl.Type != c.dst.Type {
 			obj, err := client.Stat(ctx, srcurl)
 			if err == nil {
 				object.Size = obj.Size


shaulre:
Any update?

Issue #794, opened 2025-02-18
Add extra virtual host style endpoinds support for s5cmd


LinPr:
## Question
Currently, s5cmd will use  path style endpoint by default when sending request to remote object storage, only when the remote endpoints match the following two endpoints will send request by virtual host style.
and they are hard coded in program. check this page 
https://github.com/peak/s5cmd/blob/master/storage/s3.go
```bash
	// Amazon Accelerated Transfer endpoint
	transferAccelEndpoint = "s3-accelerate.amazonaws.com"

	// Google Cloud Storage endpoint
	gcsEndpoint = "storage.googleapis.com"
```

But in some situation,  we can’t send request to the remote endpoint with path style request. For example , the alibaba object storage service.. It just banned the path style request by default, but only support virtual host style request by default , or you could ask for the official technical support , adding a white list for your path style endpoints. that not convenient when you have a bunch of buckets. 

## issue reference
#518 

## Suggestion
So, I recommend  to add a extra command line flag, and also works by passing value by environment variable , which support the users to add their custom endpoints. when sending request to those endpoints, those requests will automatically be transformed and sent by virtual host style  Also, this fucntion won't have any side affect on the original s5cmd command line funcions when not set.

And I will make a PR for this Issue,  I have implement this and well tested, hope this could be merged, and help people who face the same issue like me.

artemnikitin2023:
+1
I'm using custom cloud that supports only virtual-host style.

LinPr:
> +1
> I'm using custom cloud that supports only virtual-host style.

unfortunately，this project seems not being updated and merged new PR anymore, I have created this PR long time ago 😅

Issue #518, opened 2022-10-02
Using virtual hosted style in Alibaba OSS


sumodirjo:
Hello,

I got the following error when trying to list OSS bucket contents. e.g " s5cmd ls s3://<oss-bucket-name>/

```
ERROR "ls s3://<oss-bucket-name/": SecondLevelDomainForbidden: Please use virtual hosted style to access. status code: 403, request id: <request-id>, host id:
```

From Readme I learned:

> s5cmd will use virtual-host style bucket resolving for S3, S3 transfer acceleration and GCS. If a custom endpoint is provided, it'll fallback to path-style.

How can I set s5cmd to use virtual-host style when working with Alibaba OSS? Thank you.

igungor:
Hi,

Currently, there's no way to configure other than changing the source code. Please feel free to open a PR.

LinPr:
Alibaba OSS  forbid the path style host when uploading object to bucket, but you can request for technical support of Alibaba OSS， and  they will help add your bucket to a white list， then you can upload or download files with path style host.

And i have already verified this，it reall works !! 

![Image](https://github.com/user-attachments/assets/522bbb0e-fae0-44b1-9cd5-7af70c8d4034)

PR #795, opened 2025-02-18
feature: add extra virtual host style endpoints support

LinPr:
This PR add the user custom endpoints that need to send request with virtual host style 

reference issues:
#794 
#518 

Here I add a command line flag and also support fetch config from environment variable, 

`  --addressing-style value       use virtual host style or path style endpoint: (path, virtual) (default: path) `

the address-style config is compatible with the aws s3 cli config，here is the aws document , and you can just search the address_style key word in the page.    https://docs.aws.amazon.com/cli/latest/topic/s3-config.html

```bash
NAME:
   s5cmd - Blazing fast S3 and local filesystem execution tool

USAGE:
   s5cmd [global options] command [command options] [arguments...]

COMMANDS:
   ls              list buckets and objects
   cp              copy objects
   rm              remove objects
   mv              move/rename objects
   mb              make bucket
   rb              remove bucket
   select          run SQL queries on objects
   du              show object size usage
   cat             print remote object content
   pipe            stream to remote from stdin
   run             run commands in batch
   sync            sync objects
   version         print version
   bucket-version  configure bucket versioning
   presign         print remote object presign url
   head            print remote object metadata
   help, h         Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --addressing-style value       use virtual host style or path style endpoint: (path, virtual) (default: path) [$S3_ADDRESSING_STYLE]
   --credentials-file value       use the specified credentials file instead of the default credentials file
   --dry-run                      fake run; show what commands will be executed without actually executing them (default: false)
   --endpoint-url value           override default S3 host for custom services [$S3_ENDPOINT_URL]
   --help, -h                     show help (default: false)
   --install-completion           get completion installation instructions for your shell (only available for bash, pwsh, and zsh) (default: false)
   --json                         enable JSON formatted output (default: false)
   --log value                    log level: (trace, debug, info, error) (default: info)
   --no-sign-request              do not sign requests: credentials will not be loaded if --no-sign-request is provided (default: false)
   --no-verify-ssl                disable SSL certificate verification (default: false)
   --numworkers value             number of workers execute operation on each object (default: 256)
   --profile value                use the specified profile from the credentials file
   --request-payer value          who pays for request (access requester pays buckets)
   --retry-count value, -r value  number of times that a request will be retried for failures (default: 10)
   --stat                         collect statistics of program execution and display it at the end (default: false)
   --use-list-objects-v1          use ListObjectsV1 API for services that don't support ListObjectsV2 (default: false)
```

And I well tested it, using this on our daily work, no bug being found till now.

I can give a example here :

1. If I want to upload a file to remote object storage , Firstly, I need to set up config with environment variable

```bash
export AWS_ACCESS_KEY_ID='xxxxxxxxxxx'
export AWS_SECRET_ACCESS_KEY='xxxxxxxxxx'
export S3_ENDPOINT_URL='http://oss-cn-hangzhou.aliyuncs.com'
```

1. Secondly, because I want to sent request with virtual host style, I need a another env  or command line flag

```bash
export S3_ADDRESSING_STYLE="virtual"      
```

1. Now you can upload or down load file with the s5cmd command 

```bash
/s5cmd cp ./testfile s3://restic-backup-good/hello/
or I can run:
/s5cmd  --addressing-style=virtual cp ./testfile s3://restic-backup-good/hello/

output:
cp ./testfile s3://restic-backup-good/hello/testfile 
```

let’s see the request detail on wireshark.  it just send the request with virtual host style ,  a bucket name as prefix of host 

image

### DIFF
--- a/command/app.go
+++ b/command/app.go
@@ -90,10 +90,14 @@ var app = &cli.App{
 			Name:  "credentials-file",
 			Usage: "use the specified credentials file instead of the default credentials file",
 		},
-		&cli.StringFlag{
-			Name:    "support-virtual-host-style-endpoints",
-			Usage:   "add extra endpoints that support virtual host style requests, delimited by comma",
-			EnvVars: []string{"S3_SUPPORT_VIRTUAL_HOST_STYLE_ENDPOINTS"},
+		&cli.GenericFlag{
+			Name:  "addressing-style",
+			Usage: "use virtual host style or path style endpoint: (path, virtual)",
+			Value: &EnumValue{
+				Enum:    []string{"path", "virtual"},
+				Default: "path",
+			},
+			EnvVars: []string{"S3_ADDRESSING_STYLE"},
 		},
 	},
 	Before: func(c *cli.Context) error {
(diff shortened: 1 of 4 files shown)


salanki:
Getting this merged would be very nice.

ppcololo:
any updates?

jprobichaud:
I would second the importance of this PR.  We are using CoreWeave wth their CAIOS storage, which requires the virtual style, and not being able to use s5cmd is very painful !

mannreis:
@ilkinulas @sonmezonur are there any hopes on getting this merged? It would be really nice to have this upstream!

narenandu:
Could we get this PR merged or even https://github.com/peak/s5cmd/pull/859

Themis3000:
@jprobichaud 
> I would second the importance of this PR. We are using CoreWeave wth their CAIOS storage, which requires the virtual style, and not being able to use s5cmd is very painful !

You can still use s5cmd with other hosts. For example this is how I use s5cmd with mega's s4 storage:

`s5cmd --endpoint-url https://s3.ca-montreal.megas4.com ls`

Note that the --endpoint-url option must be specified *before* the command.

PR #841, opened 2025-10-06
vendor: Bump extsort version

NeonWizard:
Updating the `lanrat/extsort` version to fix a panic:
https://github.com/lanrat/extsort/issues/10

tl;dr:
This panic is caused when running `s5cmd` without enough memory, causing a fallback to extsort, and also running out of storage space in `/tmp`.


vendor/modules.txt
```diff
@@ -109,8 +109,8 @@ github.com/karrick/godirwalk
 github.com/kballard/go-shellquote
 # github.com/kr/pretty v0.3.0
 ## explicit; go 1.12
-# github.com/lanrat/extsort v1.0.2
-## explicit; go 1.13
+# github.com/lanrat/extsort v1.4.2
+## explicit; go 1.23
```
NeonWizard, line 113: Unfortunately, looks like the Go version in extsort was bumped in version v1.2.0. s5cmd would need to also update Go to 1.23 to take in this change

PR #847, opened 2025-11-02
fix: honor --profile flag for AWS credential and config loading

houseofeng:
- Allow --profile to use AWS credential chain instead of forcing ~/.aws/credentials
- Add Profile field to session.Options for proper shared config loading
- Maintains backward compatibility for --credentials-file usage

Resolves issue where --profile flag was ignored and application always used default profile from shared config if no credentials file was specified.


houseofeng:
Looking at the output of the failed integration test, I don't believe that has to do with the code changes. Please let me know if I'm mistaken.

PR #856, opened 2026-02-05
fix: use multipart copy for S3 objects larger than 5 GiB

### LINKED ISSUE #29: EntityTooLarge: fail to cp file with 6.3GB from one bucket to another bucket
fail to cp file with 6.3GB from one bucket to another bucket

EntityTooLarge: Your proposed upload exceeds the maximum allowed object size

l1n:
## Summary

`CopyObject` API has a 5 GiB limit. For objects exceeding this, `Copy` now falls back to multipart copy using `CreateMultipartUpload` + `UploadPartCopy` + `CompleteMultipartUpload`.

Closes #29

## Details

- The fallback is triggered when `CopyObject` returns `EntityTooLarge` or `InvalidRequest` with a "copy source is larger" message (the actual error S3 returns in practice).
- Uses 5 GiB part size to minimize API calls since the copy is entirely server-side (no data flows through the client).
- Aborts the multipart upload on failure, using `context.Background()` so cleanup succeeds even if the parent context is cancelled.
- All metadata fields (ACL, storage class, encryption, content-type, etc.) are forwarded to `CreateMultipartUpload`.
- Also passes `ctx` through to `CopyObjectWithContext` (previously unused).

## CI fix

Replaces the removed `bitnami/minio:2023.7.18` Docker Hub image with the official `minio/minio` image run as a background container step, fixing CI that has been broken since the bitnami image was removed.

## Testing

- Unit tests for `isCopySourceTooLargeError` (both error codes, negative cases)
- Integration test verifying the full fallback path: `CopyObject` failure → `HeadObject` → `CreateMultipartUpload` → `UploadPartCopy` x2 → `CompleteMultipartUpload`
- Manually tested against a real 6 GB S3 object (server-side copy between paths in the same bucket)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,24 +43,24 @@ jobs:
 
     name: test (${{ matrix.os }}/go-${{ matrix.go-version }})
     runs-on: ${{ matrix.os }}-latest
-    services:
-      minio:
-        image: ${{ (matrix.os == 'ubuntu') && 'bitnami/minio:2023.7.18' || ''}}
-        ports:
-          - 45677:9000
-        options: >-
-          --health-cmd "curl -I http://localhost:9000/minio/health/live -s"
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
-        env:
-          MINIO_ROOT_USER: minioadmin
-          MINIO_ROOT_PASSWORD: minioadmin
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-go@v2
       with:
         go-version: ${{ matrix.go-version }}
+
+    - name: Start MinIO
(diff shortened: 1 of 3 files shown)


terryrankine:
FYI: this has been running in production via terryrankine/s5cmd since March (v2.4.0, now v2.4.1). One thing we found and fixed there (https://github.com/terryrankine/s5cmd/commit/dbb57ec): the multipart path always merged the source object's headers/metadata into the new object, so `--metadata-directive REPLACE` leaked source metadata, and a KMS-encrypted source lost its `SSEKMSKeyId` when the request set no encryption. The request-building is extracted into `newMultipartCopyInput` with unit tests for COPY vs REPLACE — happy for you to pull it into this PR.

PR #860, opened 2026-04-12
fix(log): send Debug and Trace output to stderr (#851)

### LINKED ISSUE #851: Logs get into stdout. Messing up commands that pipe from cat.
I have a command like this: 
```
s5cmd --log debug -r 5 cat s3:/${in_path}/manifest.safe > "$manifest_data_file"
xmlstarlet sel -t -m '...' -v 's1:ascendingNodeTime' "$manifest_data_file" 
```

When running, a DEBUG statement gets in the output and breaks the XML:
```
/tmp/tmp.mG42gULExq:1.1: Start tag expected, '<' not found
DEBUG retryable error: ServiceUnavailable: Service Unavailable
^
```

It also happens with `--debug trace`
A fix like this could help: https://github.com/peak/s5cmd/pull/663
Is there an alternative? Changing all occurrences of `cat` to `cp` might be a bit involved

gustcol:
## Context
Fixes #851

## Root cause
`Trace()` and `Debug()` both called `global.printf(..., os.Stdout)`.
When a user ran `s5cmd --log debug cat s3://...` and piped the output
into another program, diagnostic messages were interleaved with the
object payload, corrupting the stream.

## Fix
Changed `Trace` and `Debug` to write to `os.Stderr`. `Info` continues
to write to `os.Stdout` (it carries functional output). Only `Error`
already used `os.Stderr`; this brings Debug and Trace in line with the
same convention.

Files changed: `log/log.go`

## Tests
- Added `TestDebugWritesToStderr` — asserts Debug output appears on stderr and nothing on stdout.
- Added `TestTraceWritesToStderr` — same for Trace.
- Added `TestInfoWritesToStdout` — regression guard confirming Info still goes to stdout.

## Manual verification
```
# Before: debug lines appear in stdout and break the pipe
s5cmd --log debug cat s3://bucket/object | wc -c   # inflated count

# After: only object bytes on stdout; debug lines on stderr
s5cmd --log debug cat s3://bucket/object 2>/dev/null | wc -c   # correct
```


gustcol:
## Test results

**Unit tests (`go test -race ./...`):** PASS (after adding e2e assertion fixes)

**Changes included in this branch beyond the original fix:**
- `e2e/cp_test.go`: 4 test functions updated — moved `DEBUG` assertions from `result.Stdout()` to `result.Stderr()`, asserted stdout empty
- `e2e/pipe_test.go`: 1 test function updated — same pattern
- `e2e/sync_test.go`: 6 test functions updated — for mixed-output tests (DEBUG + cp lines), split: DEBUG lines to `result.Stderr()`, cp lines remain in `result.Stdout()`

**Wasabi smoke test:**
```
$ ./s5cmd --log debug cp -n /tmp/smoke.txt s3://bucket/smoke.txt 2>/tmp/stderr >/tmp/stdout
$ cat /tmp/stdout     # empty
$ cat /tmp/stderr
DEBUG "cp /tmp/smoke.txt s3://bucket/smoke.txt": object already exists
```
Confirmed: DEBUG lines go to stderr; stdout stays clean.

PR #861, opened 2026-04-12
fix(rm): allow deletion of DIROBJ objects via --raw (#834)

### LINKED ISSUE #834: `s5cmd rm` does not delete DIROBJ objects
`s5cmd` does not delete DIROBJ objects - objects ending with `/`.

```
$ aws s3api put-object --bucket BUCKET --key some/path/dir_object/

$ s5cmd ls --show-fullpath s3://BUCKET/some/path/
s3://BUCKET/some/path/dir_object/

$ s5cmd rm 's3://BUCKET/some/path/*'
empty output

$ s5cmd rm s3://BUCKET/some/path/dir_object/
ERROR "rm s3://BUCKET/some/path/dir_object/": s3 bucket/prefix cannot be used for delete operations (forgot wildcard character?)

$ s5cmd rm --raw s3://BUCKET/some/path/dir_object/
ERROR "rm --raw=true s3://BUCKET/some/path/dir_object/": s3 bucket/prefix cannot be used for delete operations (forgot wildcard character?)

$ s5cmd ls --show-fullpath s3://BUCKET/some/path/
s3://BUCKET/some/path/dir_object/
```

While AWS CLI utility can do this:
```
$ aws s3 rm --recursive s3://BUCKET/some/path/
delete: s3://BUCKET/some/path/dir_object/

$ s5cmd ls --show-fullpath s3://BUCKET/some/path/
ERROR "ls --show-fullpath=true s3://BUCKET/some/path/": no object found
```

gustcol:
## Context
Fixes #834

## Root cause
`url.IsPrefix()` returns `true` for any remote path ending with `/`.
`validateRMCommand` unconditionally checked this and rejected keys
ending with `/` with the message *"s3 bucket/prefix cannot be used
for delete operations"*.

However, S3 allows zero-byte **directory marker** objects (keys ending
with `/`). These are real, deletable objects — not S3 prefixes — and
are created by tools like the AWS Console, Cyberduck, or s3cmd.

## Fix
When `--raw` is set the user has explicitly opted out of wildcard and
prefix treatment. The fix extends that opt-out to the `IsPrefix()` guard
in `validateRMCommand`:

```go
if srcurl.IsBucket() || (srcurl.IsPrefix() && !srcurl.IsRaw()) {
```

`IsRaw()` was already present on `URL` (it exposed the unexported `raw`
field). No new public API was added.

Files changed: `command/rm.go`

## Tests
- Added `TestRemoveCommand_DirobjKeyWithRawFlag` — passes `s3://bucket/dirobj/` with `--raw`; expects no validation error.
- Added `TestRemoveCommand_PrefixWithoutRawFlag` — regression guard confirming the guard still rejects prefix keys without `--raw`.

## Manual verification
```sh
# create a dir-marker object with the AWS CLI
aws s3api put-object --bucket my-bucket --key "dirobj/"

# before: fails with "bucket/prefix cannot be used..."
s5cmd rm s3://my-bucket/dirobj/

# after: succeeds
s5cmd rm --raw s3://my-bucket/dirobj/
```


gustcol:
## Test results

**Specific tests:**
```
go test -race -run "TestRemoveCommand_DirobjKeyWithRawFlag|TestRemoveCommand_PrefixWithoutRawFlag" ./command/...
ok  github.com/peak/s5cmd/v2/command
```

**Wasabi smoke test:**
```
# Created a real DIROBJ (key ending with /) via aws s3api put-object
# Then:
$ ./s5cmd rm --raw s3://bucket/dirobj-key/
rm s3://bucket/dirobj-key/
# Exit: 0 — deleted successfully

# Verified deletion:
$ aws s3api head-object --bucket bucket --key "dirobj-key/"
# 404 Not Found — confirmed deleted
```
Prefix guard without `--raw` continues to reject trailing-slash keys (no regression).

PR #863, opened 2026-04-12
fix(main): exit with code 130 on SIGINT (#615)

### LINKED ISSUE #615: Correct exit codes for SIGINT signals
Currently `s5cmd` does not return proper exit codes if we terminate the process with control c. 

```
s5cmd ls 's3://backups/*'
^C
❯ echo $?
0
```
```
s5cmd cp 's3://backups/*' 'dir/'
^C
❯ echo $?
1
```
`ls` returns success while `cp` returns with exit code 1. Exit codes for all commands should be consistent.

Maybe we should return `130` since most of the cli tools exit with 130 if the process is canceled by the user.

gustcol:
## Context
Fixes #615

## Root cause
Different commands returned different exit codes on Ctrl-C:
- `ls` returned **0** — the context-cancel path swallowed the error
- `cp` returned **1** — its context-cancel path propagated an error
- Neither returned **130**, the POSIX standard (128 + signal 2)

This made it impossible to reliably detect Ctrl-C in scripts vs real failures.

## Fix
Register a dedicated `os.Interrupt` notification channel (`intCh`) in
`main()`, alongside the existing `signal.NotifyContext`. After
`command.Main` returns, a non-blocking receive on `intCh` tells us
whether SIGINT was the cause. If yes, we call `os.Exit(130)` before
evaluating the command's own error return.

```go
intCh := make(chan os.Signal, 1)
signal.Notify(intCh, os.Interrupt)
defer signal.Stop(intCh)

err := command.Main(ctx, os.Args)

select {
case <-intCh:
    os.Exit(130)
default:
}
```

Files changed: `main.go`

## Tests
- Added `TestSIGINTExitCode` in `main_test.go`: builds the binary, runs
  a command against an unreachable endpoint so it blocks, sends SIGINT,
  and asserts exit code 130.

## Manual verification
```sh
s5cmd ls s3://bucket/ &
sleep 0.2
kill -INT %1
wait %1
echo $?  # 130
```

## Notes
- Exit code 130 is the POSIX standard for SIGINT-terminated processes.
- This is technically a breaking change for scripts that checked for
  exit code 0 or 1 specifically from Ctrl-C interrupts.


gustcol:
## Test results

**Unit tests (`go test -race ./...`):** PASS (after hardening the test)

**Changes included beyond the original fix:**
- `main_test.go`: Added `--no-sign-request` and `AWS_DEFAULT_REGION=us-east-1` to skip credential/region discovery so the binary blocks immediately on the TCP dial; increased pre-signal sleep to 3 s to absorb build-time variance under parallel test load. The 500 ms sleep was insufficient when running alongside the full `./...` suite.

**Specific tests (run 3 times to confirm stable):**
```
go test -race -count=1 ./...
ok  github.com/peak/s5cmd/v2   # includes TestSIGINTExitCode
```

**Wasabi smoke test:**
```
$ ./s5cmd cp /tmp/largefile.bin s3://bucket/large-upload-test &
PID=$!
sleep 2
kill -INT $PID
wait $PID
echo $?
130
```
SIGINT exits with code 130. SIGTERM or error exit remains 1.

PR #864, opened 2026-04-12
fix(sync): declare sort output channels as receive-only <-chan (#835)

### LINKED ISSUE #835: Cannot upgrade build later than Golang 1.22.x
There appears to be an issue with how the software handles generics in Go versions past 1.22.x, particularly in terms of interfacing with the extsort library. The following build error comes up, when attempting to use a later version. This error occurs irrespective of the version of extsort used.

`
command\sync.go:372:38: cannot use extsort.New(filteredSrcObjectChannel, storage.FromBytes, storage.Less, extsortConfig) (value of type <-chan extsort.SortType) as chan extsort.SortType value in assignment

command\sync.go:419:41: cannot use extsort.New(filteredDstObjectChannel, storage.FromBytes, storage.Less, extsortConfig) (value of type <-chan extsort.SortType) as chan extsort.SortType value in assignment
`

gustcol:
## Context
Fixes #835

## Root cause
`syncObjects` declared `srcOutputChan` and `dstOutputChan` as
bidirectional `chan extsort.SortType`. The channels come from
`extsort.New()`, which the library may return as a receive-only
`<-chan extsort.SortType` in certain versions. Go's type assignability
rules are strict about channel direction; assigning a receive-only
channel to a bidirectional variable is a compile error:

```
command/sync.go:372:38: cannot use extsort.New(...) (value of type
  <-chan extsort.SortType) as chan extsort.SortType value in assignment
```

## Fix
Change both variable declarations to `<-chan extsort.SortType`.
The channels are only ever ranged over (receive operations), so the
narrower type is semantically correct and future-proof against library
changes:

```go
var (
    sorter        *extsort.SortTypeSorter
    srcOutputChan <-chan extsort.SortType   // was: chan extsort.SortType
)
```

Files changed: `command/sync.go` (2 lines)

## Tests
- Existing `command` package tests pass.
- The fix is a pure type annotation change; no runtime behavior is affected.
- The build itself on Go 1.23+ is the regression test: if the types were
  wrong, compilation would fail.

## Manual verification
```sh
go build ./...      # must succeed on Go 1.22 and 1.23+
go test ./command/... # existing tests pass
```


gustcol:
## Test results

**Build (`make build`):** PASS

**Wasabi smoke test:**
```
$ ./s5cmd sync /tmp/syncdir/ s3://bucket/sync-test/
cp /tmp/syncdir/test1.txt s3://bucket/sync-test/test1.txt
cp /tmp/syncdir/test2.txt s3://bucket/sync-test/test2.txt
# Exit: 0 — sync runs correctly with receive-only channel types
```
The `<-chan` declaration is compatible with the existing code since the channels are only received from (via `range`), never sent to. Semantically correct and aligns with Go 1.23+ type assignability rules for `extsort.New` return values.

